PR #10, opened 2024-02-04
accel/amdxdna: fix a build error

langyuxf:
xdna-driver/src/driver/amdxdna/amdxdna_devel.c:28:3: error: a label can only be part of a statement and a declaration is not a statement
	28 |   struct iommu_domain *domain;


maxzhen:
I believe this is fixed already by PR#7

Issue #3, opened 2024-01-26
Rebase the required Linux kernel on top of latest released version


keryell:
The Linux kernel required to run this `xdna-driver` is based on an old release candidate of Linux 6.7 but Linux 6.7.2 has been released since with a lot of fixes (https://cdn.kernel.org/pub/linux/kernel/v6.x/ChangeLog-6.7.2).
It would be nice to have the latest version on https://github.com/AMDESE/linux to minimize troubles.
Technically I should open this issue on https://github.com/AMDESE/linux but that GitHub repository does not allow opening issue. Strange vision of open-source... ;-)

maxzhen:
We have no control over https://github.com/AMDESE/linux. But we need the change from that repository. For the team controlling  https://github.com/AMDESE/linux, they have other reasons / dependencies to choose this as base. So, this is not as flexible as you want.

superm1:
The dependency on patches in that repo is temporary, the patches in it are on their way upstream. Some are making it for 6.8, but a few will go out to 6.9 (or later depending upon code review).

K2:
The 6.7.2 PR I sent #7 was against a patched ESE tree into 6.7.2.  The patch from https://cdn.kernel.org/pub/linux/kernel/v6.x/patch-6.7.2.xz applies cleanly against the amdese tree with only 2 reg files, the primary Makefile subversion # and an HP driver one line fix to reapply it.

keryell:
I am back into this with my Ubuntu 23.10 laptop.
To have a stable mainline kernel, I have rebased `iommu_sva_v4_v6.7-rc8` Linux on top of latest `v6.7.4` as a  my `iommu_sva_v4_v6.7.4 branch`.
Then the problem to have something nicely working kernel is to start with a good `.config`. I got some inspiration from https://github.com/zabbly/linux and https://pkgs.zabbly.com/kernel/stable/pool/main/l/linux-zabbly-6.7.4-amd64-202402052246-ubuntu22.04/linux-image-6.7.4-zabbly%2B_6.7.4-amd64-202402052246-ubuntu22.04_amd64.deb

So now my minimal recipe for the kernel is:
```
git clone --depth=0 --branch iommu_sva_v4_v6.7.4 git@github.com:keryell/linux.git
cd linux
wget --output-document=.config https://gist.github.com/keryell/885c6963ec65bee6a03664e7379c876f/raw/9ed39d5adb07896f16a89def999aacd755a345cd/gistfile1.txt
[...]

PR #24, opened 2024-03-06
xdna-driver CI 

amd-akshatah:
CI source is not enabled by default. This will be enabled after the secrets are added , runners are added and internal repo is tested.

### DIFF
--- a/.github/workflows/xdna-driver-ci.yml
+++ b/.github/workflows/xdna-driver-ci.yml
@@ -1,0 +1,104 @@
+name: XDNA DRIVER CI
+
+on:
+   push:
+     #branches: [ "master" ]
+   pull_request_target:
+     #types: [opened, reopened, synchronize]
+   #workflow_dispatch:
+
+concurrency:
+  group: ${{ github.event.pull_request.head.repo.full_name }}-${{ github.event.pull_request.head.ref }}
+  cancel-in-progress: true
+
+jobs:
+   authorize:
+    runs-on: Ubuntu-22.04
+    #permissions:
+    #  id-token: write
+    #  contents: read
+    #  issues: write
+    #  pull-requests: write
+    steps:
+      - name: Checkout private repository
+        if: ${{ github.event_name == 'pull_request_target' }}
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ secrets.INTERNAL_REPO }}/xdna-driver-int
+          path: xdna-driver-int
+          ref: main
+          token: ${{ secrets.XDNA_DRIVER_GHA_PAT }}
+          fetch-depth: 0
+      - name: authorize
+        if: ${{ github.event_name == 'pull_request_target' }}
+        uses: ./xdna-driver-int/authorize
+        with:
+          pullRequestUser: ${{ github.event.pull_request.user.login }}
+
+   build:
+    needs: authorize
+    runs-on: [ Ubuntu-22.04 ]
+    steps:
+      - name: Checkout for Push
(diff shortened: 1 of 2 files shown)


.github/workflows/xdna-driver-ci.yml
```diff
@@ -0,0 +1,104 @@
...
+    runs-on: Ubuntu-22.04
+    #permissions:
+    #  id-token: write
+    #  contents: read
+    #  issues: write
+    #  pull-requests: write
+    steps:
+      - name: Checkout private repository
+        if: ${{ github.event_name == 'pull_request_target' }}
+        uses: actions/checkout@v3
```
keryell, line 25: To update to v4 everywhere I guess.

.github/workflows/xdna-driver-ci.yml
```diff
@@ -0,0 +1,104 @@
...
+           logFile: ${{ secrets.LOG_FILE }}
+           repoRoot: ${{ secrets.REPO_ROOT }}
+           user: ${{ secrets.USER }}
+           userPwd: ${{ secrets.USER_PASSWORD }}
+           ci: "true"
+           gitInternalRepo: ${{ secrets.GIT_INTERNAL_REPO }}
+           gitExternalRepo: ${{ secrets.GIT_EXTERNAL_REPO }}
+
+   upload:
+    if: ${{ always() }}
```
keryell, line 91: Is it interesting to upload in the case the workflow has been canceled?

.github/workflows/xdna-driver-pr-comment-ci.yml
```diff
@@ -0,0 +1,81 @@
+name: XDNA DRIVER PR COMMENT CI
+
+on:
+   issue_comment:
+    types: [created]
+
+jobs:
+   check:
+    runs-on: [ Ubuntu-22.04 ]
+    if: contains('amd-akshatah, kingtamamd, maxzhen, raphaelbamd, sonals, shirishjo ', format('{0},', github.actor)) && github.event.comment.body == '/build' && github.event.issue.pull_request
```
keryell, line 10: Since this is security-related and a bit convoluted, is this safe? It is unclear.
For example, if I have a user `sonal` or `hisonals`, does it work?
Relying on string conversion looks scary to me.

.github/workflows/xdna-driver-ci.yml
amd-akshatah: upload basically uploads the log file and internal users always want to see the log irrespective of whether the workflow is cancelled. The reason why we are uploading the log is because we are hiding the console output of the flow and routing it to a log file for internal use only.

.github/workflows/xdna-driver-pr-comment-ci.yml
amd-akshatah: Agreed, Let me check if I can add a group instead of string with a change in this logic.

.github/workflows/xdna-driver-ci.yml
keryell: Good point. You could add this as a comment to enlighten the reader.

keryell:
I am curious about which kernel is actually used to run the CI.
It would be nice to have some tests against real kernels people will use in production. Cf. https://github.com/amd/xdna-driver/issues/3 for example.

maxzhen:
> I am curious about which kernel is actually used to run the CI. It would be nice to have some tests against real kernels people will use in production. Cf. #3 for example.

Production testing is separate. CI is just to sanity test the PR. As for the kernel, it will be close to latest released Linux kernel.

PR #35, opened 2024-04-12
Update to latest XRT version supporting Linux 6.8

keryell:
Use latest XRT which compiles for stable Linux 6.8 version.
Update documentation according to latest kernel.

This solves https://github.com/amd/xdna-driver/issues/3

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@ This repository is for the AMD XDNA™️ Driver (amdxdna.ko) for Linux®️ and
 ## Table of Contents
 - [Introduction](#introduction)
 - [System Requirements](#system-requirements)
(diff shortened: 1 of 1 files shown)


keryell:
Is there a result of the CI somewhere?

keryell:
I have removed the commit which change the submodule to the latest XRT for process reasons.
The latest XRT will work.
More generally, it would be nice to have in the CI the latest XRT with the latest Linux to catch the early bugs. :smile_cat:

keryell:
@maxzhen you can update XRT to the latest version
If you can get some trivial bug fix in it before like https://github.com/Xilinx/XRT/pull/8046 it will be even better. :smile:

PR #42, opened 2024-04-17
Define correct mpnpu_clk/hclk ranges for npu devices

gyang1099:
The max mpnpu_clk/hclk ranges are different between npu1 and npu2/npu4, add clock range definitions in const struct npu_dev_priv and use for clock settings

When setting npu clocks, display clock value in dmesg

### DIFF
--- a/src/driver/amdxdna/npu1_pci.h
+++ b/src/driver/amdxdna/npu1_pci.h
@@ -59,6 +59,11 @@
 	pci_resource_len(NDEV2PDEV(_ndev), (_ndev)->xdna->dev_info->mbox_bar); \
 })
 
+#define SMU_MPNPUCLK_FREQ_MAX(ndev) \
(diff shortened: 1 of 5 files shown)


maxzhen (approved):

mamin506:
The change looks good.

After this change, the clock config info will be printed on dmesg log. For example,

```
[ 1036.133039] amdxdna 0000:c3:00.1: enabling device (0000 -> 0002)
[ 1036.173487] amdxdna 0000:c3:00.1: set mpnpu_clock = 600 mhz
[ 1036.193417] amdxdna 0000:c3:00.1: set npu_hclock = 1024 mhz
[ 1036.226052] [drm] Initialized amdxdna_accel_driver 1.0.0 20240124 for 0000:c3:00.1 on minor 0
```

PR #44, opened 2024-04-18
Mailbox sanity check in rx worker

mamin506:
1. When destroy channel, disalbe irq and cancel rx work in front.
2. Check value read from mailbox. Even firmware is untrusted, host driver should not crash.
3. When RX worker read corrupted data from ringbuf. Disable IRQ and mark channel as bad state.

### DIFF
--- a/src/driver/amdxdna/npu_mailbox.c
+++ b/src/driver/amdxdna/npu_mailbox.c
@@ -613,6 +613,13 @@ int xdna_mailbox_destroy_channel(struct mailbox_channel *mb_chann)
 	list_del(&mb_chann->chann_entry);
 	mutex_unlock(&mb_chann->mb->mbox_lock);
 
+	/* Disalbe an irq and wait. This might sleep. */
+	disable_irq(mb_chann->msix_irq);
+
+	/* Cancel RX work and wait for it to finish */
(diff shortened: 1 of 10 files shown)


houlz0507 (approved):

src/driver/amdxdna/npu_mailbox.c
```diff
@@ -613,6 +647,13 @@ int xdna_mailbox_destroy_channel(struct mailbox_channel *mb_chann)
 	list_del(&mb_chann->chann_entry);
 	mutex_unlock(&mb_chann->mb->mbox_lock);
 
+	/* Disalbe an irq and wait. This might sleep. */
+	disable_irq(mb_chann->msix_irq);
+
+	/* Cancel RX work and wait for it to finish */
+	cancel_work_sync(&mb_chann->rx_work);
```
maxzhen: I'm OK with taking disable/cancel type of action, but not comfortable with doing "destroy" and "free", before the context is completed removed from device. Can we split them and only do the "disable" and "cancel" part upfront?

maxzhen (changes requested):

maxzhen (approved):

PR #47, opened 2024-04-19
Bug fixes

mamin506:
1. OS can call flush fop multiple times, and list_del() will set poison pointer. This leads to kernel panic. Use list_del_init() instead.
2. If both dev_lock and hwctx_srcu lock need to be held, should acquire dev_lock first. See amdxdna_drm_config_hwctx_ioctl() change.
3. Some minor mailbox related fixes.

### DIFF
--- a/src/driver/amdxdna/npu1_message.c
+++ b/src/driver/amdxdna/npu1_message.c
@@ -418,10 +418,8 @@ int npu1_config_cu(struct amdxdna_hwctx *hwctx)
 	req.num_cus = hwctx->cus->num_cus;
 
 	ret = npu_send_msg_wait(xdna, chann, &msg);
-	if (ret == -ETIME) {
-		xdna_mailbox_stop_channel(chann);
-		xdna_mailbox_destroy_channel(chann);
-	}
+	if (ret == -ETIME)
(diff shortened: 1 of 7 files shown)


src/driver/amdxdna/npu1_message.c
```diff
@@ -416,8 +418,10 @@ int npu1_config_cu(struct amdxdna_hwctx *hwctx)
 	req.num_cus = hwctx->cus->num_cus;
 
 	ret = npu_send_msg_wait(xdna, chann, &msg);
-	if (ret == -ETIME)
-		npu1_destroy_context(xdna->dev_handle, hwctx);
+	if (ret == -ETIME) {
+		xdna_mailbox_stop_channel(chann);
+		xdna_mailbox_destroy_channel(chann);
```
maxzhen: No need to do below here?
hwctx->priv->mbox_chann = NULL;
hwctx->fw_ctx_id = -1;

mamin506: we should do npu1_destroy_context(), since this channel doesn't response.

maxzhen (approved):

keryell:
The PR description is quite interesting and this should appear somehow in some code comments too, otherwise this information is lost for the future generations.

PR #48, opened 2024-04-19
add 5 seconds DRM scheduler timeout

mamin506:
1. Set 5 seconds DRM scheduler timeout.
2. hwctx->name contains id. Dont duplicate. The hwctx-name is like: "hwctx.<pid>.<id>"


maxzhen (approved):

keryell:
Feel free to add (a lot of) comments, so we might have a chance to understand the code. ;-)

PR #51, opened 2024-04-22
Some bug fix

mamin506:
1. Change mm_lock to spin lock.
2. npu1_hwctx_restart() should always restart drm scheduler to free jobs.
3. Hold dev_lock in debugfs nputest, to avoid warning.
4. Set DRM scheduler timeout to 16 seconds.

### DIFF
--- a/src/driver/amdxdna/amdxdna_drv.c
+++ b/src/driver/amdxdna/amdxdna_drv.c
@@ -77,7 +77,7 @@ skip_sva_bind:
 	mutex_init(&client->hwctx_lock);
 	init_srcu_struct(&client->hwctx_srcu);
 	idr_init(&client->hwctx_idr);
-	mutex_init(&client->mm_lock);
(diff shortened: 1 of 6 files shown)


keryell:
> Set DRM scheduler timeout to 16 seconds.

It is unclear to me what it is used for and what the concrete impact for an end-user is.
For example, does-it mean that an NPU application should execute in less than 16 seconds or it is a failure?

fifield:
> > Set DRM scheduler timeout to 16 seconds.
> 
> It is unclear to me what it is used for and what the concrete impact for an end-user is. For example, does-it mean that an NPU application should execute in less than 16 seconds or it is a failure?

Yes, it's a failure to not finish in 16 seconds. It fails with `ERT_CMD_STATE_ABORT`. I just learned this the hard way debugging a failing test in mlir-air where this was the culprit. I was getting zeros in my output until I increased the timeout in the driver.

PR #75, opened 2024-05-09
Support command list execution (part 1)

mamin506:
Added command list support in driver.

### DIFF
--- a/src/driver/amdxdna/aie2_ctx.c
+++ b/src/driver/amdxdna/aie2_ctx.c
@@ -511,19 +511,20 @@ int aie2_hwctx_init(struct amdxdna_hwctx *hwctx)
 		return -ENOMEM;
 
 	hwctx->priv = priv;
-	heap = amdxdna_gem_get_obj(client, client->dev_heap, AMDXDNA_BO_DEV_HEAP);
+	heap = client->dev_heap;
 	if (!heap) {
+		XDNA_ERR(xdna, "The client dev heap object not exist");
 		ret = -EINVAL;
-		XDNA_ERR(xdna, "Cannot get dev heap object, ret %d", ret);
 		goto free_priv;
 	}
 
 	ret = amdxdna_gem_pin(heap);
 	if (ret) {
 		XDNA_ERR(xdna, "Dev heap pin failed, ret %d", ret);
-		goto put_heap;
+		goto free_priv;
 	}
 	priv->heap = heap;
+	drm_gem_object_get(to_gobj(heap));
 
 	for (i = 0; i < ARRAY_SIZE(priv->cmd_buf); i++) {
 		struct amdxdna_gem_obj *abo;
@@ -600,8 +601,6 @@ free_cmd_bufs:
 		drm_gem_object_put(to_gobj(priv->cmd_buf[i]));
 	}
 	amdxdna_gem_unpin(heap);
-put_heap:
-	amdxdna_gem_put_obj(heap);
 free_priv:
 	kfree(priv);
 	return ret;
@@ -611,7 +610,7 @@ void aie2_hwctx_fini(struct amdxdna_hwctx *hwctx)
 {
 	struct amdxdna_sched_job *job;
 	struct amdxdna_dev *xdna;
-	int idx, i;
+	int idx;
 
 	xdna = hwctx->client->xdna;
 	drm_sched_wqueue_stop(&hwctx->priv->sched);
@@ -640,10 +639,10 @@ void aie2_hwctx_fini(struct amdxdna_hwctx *hwctx)
 	}
 	XDNA_DBG(xdna, "%s sequence number %lld", hwctx->name, hwctx->priv->seq);
 
-	for (i = 0; i < ARRAY_SIZE(hwctx->priv->cmd_buf); i++)
-		drm_gem_object_put(to_gobj(hwctx->priv->cmd_buf[i]));
+	for (idx = 0; idx < ARRAY_SIZE(hwctx->priv->cmd_buf); idx++)
+		drm_gem_object_put(to_gobj(hwctx->priv->cmd_buf[idx]));
 	amdxdna_gem_unpin(hwctx->priv->heap);
-	amdxdna_gem_put_obj(hwctx->priv->heap);
+	drm_gem_object_put(to_gobj(hwctx->priv->heap));
 #ifdef AMDXDNA_DEVEL
 	if (priv_load)
 		aie2_unregister_pdis(hwctx);
--- a/src/driver/amdxdna/aie2_ctx.c
+++ b/src/driver/amdxdna/aie2_ctx.c
@@ -8,8 +8,14 @@
 #include "amdxdna_trace.h"
 #include "aie2_pci.h"
 #include "aie2_solver.h"
+#include "aie2_msg_priv.h"
+
+bool force_cmdlist;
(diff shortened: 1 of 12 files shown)


src/driver/amdxdna/aie2_ctx.c
```diff
@@ -214,22 +220,81 @@ aie2_sched_resp_handler(void *handle, const u32 *data, size_t size)
...
-		job->cmd->state = ERT_CMD_STATE_COMPLETED;
-	XDNA_DBG(job->hwctx->client->xdna, "status 0x%x", status);
+		amdxdna_cmd_set_state(job, 0, ERT_CMD_STATE_ERROR);
+
+out:
+	dma_fence_signal(job->fence);
+	trace_xdna_job(job->hwctx->name, "signaled fence", job->seq);
+	dma_fence_put(job->fence);
+	mmput(job->mm);
+	amdxdna_job_put(job);
```
maxzhen: create a function for these code? May be same for all notification handlers?

makslevental:
Just curious: is this the same cmd list [aie-rt/driver/src/io_backend/ext/xaie_linux.c#L1594-L1595](https://github.com/Xilinx/aie-rt/blob/1ad203de0b7f282b1c0659fd2ae3f218652c7274/driver/src/io_backend/ext/xaie_linux.c#L1594-L1595)?

mamin506:
> Just curious: is this the same cmd list [aie-rt/driver/src/io_backend/ext/xaie_linux.c#L1594-L1595](https://github.com/Xilinx/aie-rt/blob/1ad203de0b7f282b1c0659fd2ae3f218652c7274/driver/src/io_backend/ext/xaie_linux.c#L1594-L1595)?

I don't familiar with what driver. This command list implementation is different.

makslevental:
> > Just curious: is this the same cmd list [aie-rt/driver/src/io_backend/ext/xaie_linux.c#L1594-L1595](https://github.com/Xilinx/aie-rt/blob/1ad203de0b7f282b1c0659fd2ae3f218652c7274/driver/src/io_backend/ext/xaie_linux.c#L1594-L1595)?
> 
> I don't familiar with what driver. This command list implementation is different.

Is there an example somewhere?

src/driver/amdxdna/amdxdna_ctx.c
```diff
@@ -477,6 +477,13 @@ static int amdxdna_cmds_submit(struct amdxdna_client *client,
 unlock_srcu:
 	srcu_read_unlock(&client->hwctx_srcu, idx);
 	amdxdna_arg_bos_put(job);
+put_cmd_bo:
+	for (i = 0; i < args->cmd_bo_count; i++) {
+		if (!cmd_bo[i])
+			continue;
+		drm_gem_object_put(to_gobj(cmd_bo[i]));
```
maxzhen: amdxdna_gem_put_obj(), please, since it is created from amdxdna_gem_get_obj().

src/driver/amdxdna/amdxdna_gem.c
```diff
@@ -213,6 +258,29 @@ amdxdna_drm_create_dev_heap(struct drm_device *dev,
...
+void amdxdna_gem_unresv_heap_mem(struct amdxdna_gem_obj *heap, struct drm_mm_node *node,
+				 struct amdxdna_mem *mem)
+{
+	struct amdxdna_client *client = heap->client;
+
+	mutex_lock(&client->mm_lock);
+	amdxdna_gem_remove_node_nolock(node, mem);
+	mutex_unlock(&client->mm_lock);
+}
+
```
maxzhen: Why do we need these helpers? Can't we just alloc dev bo for each and every single one of the command slot?

maxzhen (changes requested):

src/driver/amdxdna/aie2_ctx.c
mamin506: I think this will be same for all possible handlers. Fixed

src/driver/amdxdna/amdxdna_gem.c
mamin506: We don't really need to allocate drm_gem_object and allocate handle for the reserved memory for command list. Right?
The hwctx will hold ref of the dev heap. It is safe to insert node when hwctx is alive. Thus, I choice this implementation.

Actually, you remind me that I forgot to change amdxdna_drm_alloc_dev_bo() to call amdxdna_gem_insert_node_nolock().

src/driver/amdxdna/amdxdna_ctx.c
mamin506: fixed.

src/driver/amdxdna/amdxdna_gem.c
maxzhen: > We don't really need to allocate drm_gem_object and allocate handle for the reserved memory for command list. Right? The hwctx will hold ref of the dev heap. It is safe to insert node when hwctx is alive. Thus, I choice this implementation.
> 
> Actually, you remind me that I forgot to change amdxdna_drm_alloc_dev_bo() to call amdxdna_gem_insert_node_nolock().

Would it hurt if we just allocate normal device BO (with drm_gem_object and handle), even though we don't have to?

src/driver/amdxdna/amdxdna_gem.c
```diff
@@ -76,14 +127,34 @@ amdxdna_user_mem_fini(struct amdxdna_mem *mem)
 	memset(mem, 0, sizeof(*mem));
 }
 
+static void
+amdxdna_drm_init_bo(struct amdxdna_client *client, struct amdxdna_gem_obj *abo,
```
maxzhen: It seems that you don't need this function. Everything inside it can be folded into amdxdna_gem_create_object() with 'funcs' being passed from caller as an argument.

src/driver/amdxdna/amdxdna_gem.c
```diff
@@ -342,59 +435,40 @@ int amdxdna_drm_create_bo_ioctl(struct drm_device *dev, void *data, struct drm_f
...
-		goto err_unlock;
+	if (IS_ERR(abo))
+		return PTR_ERR(abo);
+
+	if (args->type != AMDXDNA_BO_DEV_HEAP) {
+		/* ready to publish object to userspace */
+		ret = drm_gem_handle_create(filp, to_gobj(abo), &args->handle);
+		if (ret)
+			XDNA_ERR(client->xdna, "Create handle failed");
 	}
```
maxzhen: Let's put drm_gem_handle_create() inside each amdxdna_drm_alloc_xxx_bo(). It is just one function call, not much of dup code. Any special case in the code needs careful thought since it might be missed next time when you move code around.

maxzhen (changes requested):

maxzhen (approved):

src/driver/amdxdna/amdxdna_gem.c
mamin506: As discussed offline, we want to keep handle create in the ioctl(). If other part of driver needs to create a object, then just keep object pointer and reference.

src/driver/amdxdna/amdxdna_gem.c
mamin506: fixed

src/driver/amdxdna/amdxdna_gem.c
mamin506: fixed

PR #84, opened 2024-05-14
Check that __counted_by is defined in UAPI header

atgutier:
When using this header in user space runtimes __counted_by may not be defined for all compilers. Check for its definition and define a dummy version if it is not defined to avoid compilation errors.

### DIFF
--- a/src/include/uapi/drm_local/amdxdna_accel.h
+++ b/src/include/uapi/drm_local/amdxdna_accel.h
@@ -8,13 +8,10 @@
 
 #include <drm/drm.h>
 #include <linux/const.h>
+#include <linux/stddef.h>
 
 #if defined(__cplusplus)
 extern "C" {
-#endif
-
-#ifndef __counted_by
-#define __counted_by(cnt)
 #endif
 
 #define AMDXDNA_DRIVER_MAJOR	1
--- a/src/include/uapi/drm_local/amdxdna_accel.h
+++ b/src/include/uapi/drm_local/amdxdna_accel.h
@@ -13,6 +13,10 @@
 extern "C" {
 #endif
 
+#ifndef __counted_by
+#define __counted_by(cnt)
+#endif
+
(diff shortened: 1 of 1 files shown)


src/include/uapi/drm_local/amdxdna_accel.h
```diff
@@ -13,6 +13,10 @@
 extern "C" {
 #endif
 
+#ifndef __counted_by
+#define __counted_by(cnt)
+#endif
+
```
houlz0507: Instead of defining this, you may use #include <linux/stddef.h>.

atgutier: I thought about doing that, but at least on some of my machines __counted_by is not defined yet. Since it is still kind of new, support for it seems spotty.

houlz0507: Which version of linux kernel are you running? I think this is supported in stddef.h after v6.6 kernel.

atgutier: This is the case, for example, in stock Ubuntu 22.04 where the kernel is 6.5.

houlz0507: Did you patch the v6.5 kernel? xdna driver would not be compatible with standard v6.5 kernel.

atgutier: Yes, on my dev machine I have kernel 6.8. But when I run tests my Jenkins server doesn't have the latest kernel. We don't run any tests on our XDNA code yet, but we still need this to compile.

I'll modify this PR to include stddef.h and simply maintain this current hack in my own copy.

houlz0507: Got it. Thanks for fixing this.

houlz0507 (approved):

PR #85, opened 2024-05-15
Sync dbg bo

mamin506:


makslevental:
Just double checking: this is for supporting the `OP_DUMP_REGISTERS` op code in DPU FW correct?

maxzhen (approved):

maxzhen:
> Just double checking: this is for supporting the `OP_DUMP_REGISTERS` op code in DPU FW correct?

Yes. With this change, you can try OP_DUMP_REGISTERS and sync the data back to host.

PR #90, opened 2024-05-15
support exec dpu in cmd list

mamin506:
Only the aie2_cmdlist_fill_slot_cf() code path has been test.

### DIFF
--- a/src/driver/amdxdna/aie2_msg_priv.h
+++ b/src/driver/amdxdna/aie2_msg_priv.h
@@ -343,10 +343,22 @@ struct async_event_msg_resp {
 } __packed;
 
 #define MAX_CHAIN_CMDBUF_SIZE 0x1000
-#define cmd_chain_slot_has_space(offset, payload_size) \
+#define slot_cf_has_space(offset, payload_size) \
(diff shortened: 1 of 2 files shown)


src/driver/amdxdna/aie2_msg_priv.h
```diff
@@ -352,6 +352,18 @@ struct cmd_chain_slot_execbuf_cf {
 	u32 args[] __counted_by(arg_cnt);
 };
 
+#define slot_dpu_has_space(offset, payload_size) \
+	(MAX_CHAIN_CMDBUF_SIZE - ((offset) + (payload_size)) > \
+	 offsetof(struct cmd_chain_slot_dpu, args[0]))
```
maxzhen, line 357: struct cmd_chain_slot_dpu is of variable size, do you want to use 40 DWORD for the calculation?

mamin506: I want to use the accurate size here to optimize the DMA. This macro just make sure the following memcpy() doesn't write out of bounds.

Your comment reminds me, when calling into this macro, the payload_len should minus sizeof(*sd).

maxzhen (approved):

PR #98, opened 2024-05-20
enable xrt_test.cpp

xdavidz:
trim xrt test into a separate file, now only xrt vadd is in this test file.


mamin506 (approved):

maxzhen (changes requested):
The change seems to be incomplete. We can discuss offline.

xdavidz:
> The change seems to be incomplete. We can discuss offline.

Tested shim_test and xrt_test. Both still working.

xrt_test still requires lib_coreutil, I think it is expected. Please let me know if the xrt_test.cpp is expected.

shim_test has dependencies on lib_coreutil by some existing test cases, we can use a spearate PR to clean this up.

maxzhen (approved):

xdavidz:
now it passes without any shim layer lib included from xrt_test

test/xrt_test/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+  xrt_test.cpp
+  )
+
+target_compile_definitions(${XDNA_XRT_TEST} PRIVATE
+  # below macros is required so that i/f defined in ishim.h is
+  # consistent with native xrt implementation
+  XRT_ENABLE_AIE
+  XRT_XDNA_BUILD
+  )
+
```
maxzhen, line 16: These compilation definitions can be removed. They impact shim level API definitions. XRT application does not need to specify them.

PR #100, opened 2024-05-21
support iommu_mode=1

mamin506:
NO PASID mode is useful when bring up a new device


xdavidz (approved):

maxzhen (approved):

keryell:
Does this mean it would work with a vanilla kernel?

PR #110, opened 2024-06-05
add runtime pm

amd-xiaomren:


src/driver/amdxdna/amdxdna_drv.c
```diff
@@ -294,6 +304,10 @@ static int amdxdna_probe(struct pci_dev *pdev, const struct pci_device_id *id)
 		goto failed_dev_fini;
 	}
 
+	pm_runtime_set_autosuspend_delay(dev, 5000);
```
mamin506:
```suggestion
	pm_runtime_set_autosuspend_delay(dev, 5000 /* ms */);
```
or define a macro for better readability.

src/driver/amdxdna/amdxdna_drv.c
```diff
@@ -380,16 +400,39 @@ static int amdxdna_pmops_resume(struct device *dev)
...
 
 	return 0;
 }
 
+static int amdxdna_rpmops_suspend(struct device *dev)
+{
+	struct amdxdna_dev *xdna = pci_get_drvdata(to_pci_dev(dev));
+	int ret;
+
+	ret = amdxdna_pmops_suspend(dev);
```
mamin506:
```suggestion
	mutex_lock(&xdna->dev_lock);
	WARN_ON(!list_empty(&xdna->client_list);
	ret = amdxdna_dev_suspend_nolock(xdna);
	mutex_unlock(&xdna->dev_lock);
```

The rumtime suspend and resume is called when all of the clients are closed. Those there is no need to do hwctx resume.

The amdxdna_dev_suspend_nolock() will call xdna->dev_info->ops->suspend() and this function can be reused by amdxdna_pmops_suspend().

src/driver/amdxdna/amdxdna_drv.c
```diff
@@ -380,16 +400,39 @@ static int amdxdna_pmops_resume(struct device *dev)
...
+	XDNA_DBG(xdna, "Runtime suspend done ret: %d", ret);
+	return ret;
+}
+
+static int amdxdna_rpmops_resume(struct device *dev)
+{
+	struct amdxdna_dev *xdna = pci_get_drvdata(to_pci_dev(dev));
+	int ret;
+
+	ret = amdxdna_pmops_resume(dev);
```
mamin506: See amdxdna_rpmops_suspend(), same suggestion. In this case, we can also keep XDNA_INFO() print when device resume from sleep/hibernate for easy debug purpose.

src/driver/amdxdna/amdxdna_drv.c
```diff
@@ -263,6 +272,7 @@ amdxdna_get_dev_info(struct pci_dev *pdev)
 static int amdxdna_probe(struct pci_dev *pdev, const struct pci_device_id *id)
 {
 	struct amdxdna_dev *xdna;
+	struct device *dev = &pdev->dev;
```
mamin506:
```suggestion
	struct device *dev = &pdev->dev;
	struct amdxdna_dev *xdna;
```
Reverse XMAS tree declarations please.

src/driver/amdxdna/amdxdna_drv.c
```diff
@@ -322,6 +338,10 @@ static void amdxdna_remove(struct pci_dev *pdev)
 {
 	struct amdxdna_dev *xdna = pci_get_drvdata(pdev);
 	struct amdxdna_client *client;
+	struct device *dev = &pdev->dev;
```
mamin506:
```suggestion
	struct amdxdna_dev *xdna = pci_get_drvdata(pdev);
	struct device *dev = &pdev->dev;
	struct amdxdna_client *client;
```

src/driver/amdxdna/amdxdna_drv.c
```diff
@@ -55,6 +56,12 @@ static int amdxdna_drm_open(struct drm_device *ddev, struct drm_file *filp)
 	struct amdxdna_client *client;
 	int ret;
 
+	ret = pm_runtime_resume_and_get(ddev->dev);
+	if (ret < 0) {
```
mamin506:
```suggestion
	if (ret) {
```
Although this function is documented as it will only return 0 or negative error code. But, when ret is somehow positive, the suggestion is a stronger code and good for maintenance.

mamin506 (changes requested):
Overall this looks good to me. Thanks for adding support for runtime suspend and resume. I have some suggestions, please review.

amd-xiaomren:
thanks for your comments,
fixed

src/driver/amdxdna/amdxdna_drv.c
```diff
@@ -55,6 +58,12 @@ static int amdxdna_drm_open(struct drm_device *ddev, struct drm_file *filp)
...
 
+	ret = pm_runtime_resume_and_get(ddev->dev);
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to get rpm, ret %d", ret);
+		return ret;
+	}
+
 	client = kzalloc(sizeof(*client), GFP_KERNEL);
 	if (!client)
 		return -ENOMEM;
```
mamin506: If allocate client failed, it should mark last busy and put autosuspend.

src/driver/amdxdna/amdxdna_drv.c
```diff
@@ -388,8 +423,40 @@ static int amdxdna_pmops_resume(struct device *dev)
...
+{
+	struct amdxdna_dev *xdna = pci_get_drvdata(to_pci_dev(dev));
+	int ret;
+
+	mutex_lock(&xdna->dev_lock);
+	if (!list_empty(&xdna->client_list)) {
+		XDNA_WARN(xdna, "can't suspend due to alive clients");
+		mutex_unlock(&xdna->dev_lock);
+		return -EBUSY;
+	}
```
mamin506:
```suggestion
        WARN_ONCE(!list_empty(&xdna->client_list), "Can't suspend due to alive clients");
```
Based on your design, only after 5 seconds the last client was closed, this rumtime suspend will be called.
Thus here should be a WARN_ON(). If this happen, it is a driver bug that we must fix. No need to return EBUSY.

src/driver/amdxdna/amdxdna_drv.c
```diff
@@ -294,6 +306,10 @@ static int amdxdna_probe(struct pci_dev *pdev, const struct pci_device_id *id)
 		goto failed_dev_fini;
 	}
 
+	pm_runtime_set_autosuspend_delay(dev, AMDXDNA_AUTOSUSPEND_DELAY);
+	pm_runtime_use_autosuspend(dev);
+	pm_runtime_allow(dev);
```
mamin506, line 316: If drm_dev_register() failed, should driver rewind these operations or not?

mamin506 (changes requested):

mamin506 (changes requested):
Reviewed the error handling code path.

src/driver/amdxdna/amdxdna_drv.c
amd-xiaomren: The [comment ](https://elixir.bootlin.com/linux/v6.9.1/source/drivers/base/power/runtime.c#L786) in kernel's src code mentions that the kernel has an optimization for autosuspend case:
during the resume procedure, autosuspend won't be cancelled unless resuming is successful,
and the usage counter won't change if the resuming fails.
So I think there's no need to call put/suspend related API in the error handling path here.

Also, upon going through the drivers/ folder,  it appears that existing codes use the same error handling approach for the pm_runtime_resume_and_get() API.

src/driver/amdxdna/amdxdna_drv.c
amd-xiaomren: use WARN_ON()

src/driver/amdxdna/amdxdna_drv.c
amd-xiaomren: call pm_runtime_forbid() on error.

src/driver/amdxdna/amdxdna_drv.c
mamin506: I mean when "client = kzalloc(sizeof(*client), GFP_KERNEL);" failed, we need to handle the reference counter of autosuspend.

mamin506 (changes requested):
See my reply.

src/driver/amdxdna/amdxdna_drv.c
amd-xiaomren: my bad, I put all my eyes on the green block.
fixed

mamin506 (approved):

PR #111, opened 2024-06-06
Support kernel 6.10

mamin506:


firstwaterer:
So, we no longer need to build [https://github.com/AMD-SW/linux](url) right?

mamin506:
> So, we no longer need to build [https://github.com/AMD-SW/linux](url) right?

That is correct. We were waiting for the official release of 6.10. Then we will update readme.
For now, you can build and test with 6.10 RC kernel.

PR #115, opened 2024-06-10
add cash line flush for umq submit

xdavidz:


src/shim/umq/hwq.cpp
```diff
@@ -202,6 +216,10 @@ fill_slot_and_send(volatile host_queue_packet_t *pkt, void *payload, size_t size
   std::memcpy(data, payload, size);
```
maxzhen, line 219: I believe you need a flush right after memcpy.

src/shim/umq/hwq.cpp
```diff
@@ -202,6 +216,10 @@ fill_slot_and_send(volatile host_queue_packet_t *pkt, void *payload, size_t size
...
   /* Always done as last step. */
   mark_slot_valid(pkt);
+
+  /* flash cache line for non coherence memory */
+  pkt_clflush(pkt);
```
maxzhen: Here you do flush only for the valid bit change, so no need to flush the whole pkg.

maxzhen (changes requested):
where is the change to make write index and read index in different cacheline? we probably should make sure each pkt is also in its own cacheline.

src/shim/umq/hwq.cpp
xdavidz: If we only flush the valid "uint16_t type: 8;", the memcpy(data, ...) could be still not flushed right? So the device can get dirty data?

Flushing the valid bit is easier than flush the whole data. 

I think we'd better have a chat on this.

mamin506: @xdavidz If clflush is for both data and valid bit, let's assume valid bit is flushed to memory but data is still not fully flushed, will the device start access data because of valid bit is set? Is there any race in this case?

xdavidz: > @xdavidz If clflush is for both data and valid bit, let's assume valid bit is flushed to memory but data is still not fully flushed, will the device start access data because of valid bit is set? Is there any race in this case?

I think we could should make sure the date is flushed first;
then we start writing the valid bit second;
finally we should flush the valid bit again;

I will discuss this flow with Max when he comes back online. Otherwise, there is window that data is in the cache but not in the memory.

src/shim/umq/hwq.cpp
xdavidz: yes, added the flush here.

mamin506 (approved):

src/shim/umq/host_queue.h
```diff
@@ -162,6 +162,8 @@ typedef struct
   uint64_t data_address;
   
   // TODO Ready signal?
+  // padding to 64Bytes cache line
```
maxzhen: I thought we need to make sure read_index and write_index are in different cache line, no?

maxzhen (changes requested):

src/shim/umq/host_queue.h
```diff
@@ -157,6 +157,19 @@ typedef struct
...
+   *  to flush the cache of the host queue.
+   *
+   *  Most importantly, the read_index has to be in different cache line
+   *  (64Bytes in linux) than the write_index. Because the read_index might be
+   *  flushed from a different context from kernel driver that is monitoring
+   *  the completed message. While at the same time, the write_index might be 
+   *  being flushed from UMQ.
+   */ 
+  uint64_t padding[6];
+
```
xdavidz, line 172: @maxzhen @mamin506 would the above comments and code change make sense?

mamin506: Instead of calculating padding manually. I would suggest to use compiler attributes.
For example,
uint32_t write_index __attribute__((aligned(64)));

mamin506 (approved):

src/shim/umq/host_queue.h
mamin506: Instead of calculating padding manually. I would suggest to use compiler attributes.
For example,
`uint32_t write_index __attribute((aligned(64)));`

xdavidz: let me try this on the device compiler.

maxzhen: It might be better to insert padding manually. This data structure is shared b/w firmware and host and compiled separately with different compilers.

maxzhen (approved):

src/shim/umq/host_queue.h
xdavidz: > It might be better to insert padding manually. This data structure is shared b/w firmware and host and compiled separately with different compilers.

Asked Brian, he also think using padding manually like this looks ok to him.

PR #125, opened 2024-06-17
add rpm for debugfs

amd-xiaomren:


mamin506 (commented):
Hi @amd-xiaomren , I don't think this is really needed.

To debug or test through debugfs, we can disable autosuspend by running below command,
echo on > /sys/bus/pci/devices/<BDF>/power/control

To enable autosuspend, run
echo auto > /sys/bus/pci/devices/<BDF>/power/control

See this document for more details, https://www.kernel.org/doc/Documentation/usb/power-management.txt

amd-xiaomren:
Hi mamin,
I agree that echo command can work as an approach to use our debugfs.
However I have a concern for other users, that they are likely not aware that the rpm feature is enabled in our driver,
hence the first idea they might have is just to mount the debugfs, then echo to a node (e.g. echo > /.../npuclock).
A concrete case happens for our internal repo, a testcase is just doing an echo to a node,
althouth I can change the testcase to disable rpm first.
Basically I think this change can ease 3rd party users, what's your opinion?

mamin506:
> Hi mamin, I agree that echo command can work as an approach to use our debugfs. However I have a concern for other users, that they are likely not aware that the rpm feature is enabled in our driver, hence the first idea they might have is just to mount the debugfs, then echo to a node (e.g. echo > /.../npuclock). A concrete case happens for our internal repo, a testcase is just doing an echo to a node, althouth I can change the testcase to disable rpm first. Basically I think this change can ease 3rd party users, what's your opinion?

The debugfs is not a production interface for "user". It is just a method to debug some features for AMD internal XDNA driver developers. If anyone use debugfs to test something, our developers should teach them what to do.

If autosuspend is enabled and echo to a debugfs node, will it crash the system? We just need to fix those crash issues in my opinions.

amd-xiaomren:
> If autosuspend is enabled and echo to a debugfs node, will it crash the system? We just need to fix those crash issues in my opinions.

yes it will, for some nodes (e.g. state) that need to do transaction aie2_send_mgmt_msg_wait().
send you our CI test failure log offline.
it's easy to reproduce the crash locally by running:
echo suspend | sudo tee /sys/kernel/debug/accel/0000:c5:00.1/state

maxzhen:
Is it possible for driver to disable auto-pm as long as debugfs is mounted, instead of doing this dance for each debugfs node access?

amd-xiaomren:
Max and Mamin, I understand your concerns,
I'm going to draft this PR first and think about your suggestions or other approaches.

mamin506 (commented):
Hi Xiaoming, overall this change looks good to me.

Can you please try make ringbuf node like below?
AIE2_DBGFS_FOPS(ringbuf, aie2_ringbuf_show, NULL); 
and
AIE2_DBGFS_FILE(ringbuf, 0400),

 If "echo 1 > ringbuf" return permission error, then this is good.
 
 The ringbuf and msg_queue are read only.

amd-xiaomren:
> Hi Xiaoming, overall this change looks good to me.
> 
> Can you please try make ringbuf node like below? AIE2_DBGFS_FOPS(ringbuf, aie2_ringbuf_show, NULL); and AIE2_DBGFS_FILE(ringbuf, 0400),
> 
> If "echo 1 > ringbuf" return permission error, then this is good.
> 
> The ringbuf and msg_queue are read only.

returns Invalid argument

```
ipu@atg-linux-ipu:/home/npu4_test$ echo 1 | sudo tee /sys/kernel/debug/accel/0000:c5:00.1/ringbuf
1
tee: '/sys/kernel/debug/accel/0000:c5:00.1/ringbuf': Invalid argument
```

mamin506 (approved):

maxzhen (approved):

PR #141, opened 2024-06-28
Use raw userptr in aie mem read/write

mamin506:
1. In 6.10, if device supports SVA, the dma_map_sg() might return noncontiguous mapping in this path. Error out if mapping is non-contiguous.
2. Use raw pointer in aie_mem read/write.


maxzhen (approved):

xdavidz (approved):
I think it is ok. Because the dma address should be contiguous, otherwise it is useless.

PR #144, opened 2024-07-02
Include linux/bitfield.h

atgutier:
Compilation currently fails on Ubuntu 22.04.4 because FIELD_GET is not defined in amdxdna_ctx.h. This change includes <linux/bitfield.h> in amdxdna_ctx.h to fix this issue.


hkboujrida:
@atgutier thank you sir, that was the missing step for me to generate the last deb

houlz0507 (approved):

PR #152, opened 2024-07-09
mailbox_polling is needed for debug purpose

xdavidz:


maxzhen (approved):

src/driver/amdxdna/amdxdna_mailbox.c
```diff
@@ -378,6 +386,22 @@ static irqreturn_t mailbox_irq_handler(int irq, void *p)
...
+#ifdef AMDXDNA_DEVEL
+static void mbox_timer(struct timer_list *t)
+{
+	struct mailbox_channel *mb_chann = from_timer(mb_chann, t, event_timer);
+
+	if (!mb_chann) {
+		pr_warn("DZ_ mbox is gone, stop timer!\n");
+		return;
+	}
+
```
mamin506: Do we really need this check? If this happen, it looks more like a bug.

For other functions check "!mb_chann" is because they are called from outside of this file. We don't want every caller to check "!mb_chann".

Right before mb_chann is destroied, we must call timer_delete_sync() to wait for this timer function to finish.

src/driver/amdxdna/amdxdna_mailbox.c
```diff
@@ -45,6 +45,10 @@
 #define MSG_RX_TIMER			200 /* milliseconds */
 #define MAILBOX_NAME			"xdna_mailbox"
 
+int mailbox_polling;
+module_param(mailbox_polling, int, 0644);
+MODULE_PARM_DESC(mailbox_polling, "1:Enable mailbox polling mode");
```
mamin506, line 51:
```suggestion
#ifdef AMDXDNA_DEVEL
int mailbox_polling;
module_param(mailbox_polling, int, 0644);
MODULE_PARM_DESC(mailbox_polling, "1:Enable mailbox polling mode");
#endif
```

src/driver/amdxdna/amdxdna_mailbox.c
```diff
@@ -675,7 +713,14 @@ void xdna_mailbox_stop_channel(struct mailbox_channel *mb_chann)
 		return;
 
 	/* Disalbe an irq and wait. This might sleep. */
+#ifdef AMDXDNA_DEVEL
+	if (mailbox_polling)
+		del_timer_sync(&mb_chann->event_timer);
```
mamin506: Sorry, I was giving a wrong pointer. The linux said "Do not use in new code. Use timer_delete_sync() instead."

```suggestion
		timer_delete_sync(&mb_chann->event_timer);
```

mamin506 (changes requested):

src/driver/amdxdna/amdxdna_mailbox.c
xdavidz: I will change this

src/driver/amdxdna/amdxdna_mailbox.c
xdavidz: Thanks, I have fixed this basically removed this checking.

mamin506 (approved):

PR #165, opened 2024-07-18
bring back physcial address for devel only when IOMMU is disabled

xdavidz:


src/driver/amdxdna/amdxdna_devel.c
```diff
@@ -116,9 +116,14 @@ int amdxdna_mem_map(struct amdxdna_dev *xdna, struct amdxdna_mem *mem)
 	}
 
 	mem->sgt = sgt;
-	mem->dma_addr = sg_dma_address(sgt->sgl);
-
-	XDNA_DBG(xdna, "dma_addr 0x%llx", mem->dma_addr);
+	if (iommu_mode == AMDXDNA_IOMMU_BYPASS) {
+		mem->dma_addr = sg_phys(sgt->sgl);
+		XDNA_DBG(xdna, "iommu_bypass phy_addr 0x%llx", mem->dma_addr);
+		XDNA_DBG(xdna, "sgt off 0x%d", sgt->sgl->offset);
```
mamin506, line 122: How this work? I don't think it will be physical contiguous. Have you test more than 4KB?

xdavidz: Without hugepage, yeah we only have 1 page. And, now we use kernel to create the BO, only first page works.

This is for 1st step to prove the physical address can be accessed by the FW. Maybe we should assert the size has to be within 4k(1 page). However, once we enabled the iommu, then we have to remove the size. 

This is only for bringup.

mamin506: Can we failed this branch when size > 4KB? Then this looks good to me.

xdavidz: @mamin506 actually, I think this is optional, because we have the amdxdna_mem with sgt. For debug hello only, we can hardcode the sg_phys(sgt->sgl) in hello only source code.

The iommu_mode is a global variable, but we don't have the AMDXDNA_IOMMU_BYPASS (2) which is okish in test only code.

Thus, I am closing this PR as will not fix.

PR #166, opened 2024-07-19
device wide timeout detect and recovery

mamin506:
The per hardware context timeout design is not working well. In the real world, a hardware context is not able to know how to set the timeout.

This PR is to build up a device wide timeout detect and recovery mechanism in AMD XDNA driver.
1. A per device periodic timer is setup and running. It will check the status of each hardware context. If no hardware context makes any progress after last check, it will trigger a device wide recovery. All contexts will be destroyed and restarted.
2. The timer expires is determined by a module parameter, timeout_in_sec. User can change it when loading driver. Default is 2.

### DIFF
--- a/src/driver/amdxdna/amdxdna_ctx.h
+++ b/src/driver/amdxdna/amdxdna_ctx.h
@@ -111,6 +111,15 @@ struct amdxdna_hwctx {
 #define HWCTX_STAT_STOP  2
 	u32				status;
 	u32				old_status;
+	/*
+	 * Sequence number for next job; As this is start from 0,
+	 * it can be used as the submitted job counter.
+	 */
+	u64				seq;
+	/* Completed job counter */
+	u64				completed;
+	/* For TDR worker to keep last completed */
+	u64				tdr_last;
 
 	struct amdxdna_qos_info		     qos;
 	struct amdxdna_hwctx_param_config_cu *cus;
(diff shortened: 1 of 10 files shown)


src/driver/amdxdna/amdxdna_ctx.h
```diff
@@ -111,6 +111,15 @@ struct amdxdna_hwctx {
 #define HWCTX_STAT_STOP  2
 	u32				status;
 	u32				old_status;
+	/*
+	 * Sequence number for next job; As this is start from 0,
+	 * it can be used as the submitted job counter.
+	 */
+	u64				seq;
```
maxzhen: "seq" -> "submitted" and remove the comment

src/driver/amdxdna/amdxdna_ctx.h
```diff
@@ -111,6 +111,15 @@ struct amdxdna_hwctx {
...
+	/* Completed job counter */
+	u64				completed;
+	/* For TDR worker to keep last completed */
+	u64				tdr_last;
```
maxzhen: "tdr_last" -> "tdr_last_completed", pls.

maxzhen (approved):
just some nits.

src/driver/amdxdna/amdxdna_ctx.h
mamin506: fixed

src/driver/amdxdna/amdxdna_ctx.h
mamin506: fixed

PR #169, opened 2024-07-22
Fix packaging issue and <drm/drm.h> include directive

xuwd1:
Hello, I'm an Arch User Repository (AUR) packager who is interested in your XDNA accelerator. While building and packaging this driver, I encountered some problems and tried to fix them with this commit. Here are some details:
1. I'm working on endeavourOS (An Archlinux derived distro), and I noticed that the built deb package contains quotes in its name. This is caused by the linux flavor querying logic inside `pkg.cmake` file not removing the quotes.
2. When building this driver, I encountered missing header error, which is triggered by the `<drm/drm.h>` inclusion in the `amdxdna_accel.h` header. This header is used for both compiling the kernel module and the userspace library. While it's fine to include `<drm/drm.h>` in kernel module compilation, this inclusion will cause error when compiling the userspace library with the default compiler setting. Since including kernel header in userspace code is generally considered bad, I implemented a logic to conditionally choose `<drm/drm.h>` or `<libdrm/drm.h>`.
3. Before committing these changes, I tried to check my patch with `checkpatch.pl` as mentioned in `README.md`. However at the first glance I got quite confused with the name `amd-aie`. I think maybe this should have been changed to `xdna-driver`, which is the current name of this repository.

### DIFF
--- a/CMake/pkg.cmake
+++ b/CMake/pkg.cmake
@@ -14,6 +14,7 @@ execute_process(
   )
 execute_process(
   COMMAND awk -F= "$1==\"ID\" {print $2}" /etc/os-release
+  COMMAND tr -d "\""
   OUTPUT_VARIABLE XDNA_CPACK_LINUX_FLAVOR
   OUTPUT_STRIP_TRAILING_WHITESPACE
   )
--- a/README.md
+++ b/README.md
@@ -144,6 +144,6 @@ A: Create a debug DEB package, see above question. Then install debug DEB packag
 ### Checkpatch
 There is a pre-commit script for this purpose.
 ``` bash
-cp amd-aie/tools/pre-commit <root-of-source-tree>/.git/hooks/
+cp xdna-driver/tools/pre-commit <root-of-source-tree>/.git/hooks/
 ```
 `git commit` will reject the commit if error/warning is found, until you make `checkpatch.pl` happy.


maxzhen:
The change 1 and 3 looks sane. But change #2 does not look right. We did the same as other drm drivers do. I don't see any one using drm.h from libdrm directory. Maybe you're missing linux header package or libc package matching the kernel you're using? BTW, you need to use 6.10 kernel source now.

maxzhen (changes requested):
See my comment above.

xuwd1:
To justify my changes and explain in more detail:
1. Yes I'm currently on `6.10.0-1-mainline`. I built this new version of the kernel specifically for this driver.
2. I'm pretty sure it's not because I have missing headers. <drm/drm.h> is a kernel header which should resides in `/usr/lib/modules/$(uname -r)/build/include/uapi/`, which is strictly a kernel header and should not be introduced into the userspace code. I see that in this repo, the reason <drm/drm.h> is used is that the library needs its macro definition and some other struct defs. In such case, <libdrm/drm.h> should be used, which is exactly the reason libdrm exists. 
3. For a more concrete example of libdrm is used as a replacement for the kernel header in userspace code, `xrt/src/runtime_src/core/edge/include/zynq_ioctl.h` is a good case.

maxzhen:
> To justify my changes and explain in more detail:
> 
> 1. Yes I'm currently on `6.10.0-1-mainline`. I built this new version of the kernel specifically for this driver.
> 2. I'm pretty sure it's not because I have missing headers. <drm/drm.h> is a kernel header which should reside in `/usr/lib/modules/$(uname -r)/build/include/uapi/`, and is strictly a kernel header that should not be introduced into the userspace code. I see that in this repo, the reason <drm/drm.h> is used is that the library needs its macro definition and some other struct defs. In such case, <libdrm/drm.h> should be used, which is exactly the reason libdrm exists.
> 3. For a more concrete example of libdrm is used as a replacement for the kernel header in userspace code, `xrt/src/runtime_src/core/edge/include/zynq_ioctl.h` is a good case.

I'm confused. UAPI headers are meant for user space code to use. It is interface between kernel and user code. Also, the user space code in this repository build and run just fine. The user space code also include this header so that it can interact with driver. Not sure why it does not work for you?

xuwd1:
> > To justify my changes and explain in more detail:
> > 
> > 1. Yes I'm currently on `6.10.0-1-mainline`. I built this new version of the kernel specifically for this driver.
> > 2. I'm pretty sure it's not because I have missing headers. <drm/drm.h> is a kernel header which should reside in `/usr/lib/modules/$(uname -r)/build/include/uapi/`, and is strictly a kernel header that should not be introduced into the userspace code. I see that in this repo, the reason <drm/drm.h> is used is that the library needs its macro definition and some other struct defs. In such case, <libdrm/drm.h> should be used, which is exactly the reason libdrm exists.
> > 3. For a more concrete example of libdrm is used as a replacement for the kernel header in userspace code, `xrt/src/runtime_src/core/edge/include/zynq_ioctl.h` is a good case.
> 
> I'm confused. UAPI headers are meant for user space code to use. It is interface between kernel and user code. Also, the user space code in this repository build and run just fine. The user space code also include this header so that it can interact with driver. Not sure why it does not work for you?

The current design will not build for any Arch installation, not just me. I see that uapi interfaces are indeed designed to appear in userspace codes, but only when they leave the kernel header tree, just like <libdrm/drm.h> actually keeps in sync with <drm/drm.h>. Actually, by checking the compiler system include pathes:
```
/usr/lib/gcc/x86_64-pc-linux-gnu/14.1.1/../../../../include/c++/14.1.1
 /usr/lib/gcc/x86_64-pc-linux-gnu/14.1.1/../../../../include/c++/14.1.1/x86_64-pc-linux-gnu
 /usr/lib/gcc/x86_64-pc-linux-gnu/14.1.1/../../../../include/c++/14.1.1/backward
 /usr/lib/gcc/x86_64-pc-linux-gnu/14.1.1/include
 /usr/local/include
 /usr/lib/gcc/x86_64-pc-linux-gnu/14.1.1/include-fixed
 /usr/include
```
You could clearly see that kernel headers are not there. Forcing the compiler to include ".../uapi/include/" would eventually cause the compiler to emit a warning from the kernel source:
```
/usr/lib/modules/6.10.0-1-mainline/build/include/uapi/linux/types.h:10:2: warning: #warning "Attempt to use kernel headers from user space, see https://kernelnewbies.org/KernelHeaders"
```

Anyway, the only reasonable explanation to me is that, unlike Arch, Ubuntu might have pre-copied all those drm headers directly under `/usr/include` and thus you find the code builds well in your production environment. And I would still insist my opinion that the current design should be changed, especially considering that the `xrt` codes have already adopted it.  If you did not like these changes, then I would just separately maintain a patch, and revert the changes made in that header.

mamin506:
Hi @xuwd1 , if you built that kernel, there should be a linux-libc-dev package. You should install that package as well.

After that, you should be able to see,
``` shell
$find /usr/include/ -name drm.h
/usr/include/drm/drm.h
/usr/include/libdrm/drm.h
```

You don't need to include "uapi/" and compile also don't need to aware of this path.
In the user space code, just "#include drm/drm.h".

If you don't see linux-libc-dev package, please follow this step "https://github.com/amd/xdna-driver?tab=readme-ov-file#linux-compilation-and-installation" to build your kernel.

xuwd1:
> Hi @xuwd1 , if you built that kernel, there should be a linux-libc-dev package. You should install that package as well.
> 
> After that, you should be able to see,
> 
> ```shell
> $find /usr/include/ -name drm.h
> /usr/include/drm/drm.h
> /usr/include/libdrm/drm.h
> ```
> 
> You don't need to include "uapi/" and compile also don't need to aware of this path. In the user space code, just "#include drm/drm.h".
> 
> If you don't see linux-libc-dev package, please follow this step "https://github.com/amd/xdna-driver?tab=readme-ov-file#linux-compilation-and-installation" to build your kernel.

Hi. Thanks for your advice, but sadly that is a distro specific thing. Actually, querying the arch database for `drm.h` has the following result:
```
core/linux-headers 6.9.10.arch1-1 [installed]
    usr/lib/modules/6.9.10-arch1-1/build/include/uapi/drm/drm.h
core/linux-lts-headers 6.6.40-1
    usr/lib/modules/6.6.40-1-lts/build/include/uapi/drm/drm.h
extra/aarch64-linux-gnu-linux-api-headers 6.7-1
    usr/aarch64-linux-gnu/include/drm/drm.h
extra/hyprland 0.41.2-2
    usr/include/hyprland/wlroots-hyprland/backend/drm/drm.h
    usr/include/hyprland/wlroots-hyprland/wlr/backend/drm.h
extra/kernel-headers-musl 4.19.88-2
    usr/lib/musl/include/drm/drm.h
extra/libdrm 2.4.122-1 [installed]
    usr/include/libdrm/drm.h
extra/linux-hardened-headers 6.9.9.hardened1-1
    usr/lib/modules/6.9.9-hardened1-1-hardened/build/include/uapi/drm/drm.h
extra/linux-rt-headers 6.7.0.6.realtime1-4
    usr/lib/modules/6.7.0.6.realtime1-4-rt/build/include/uapi/drm/drm.h
extra/linux-rt-lts-headers 6.6.23.28.realtime1-4
    usr/lib/modules/6.6.23.28.realtime1-4-rt-lts/build/include/uapi/drm/drm.h
extra/linux-zen-headers 6.9.10.zen1-1 [installed]
    usr/lib/modules/6.9.10-zen1-1-zen/build/include/uapi/drm/drm.h
extra/mesa 1:24.1.3-1 [installed: 1:24.1.4-2]
    usr/include/d3dadapter/drm.h
extra/riscv64-linux-gnu-linux-api-headers 6.8-1
    usr/riscv64-linux-gnu/include/drm/drm.h
extra/wlroots 0.18.0-1
    usr/include/wlroots-0.18/wlr/backend/drm.h
extra/wlroots0.17 0.17.4-1
    usr/include/wlroots0.17/wlr/backend/drm.h
extra/zig 0.13.0-1
    usr/lib/zig/libc/include/any-linux-any/drm/drm.h
```

xuwd1:
Now reverted the changes made to `amdxdna_accel.h` header, please review the current version. I'll just keep the separate patch in AUR then.

maxzhen (approved):

PR #170, opened 2024-07-23
Add transaction ELF flow test case in shim test

maxzhen:
This PR also refactored shim_test.cpp and split it into smaller files with helper classes added for easy maintenance and extension in the future.


makslevental:
qq: which arch/fw support transaction ELF flow?

maxzhen:
> qq: which arch/fw support transaction ELF flow?

I have tested it on NPU1 and NPU5. NPU4 should work the same as NPU5.
NPU3 does not support it. NPU2 may not support it.

mamin506 (approved):

PR #181, opened 2024-08-01
Fix some bugs

mamin506:
1. Fix IO_PAGE_FAULT issue when IOMMU_MODE=1, the root cause is that FW somehow read  async message buffer but driver map that buffer as device write only. This is a workaround, will double check with FW team to see why they read the buffer.
2. Fix DRM VMA manager not cleanup issue when remove driver.
3. Telemetry buffer needs do clflush to make sure device can access.
4. Add IO PAGE FAULT flag parser and packaged it.

For example, if we see below error in dmesg,
``` shell
AMD-Vi: Event logged [IO_PAGE_FAULT domain=0x001a address=0xfffc8000 flags=0x0050]
```

Copy flags value, then do
``` shell
$ /opt/xilinx/xrt/amdxdna/io_page_fault_flags 0x0050
GN: guest/nested:
        0 = Transaction used a nested address (GPA)
NX: no execute:
        0 = when the upstream transaction lacks a PASID TLP prefix
US: user-supervisor:
        0 = Supervisor privileges were asserted
I: interrupt:
        0 = transaction was a memory request
PR: present:
        1 = transaction was to a page marked as present or (including V=1b in DTE) or interrupt marked as remapped (RemapEn=1)
RW: read-write (only meaningful when PR=1, TR=0, I=0):
        0 = transaction was a read
PE: permission indicator (only meaningful when PR=1):
        1 = peripheral did not have the permissions required to perform the transaction
RZ: reserved bit not zero or invalid level encoding (only meaningful when PR=1):
        0 = I/O page fault was caused by an invalid level encoding
TR: translation:
        0 = transaction that caused the Device Table lookup was a transaction request
```


src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -453,6 +454,7 @@ static int aie2_telemetry(struct seq_file *m, u32 type)
 	if (!buff)
 		return -ENOMEM;
 
+	drm_clflush_virt_range(buff, size); /* device can access */
```
maxzhen, line 457: Any reason why this is not required for other buffers alloc'ed through dma_alloc_noncoherent?

mamin506: This is required. See aie2_error.c. I just forgot to do clflush here. This is a bug fix.

maxzhen: > This is required. See aie2_error.c. I just forgot to do clflush here. This is a bug fix.

OK.

maxzhen (approved):

PR #195, opened 2024-08-11
Add host side command fence support

maxzhen:
This PR adds command fence support:
- fence creation
- fence destroy
- fence export
- fence import
- wait on host side
- signal on host side

The wait/signal from device side needs more work in driver. It will be supported by a follow up PR.


src/include/uapi/drm_local/amdxdna_accel.h
```diff
@@ -528,6 +531,19 @@ struct amdxdna_drm_set_state {
 	__u64 buffer; /* in */
 };
 
+
+/**
+ * struct amdxdna_drm_syncobjs - Signal or wait on array of DRM timelined sync objects.
+ * @handles: Array of handles of sync objects.
+ * @points: Array of time points for each sync objects.
```
mamin506, line 538: @count: ?

mamin506 (approved):
Looks good to me. Just one comment change in amdxdna_accel.h.

src/include/uapi/drm_local/amdxdna_accel.h
maxzhen: Oh, right. I forgot to add comment for "count" argument. I'll add it.

PR #202, opened 2024-08-15
enable move memtile

xdavidz:
This is the shim_test part of move_memtiles which is the few tests that we can run for npu3 devices.


test/shim_test/shim_test.cpp
```diff
@@ -532,6 +533,9 @@ std::vector<test_case> test_list {
   test_case{ "measure no-op kernel throughput listed command",
     TEST_POSITIVE, dev_filter_is_aie2, TEST_io_throughput, { IO_TEST_NOOP_RUN }
   },
+  test_case{ "npu3 shim move memory tiles",
+    TEST_POSITIVE, dev_filter_is_aie4, TEST_shim_umq_memtiles, {}
+  },
```
maxzhen: Do you want to add this to the end of the list so that we don't have to change the test case sequence number?

maxzhen (approved):

test/shim_test/shim_test.cpp
xdavidz: thanks, yes, I changed it to the end of the list.

mamin506 (approved):

PR #204, opened 2024-08-15
enhance mailbox_polling

mamin506:
Enhance the mailbox_polling parameter. 

In before, mailbox_polling is a switch, it switch mailbox as polling base or interrupt base.

After this change,
when mailbox_polling is 0, this is default, mailbox working on interrupt mode.
when mailbox_polling > 0, mailbox will check message in every 'mailbox_polling' miliseconds.
when mailbox_polling < 0, mailbox will be a busy polling. It will call schedule() to yield CPU to avoid soft lock warning. It will sleep when all channels are idle.

``` shell
# Interrupt mode
$ sudo modprobe amdxdna

# Busy polling mode
$ sudo modprobe amdxdna mailbox_polling=-1

# Interval polling mode
$ sudo modprobe amdxdna mailbox_polling=1             # poll every 1 ms
$ sudo modprobe amdxdna mailbox_polling=100        # poll every 100 ms
```

### DIFF
--- a/src/driver/amdxdna/aie2_ctx.c
+++ b/src/driver/amdxdna/aie2_ctx.c
@@ -357,6 +357,7 @@ aie2_sched_job_run(struct drm_sched_job *sched_job)
 	if (!hwctx->priv->mbox_chann)
 		return ERR_PTR(-ENODEV);
 
+	trace_xdna_job(sched_job, hwctx->name, "job run", job->seq);
 	kref_get(&job->refcnt);
 	fence = dma_fence_get(job->fence);
 
@@ -386,7 +387,6 @@ out:
 		mmput(job->mm);
 		fence = ERR_PTR(ret);
 	}
-	trace_xdna_job(sched_job, hwctx->name, "sent to device", job->seq);
 
 	return fence;
 }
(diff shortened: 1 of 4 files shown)


xdavidz (approved):

maxzhen (changes requested):
Linux kernel scheduler will not help here. We're burning CPU power for nothing if there is no outstanding command. We should poll only when we have outstanding commands.

xdavidz:
> Linux kernel scheduler will not help here. We're burning CPU power for nothing if there is no outstanding command. We should poll only when we have outstanding commands.

Maybe the way we use in the XRT scheduler is a good way? Like poll a little while right after submit a command and then wait for interrupt.

mamin506:
> > Linux kernel scheduler will not help here. We're burning CPU power for nothing if there is no outstanding command. We should poll only when we have outstanding commands.
> 
> Maybe the way we use in the XRT scheduler is a good way? Like poll a little while right after submit a command and then wait for interrupt.

This change will keep the interrupt code path unchanged. The interrupts from NPU are not a problem at the moment. Let's do polling + interrupt when interrupt storm happens.

mamin506:
I have updated the implementation of busy polling mode.
1. Create per device kthread, polld, to do busy polling, instead of creating kthread for each hardware context.
2. Once all mbox channels are idle, the polld kthread will sleep and wait.

maxzhen (approved):

PR #208, opened 2024-08-20
Add DPM support as part of QoS

vengutta18:
The PMFW provides an option to set different DPM levels. The driver should be able to handle the required dpm level based on the requested QoS.


src/driver/amdxdna/aie2_smu.c
```diff
@@ -15,6 +15,9 @@
 #define AIE2_SMU_SET_SOFT_DPMLEVEL	0x7
 #define AIE2_SMU_SET_HARD_DPMLEVEL	0x8
 
+#define AIE2_SMU_SET_SOFT_HCLK_DPMLEVEL 0x7
+#define AIE2_SMU_SET_HARD_HCLK_DPMLEVEL 0x8
+
```
mamin506: @vengutta18 These two Macros was defined previously for the S0i3 issue.  Keep the latest one and replace the old one?

src/driver/amdxdna/aie2_smu.c
```diff
@@ -134,6 +159,29 @@ int aie2_smu_set_dpm_level(struct amdxdna_dev_hdl *ndev, u32 dpm_level)
...
+	int ret;
+
+	if (SMU_DPM_MAX(ndev) == 0 || dpm_level > SMU_DPM_MAX(ndev))
+		return 0;
+
+	if (!ndev->priv->smu_rev) {
+		ret = aie2_smu_set_dpm_level_v0(ndev, dpm_level);
+	} else {
+		ret = aie2_smu_set_dpm_level_v1(ndev, dpm_level);
+	}
```
mamin506:
```suggestion
	if (!ndev->priv->smu_rev)
		ret = aie2_smu_set_dpm_level_v0(ndev, dpm_level);
	else
		ret = aie2_smu_set_dpm_level_v1(ndev, dpm_level);
```

See https://www.kernel.org/doc/html/v4.10/process/coding-style.html search "Do not unnecessarily use braces where a single statement will do."

src/driver/amdxdna/aie2_solver.h
```diff
@@ -66,41 +68,20 @@ struct xrs_action_load {
...
-	POWER_LEVEL_0   = 0x1,
-	POWER_LEVEL_1   = 0x2,
-	POWER_LEVEL_2   = 0x3,
-	POWER_LEVEL_3   = 0x4,
-	POWER_LEVEL_4   = 0x5,
-	POWER_LEVEL_5   = 0x6,
-	POWER_LEVEL_6   = 0x7,
-	POWER_LEVEL_7   = 0x8,
-	POWER_LEVEL_NUM,
-};
```
mamin506, line 89: This enum has been moved to aie2_pci.h, which make sense to me.
I aware of that the POWER_LEVEL_MIN was removed in the new definition. Is this expected?

mamin506 (changes requested):
Just two minor changes.

src/driver/amdxdna/aie2_smu.c
vengutta18: Done

src/driver/amdxdna/aie2_smu.c
vengutta18: I think I'll remove the latest ones, these came as part of the rebase

src/driver/amdxdna/aie2_solver.h
vengutta18: I guess we don't need Power_level_min, power_level_min and power_level_0 are same. so I removed it.

mamin506 (approved):

maxzhen (approved):

PR #241, opened 2024-09-04
indirect dpu pkt handling with device firmware host_queue changes

xdavidz:
With this fix, we enable the multiple column request, aka. the remote barrier. During the test, we can see 3 columns received correct patched paddr. We would like to enable this end-to-end test for future development.

Note: the end-to-end test for this case is not fully passed, but we have made the runtime and fimware test ready. @Brian Xu for more details about what is still not working yet.

### DIFF
--- a/src/driver/amdxdna/amdxdna_mailbox.c
+++ b/src/driver/amdxdna/amdxdna_mailbox.c
@@ -487,10 +487,8 @@ static void mailbox_timer(struct timer_list *t)
 
 	/* The timer mimic interrupt. It is good to reuse irq routine */
 	tail = mailbox_get_tailptr(mb_chann, CHAN_RES_I2X);
-	if (tail) {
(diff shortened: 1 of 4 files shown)


src/driver/amdxdna/amdxdna_mailbox.c
```diff
@@ -488,7 +488,7 @@ static void mailbox_timer(struct timer_list *t)
 	/* The timer mimic interrupt. It is good to reuse irq routine */
 	tail = mailbox_get_tailptr(mb_chann, CHAN_RES_I2X);
 	if (tail) {
-		MB_DBG(mb_chann, "Mimic interrupt...");
```
mamin506, line 491: Can we just delete it?

xdavidz: > Can we just delete it?

sure, deleted it.

maxzhen (approved):

PR #245, opened 2024-09-06
Adding buffer size in GET_INFO

ypapadop-amd:
Adding size in amdxdna_drm_get_bo_info. Used when importing via dma-buf.


maxzhen:
User space code can get the size by lseek(fd, 0, SEEK_END), where the fd is the dma buf handle. Is this not enough?

ypapadop-amd:
If that's the acceptable way, then it should be enough.

PR #250, opened 2024-09-11
Support turbo mode for better latency but burn more power on CPU & NPU

mamin506:
1. Mailbox channel support three types: MGMT, USER_NORMAL and USER_POLL
- MGMT channel is interrupt based and good for not critical path.
- USER_NORMAL channel is interrupt based and good for critical path.
- USER_POLL channel is polling based and good for critical path which needs the best latency.
2. Support "xrt-smi configure --pmode turbo"
- Only when no context is opened, change pmode from/to turbo are allowed.
- In turbo mode, all contexts are creating USER_POLL mailbox channel for best latency.  
3. Support hwq::poll_command() in shim
4. Add shim_test for polling

|  test name  |  ID (wait in ioctl)   |   ID (polling)  |
|---------|------------------|-----------|
| latency           | 25 | 41 |
| throughput    | 40 | 42 |
| runlist latency | 39 | 43 |
| runlist throughput | 30 | 44 |

### DIFF
--- a/src/driver/amdxdna/aie2_pm.c
+++ b/src/driver/amdxdna/aie2_pm.c
@@ -33,6 +33,32 @@ static int aie2_pm_clock_gating(struct amdxdna_dev_hdl *ndev, bool enable)
 	return ret;
 }
 
+bool aie2_pm_is_turbo(struct amdxdna_dev_hdl *ndev)
+{
+	return ndev->pw_mode == POWER_MODE_TURBO;
+}
+
+static int aie2_pm_turbo(struct amdxdna_dev_hdl *ndev, bool enable)
+{
+	struct amdxdna_dev *xdna = ndev->xdna;
+	struct amdxdna_client *client;
+	bool clk_gate_on;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+	list_for_each_entry(client, &xdna->client_list, node) {
+		bool empty;
+
+		mutex_lock(&client->hwctx_lock);
+		empty = idr_is_empty(&client->hwctx_idr);
+		mutex_unlock(&client->hwctx_lock);
+		if (!empty)
+			return -EBUSY;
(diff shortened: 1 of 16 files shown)


src/driver/amdxdna/aie2_pm.c
```diff
@@ -41,7 +67,7 @@ int aie2_pm_set_mode(struct amdxdna_dev_hdl *ndev, enum amdxdna_power_mode_type
 	if (ndev->pw_mode == target)
 		return 0;
 
-	if (target == POWER_MODE_LOW || target == POWER_MODE_MEDIUM)
+	if (target != POWER_MODE_DEFAULT && target != POWER_MODE_TURBO)
```
maxzhen: We need to support performance/default/turbo.

test/shim_test/io_test.cpp
```diff
@@ -83,6 +84,22 @@ io_test_init_runlist_cmd(bo* cmd_bo, std::vector<bo*>& cmd_bos)
 }
 
 #define IO_TEST_TIMEOUT 5000 /* millisecond */
+
+void io_test_cmd_wait(hwqueue_handle *hwq, std::shared_ptr<bo> bo)
+{
+    if (io_test_parameters.wait == IO_TEST_POLL_WAIT) {
+        auto start = clk::now();
+        while(!hwq->poll_command(bo->get())) {
+            auto now = clk::now();
```
maxzhen: It might be too heavy to call clk::now() for every poll.

maxzhen (changes requested):

src/driver/amdxdna/aie2_pm.c
mamin506: fixed

test/shim_test/io_test.cpp
mamin506: fixed

maxzhen (approved):

PR #267, opened 2024-09-25
Adding UMQ NOP xrt test case

hlaccabu:


src/shim/device.cpp
```diff
@@ -605,6 +605,9 @@ struct sequence_name
     case xrt_core::query::sequence_name::type::tct_all_column:
       seq_name = "tct_4col.txt";
       break;
+    case xrt_core::query::sequence_name::type::nop:
+      seq_name = "nop.elf";
```
xdavidz: @maxzhen I don't understand why we add nop.elf here.
Could you please review this code too?

hlaccabu: Apologies, I think that was leftover from my work on smi. Let me check on that

hlaccabu: Okay that change is not needed, my mistake I will remove it

xdavidz (approved):

PR #279, opened 2024-10-08
Turn on the SMU before setting the clks

vengutta18:
For NPU1 there are two revisions that uses different DPM tables. So we are using set clk and retrieving it to decide the revision. But, without turning on the SMU we won't be able to read the proper clk values, so fix that up.

### DIFF
--- a/src/driver/amdxdna/aie2_smu.c
+++ b/src/driver/amdxdna/aie2_smu.c
@@ -181,7 +181,8 @@ static int aie2_smu_set_dpm_level(struct amdxdna_dev_hdl *ndev, u32 dpm_level)
 	if (dpm_level == ndev->smu.curr_dpm_level)
 		return 0;
 
-	if (!ndev->priv->smu_rev)
+	/* default is v1 */
+	if (ndev->priv->smu_rev == SMU_REVISION_NPU1)
 		ret = aie2_smu_set_dpm_level_v0(ndev, dpm_level);
 	else
 		ret = aie2_smu_set_dpm_level_v1(ndev, dpm_level);
@@ -211,6 +212,9 @@ int aie2_smu_set_power_on(struct amdxdna_dev_hdl *ndev)
 {
 	int ret;
 
+	if (ndev->smu.power_state == SMU_POWER_ON)
+		return 0;
+
 	ret = aie2_smu_exec(ndev, AIE2_SMU_POWER_ON, 0, NULL);
 	if (ret)
 		return ret;
@@ -222,6 +226,9 @@ int aie2_smu_set_power_on(struct amdxdna_dev_hdl *ndev)
 int aie2_smu_set_power_off(struct amdxdna_dev_hdl *ndev)
 {
 	int ret;
+
+	if (ndev->smu.power_state == SMU_POWER_OFF)
+		return 0;
 
 	ret = aie2_smu_exec(ndev, AIE2_SMU_POWER_OFF, 0, NULL);
 	if (ret)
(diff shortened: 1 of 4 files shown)


gyang1099 (approved):

src/driver/amdxdna/aie2_smu.c
```diff
@@ -218,10 +218,12 @@ int aie2_smu_start(struct amdxdna_dev_hdl *ndev)
 	struct smu *smu = &ndev->smu;
 	int ret;
 
-	ret = aie2_smu_set_power_on(ndev);
-	if (ret) {
-		XDNA_ERR(ndev->xdna, "Power on failed, ret %d", ret);
-		return ret;
+	if (ndev->smu.power_state == SMU_POWER_OFF) {
```
maxzhen: Can we move the check into aie2_smu_set_power_on()? Is there a reason we need to expose it to caller outside of aie2_smu_set_power_on/_off()?

maxzhen (changes requested):

src/driver/amdxdna/aie2_smu.c
```diff
@@ -250,6 +252,12 @@ void aie2_smu_stop(struct amdxdna_dev_hdl *ndev)
 void aie2_smu_setup(struct amdxdna_dev_hdl *ndev)
 {
 	struct smu *smu = &ndev->smu;
+	int ret;
+
+	ret = aie2_smu_set_power_on(ndev);
+	if (ret) {
+		XDNA_ERR(ndev->xdna, "Power on failed, ret %d", ret);
+	}
```
maxzhen: also don't you want to return error here so that we can failed to load driver?

src/driver/amdxdna/aie2_smu.c
vengutta18: I guess that makes sense, I just didn't want to call the power on again in the initial flow, I think your suggestion makes sense too. I'll update this.

src/driver/amdxdna/aie2_smu.c
vengutta18: done,

src/driver/amdxdna/aie2_smu.c
maxzhen: I understand what you want to do and moving the check into aie2_smu_set_power_on() does not change the logic. Then, we can keep smu.power_state internal to  aie2_smu_set_power_on()/_off() so it'll become a cleaner solution.

src/driver/amdxdna/aie2_pci.c
```diff
@@ -583,7 +583,11 @@ static int aie2_init(struct amdxdna_dev *xdna)
 	}
 	xdna->dev_handle = ndev;
 
-	aie2_smu_setup(ndev);
+	ret = aie2_smu_setup(ndev);
+	if (ret) {
+		XDNA_ERR(xdna, "smu setup failed, ret %d", ret);
+		goto disable_sva;
```
maxzhen: don't you need to introduce a new goto label? Now you need to power off npu when you fail in below code, right? To make the code symmetric and easy to maintain, you can add aie2_smu_teardown() to undo what is done in aie2_smu_setup().

vengutta18: The clean up is being taken care in the aie2_smu_setup.

vengutta18: Thanks max for the offline clarification, also I see I can call aie2_smu_set_poweroff to undo what is done in aie2_smu_setup() instead of having a separate function aie2_smu_teardown() which just calls aie2_smu_set_poweroff
.I  hope that's okay

vengutta18: Also do we need to set the clks back to min value during the aie2_smu_teardown path?

maxzhen: If that is safer to power it off, we should do it. Basically, we should undo all the changes done in aie2_smu_setup(). I think aie2_smu_teardown() is needed to clearly undo all of these in the error path **and** needs to be called in driver's remove code path. Anything asymmetric in the code almost always means something is not right.

vengutta18: I have added aie2_smu_teardown() and added the undo functionality of aie2_smu_setup(). 
Onething is I didn't add this aie2_smu_teardown in the driver's remove code path as that's redundant with aie2_hw_stop().
This aie2_hw_stop() is  being  called in aie2_fini() which is setting the dpm level 0 and smu power off.

PR #280, opened 2024-10-09
setting dpm level for performance mode

maxzhen:


src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -214,19 +214,21 @@ static int aie2_state_show(struct seq_file *m, void *unused)
...
-	ret = kstrtouint_from_user(ptr, len, 10, &val);
+	ret = kstrtoint_from_user(ptr, len, 10, &val);
 	if (ret) {
 		XDNA_ERR(ndev->xdna, "Invalid input value: %d", val);
 		return ret;
 	}
 
-	ret = aie2_smu_set_dpm_level(ndev, val);
+	mutex_lock(&ndev->xdna->dev_lock);
+	ret = aie2_smu_set_fixed_dpm_level(ndev, val);
```
vengutta18, line 230: So this "aie2_smu_set_fixed_dpm_level" called in the debugfs. The expectation is that, if we set certain dpm level through debugfs, do we want that to be constant irrespective of the session dpm level that was computed by the resolver? 
What happens to the session that wanted a different dpm level?

vengutta18:
overall LGTM

src/driver/amdxdna/aie2_debugfs.c
maxzhen: Any time fixed dpm level is set (through debug fs or performance mode), resource solver is not allowed to change the dpm level. All sessions will use the fixed one. This is for debug purposes only. When fixed dpm level is not set, the default level will be used which is under control by resource solver.

vengutta18 (approved):

PR #287, opened 2024-10-18
Support for elf test flow and changing umq/hwctx.cpp for throughput t…

hlaccabu:
…est. Dependent on XRT merge for query_requests.h.


maxzhen (commented):
You probably also want to add the elf to WHENCE. We may need to have sub-dir under elf for nop.elf under different platforms in the future...

AMDGithubSCIMAdmin:

xdavidz (changes requested):
@hlaccabu since you add validate.xclbin and nop.elf, can you please add those into the WHENCE for licensing?
https://github.com/amd/xdna-driver/blob/main/WHENCE

Also you need to follow the procedure to file a test. Please ping me for how to test it.

xdavidz:
> > @hlaccabu since you add validate.xclbin and nop.elf, can you please add those into the WHENCE for licensing? https://github.com/amd/xdna-driver/blob/main/WHENCE
> > Also you need to follow the procedure to file a test. Please ping me for how to test it.
> 
> I am waiting on an xrt main branch merge before I test it with amd-aie if that's what you are referring to. Without my PR merge on the XRT submodule this PR won't work. WHENCE file has been updated.

Thanks. I think we should do this in the following sequence based on your suggestions:
1) wait for your xrt main PR merged.
2) test it internally
3) sync xdna-driver xrt submodule to xrt main correct version which contains your PR
4) fully test this.

But, if this PR would not affect any existing feature. Meaning all current test would pass even with this change, then you can go ahead and request merge this into xdna-driver repo for now. We just don't enable this specific test until every other parts are all set to go.

BTW: I still have comments on one of the change.

src/shim/umq/hwctx.cpp
```diff
@@ -11,7 +11,7 @@ namespace shim_xdna {
 
 hw_ctx_umq::
 hw_ctx_umq(const device& device, const xrt::xclbin& xclbin, const xrt::hw_context::qos_type& qos)
-  : hw_ctx(device, qos, std::make_unique<hw_q_umq>(device, 8), xclbin)
+  : hw_ctx(device, qos, std::make_unique<hw_q_umq>(device, 16), xclbin)
```
xdavidz: I understand this 8 -> 16 is a temporary change due to the nop can flood the queue with simnow on slow machine, but 16 might be good enough or not.

We need to think a way to handle this rather than simply double the size. Let's discuss this before take actions.

maxzhen (approved):

xdavidz (approved):

PR #303, opened 2024-11-08
get fence through syncobj

houlz0507:


AMDGithubSCIMAdmin:

maxzhen (approved):

mamin506 (approved):

src/driver/amdxdna/aie2_ctx.c
```diff
@@ -1057,10 +1027,17 @@ int aie2_cmd_submit(struct amdxdna_hwctx *hwctx, struct amdxdna_sched_job *job,
 		return ret;
 	}
 
+	job->chain = dma_fence_chain_alloc();
```
maxzhen: can we use a local variable instead of adding chain to job data structure?

houlz0507: Good point. Will fix it.

PR #304, opened 2024-11-08
make rpm for fedora or other flavors of linux

harrisonvanderbyl:
Idk, it just needs some if statements in there tbh


AMDGithubSCIMAdmin:

harrisonvanderbyl:
There was a couple dependencies I needed to manually find, as they had different names, like kernel-headers and libgudev I think,
I also had to copy the .ko file over and then manually add it with insmod,
image

tools/amdxdna_deps.sh
```diff
@@ -6,6 +6,11 @@
 
 SCRIPT_DIR=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
 
-apt-get install -y jq
+# if has apt-get
+if [ -x "$(command -v apt-get)" ]; then
+    apt-get install -y jq
+elif [ -x "$(command -v dnf)" ]; then
+    dnf install -y jq
+fi
```
ankon, line 14: Seems the other deps script of xrt uses yum instead of dnf (probably for compatibility?)

So, maybe here also use yum?

TheFlagCourier: DNF has been the default package manager since EL 8. How about let's _not_ use it's legacy alias?

lamikr:
I think it's better to add option for building the rpm in addition of deb files instead of replacing the deb-packaging with rpm packaging. This is slightly modified version of original patch that does that. I have tested it on fedora 40.

I am not sure whether we need to change the /lib/firmware/amdnpu to /usr/lib/firmware/amdnpu, so I dropped that part.

[0001-make-rpm-for-fedora-or-other-flavors-of-linux.txt](https://github.com/user-attachments/files/18366494/0001-make-rpm-for-fedora-or-other-flavors-of-linux.txt)

maxzhen:
Finally find some time to incorporate this change as https://github.com/amd/xdna-driver/pull/394. Closing this one since it is obsolete now.

PR #305, opened 2024-11-08
some cleanup of context status; disallowed user submit command when state is init

mamin506:


AMDGithubSCIMAdmin:

maxzhen (approved):

PR #312, opened 2024-11-19
Coverity Scan issues CID-65554, 481888, 48193

hlaccabu:
Fixing small issues found by coverity scan, all located in src/shim/
Uncaught exception in hwctx.cpp, parse warning in bo.cpp, AUTO_CAUSES_COPY in kmq/bo.cpp

### DIFF
--- a/src/shim/bo.cpp
+++ b/src/shim/bo.cpp
@@ -13,7 +13,7 @@ alloc_drm_bo(const shim_xdna::pdev& dev, amdxdna_bo_type type, void* buf, size_t
   amdxdna_drm_create_bo cbo = {
     .vaddr = reinterpret_cast<uintptr_t>(buf),
     .size = size,
-    .type = type,
+    .type = static_cast<uint32_t>(type),
   };
   dev.ioctl(DRM_IOCTL_AMDXDNA_CREATE_BO, &cbo);
   return cbo.handle;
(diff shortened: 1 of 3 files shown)


src/shim/bo.cpp
```diff
@@ -8,7 +8,7 @@
 namespace {
 
 uint32_t
-alloc_drm_bo(const shim_xdna::pdev& dev, amdxdna_bo_type type, void* buf, size_t size)
+alloc_drm_bo(const shim_xdna::pdev& dev, uint32_t type, void* buf, size_t size)
```
xdavidz: I'd keep the amdxdna_bo_type here for this api even it is prviate anonymous namespace.

And cast it to u32 at line16. Does that make sense?

hlaccabu: I tried it, the compiler doesn't like it it won't build that way.

keryell (changes requested):
It would be great to have commit messages that make sense for an open-source project.

src/shim/bo.cpp
xdavidz: > I tried it, the compiler doesn't like it it won't build that way.

this should work:
.type = static_cast<uint32_t>(type),

keryell (approved):
Thanks!

xdavidz (approved):

PR #315, opened 2024-11-26
Adding CERT log_bo metadata

hlaccabu:


src/shim/hwctx.cpp
```diff
@@ -244,14 +244,30 @@ delete_ctx_on_device()
   fini_log_buf();
 }
 
+void
+hw_ctx::
+set_metadata()
+{
+  m_metadata.magic_no = 0xBAADBEEF;
```
xdavidz: what about "43 45 52 54 4C 4F 47 53" == CERTLOGS

src/shim/hwctx.cpp
```diff
@@ -244,14 +244,30 @@ delete_ctx_on_device()
...
+  m_metadata.major = 1;
+  m_metadata.minor = 4;
```
xdavidz: this doesn't have to be the same as mailbox. let's set it to 0.1
major 0
minor 1

src/shim/hwctx.cpp
```diff
@@ -244,14 +244,30 @@ delete_ctx_on_device()
...
+  m_metadata.num_cols = m_num_cols;
+  for (int i = 0; i < m_num_cols; i++) {
+    m_metadata.col_offset[i] = 1024 * i;
```
xdavidz: I'd think both the num_cols and size can be the arguments, thus the code can be a little dynamic here. for example:
set_metadata(num_cols, size) {
...
for (... i < num_cols; ...)
   col_size[i] = size;
   col_offset[i] = size * i;
...
}

hlaccabu: What should I make the scope of this variable? Because m_num_cols is a private global variable here.

xdavidz: > What should I make the scope of this variable? Because m_num_cols is a private global variable here.

I'd prefer make the num_cols as an argument, and each time you pass the m_num_cols into the set_metadata
for example:
caller to call:
set_metadata(m_num_cols, size)

declare the func like:
void set_metadata(int num_cols, size_t size)

Basically, I want the set_metadata can be self-contained (not rely on global variables).

src/shim/hwctx.cpp
```diff
@@ -244,14 +244,31 @@ delete_ctx_on_device()
   fini_log_buf();
 }
 
+void
+hw_ctx::
+set_metadata(int num_cols, size_t size)
+{
+  m_metadata.magic_no = 0x434552544C4F4753;
```
xdavidz: This seems beyond the size of u32 (0xFFFFFFFF).

src/shim/hwctx.cpp
```diff
@@ -244,14 +244,31 @@ delete_ctx_on_device()
...
+  m_metadata.major = 0;
+  m_metadata.minor = 1;
+  m_metadata.num_cols = num_cols;
+  for (int i = 0; i < num_cols; i++) {
+    m_metadata.col_offset[i] = size * i;
```
xdavidz: There are 2 additional thoughts here:
1). the offset should be considered the metadata size at the offset 0x0, then the real data. By anytime, the metadata should not be corrupted by the data. Thus the first offset should be the sizeof(metadata) + i * size.
2). I'd think we are able to add device (absolute dma address) instead of offset, thus the firmware doesn't have to be recalculate the address. I will update the design.

hlaccabu: Solved 1), we'll discuss 2)

src/shim/hwctx.h
```diff
@@ -74,6 +74,24 @@ class hw_ctx : public xrt_core::hwctx_handle
     std::vector<uint8_t> m_pdi;
   };
 
+  enum cert_log_flag {
+    debug_buffer = 0,
+    trace_buffer
+  };
+
+  struct cert_log_metadata {
+    uint32_t magic_no;  // 0x43455254 "CERT"
```
mamin506:
```suggestion
#define CERT_MAGIC_NO 0x43455254 // "CERT"
    uint32_t magic_no;
```

src/shim/hwctx.cpp
```diff
@@ -244,14 +244,33 @@ delete_ctx_on_device()
   fini_log_buf();
 }
 
+void
+hw_ctx::
+set_metadata(int num_cols, size_t size, uint64_t bo_paddr, uint8_t flag)
+{
+  m_metadata.magic_no = 0x43455254;
```
mamin506:
```suggestion
  m_metadata.magic_no = CERT_MAGIC_NO;
```

mamin506 (changes requested):

mamin506 (approved):

PR #320, opened 2024-12-10
fix compilation errors

ThatOneCalculator:
Compilation fails for me on Arch Linux x86_64 without these changes:

- put "DMA_BUF" in quotes when importing module
- include `libdrm/drm.h` instead of `dm/drm.h` depending on `__KERNEL__` context

### DIFF
--- a/src/driver/amdxdna/amdxdna_gem.c
+++ b/src/driver/amdxdna/amdxdna_gem.c
@@ -21,7 +21,7 @@
 
 #define XDNA_MAX_CMD_BO_SIZE	SZ_32K
 
-MODULE_IMPORT_NS(DMA_BUF);
+MODULE_IMPORT_NS("DMA_BUF");
 
 static int
 amdxdna_gem_insert_node_locked(struct amdxdna_gem_obj *abo, bool use_vmap)
--- a/src/driver/amdxdna/amdxdna_gem.c
+++ b/src/driver/amdxdna/amdxdna_gem.c
@@ -21,7 +21,11 @@
 
 #define XDNA_MAX_CMD_BO_SIZE	SZ_32K
 
+#if LINUX_VERSION_CODE < KERNEL_VERSION(6, 13, 0)
+MODULE_IMPORT_NS(DMA_BUF);
+#else
 MODULE_IMPORT_NS("DMA_BUF");
+#endif
 
 static int
 amdxdna_gem_insert_node_locked(struct amdxdna_gem_obj *abo, bool use_vmap)
--- a/src/include/uapi/drm_local/amdxdna_accel.h
+++ b/src/include/uapi/drm_local/amdxdna_accel.h
@@ -6,7 +6,11 @@
 #ifndef AMDXDNA_ACCEL_H_
 #define AMDXDNA_ACCEL_H_
 
+#ifdef __KERNEL__
 #include <drm/drm.h>
+#else
+#include <libdrm/drm.h>
+#endif
 #include <linux/const.h>
 #include <linux/stddef.h>
 


AMDGithubSCIMAdmin:

ThatOneCalculator:
@AMDGithubSCIMAdmin updated this PR with another fix that took me a bit to fully figure out/narrow down (and updated the top comment accordingly, since now there's two changes instead of one).

src/driver/amdxdna/amdxdna_gem.c
```diff
@@ -21,7 +21,7 @@
 
 #define XDNA_MAX_CMD_BO_SIZE	SZ_32K
 
-MODULE_IMPORT_NS(DMA_BUF);
+MODULE_IMPORT_NS("DMA_BUF");
```
houlz0507, line 28: This will fail for compiling with latest linux kernel. Just wondering what linux kernel are you build with? And what error did you see?

ThatOneCalculator: I actually compiled this against Linux 6.13rc2,  the absolute latest available kernel version. 

The error was:
```
[ 38%] Building CXX object xrt/src/runtime_src/core/common/api/CMakeFiles/core_common_api_library_objects.dir/aie/xrt_graph.cpp.o
In file included from /usr/lib/modules/6.13.0-rc2-1-mainline-um5606-dirty/build/include/linux/module.h:22,
                 from /usr/lib/modules/6.13.0-rc2-1-mainline-um5606-dirty/build/include/linux/device/driver.h:21,
                 from /usr/lib/modules/6.13.0-rc2-1-mainline-um5606-dirty/build/include/linux/device.h:32,
                 from /usr/lib/modules/6.13.0-rc2-1-mainline-um5606-dirty/build/include/linux/dma-mapping.h:5,
                 from /usr/lib/modules/6.13.0-rc2-1-mainline-um5606-dirty/build/include/linux/dma-buf.h:21,
                 from amdxdna_gem.c:7:
amdxdna_gem.c:24:18: error: expected ‘,’ or ‘;’ before ‘DMA_BUF’
   24 | MODULE_IMPORT_NS(DMA_BUF);
      |                  ^~~~~~~
/usr/lib/modules/6.13.0-rc2-1-mainline-um5606-dirty/build/include/linux/moduleparam.h:26:61: note: in definition of macro ‘__MODULE_INFO’
   26 |                 = __MODULE_INFO_PREFIX __stringify(tag) "=" info
      |                                                             ^~~~
/usr/lib/modules/6.13.0-rc2-1-mainline-um5606-dirty/build/include/linux/module.h:299:33: note: in expansion of macro ‘MODULE_INFO’
  299 | #define MODULE_IMPORT_NS(ns)    MODULE_INFO(import_ns, ns)
      |                                 ^~~~~~~~~~~
amdxdna_gem.c:24:1: note: in expansion of macro ‘MODULE_IMPORT_NS’
   24 | MODULE_IMPORT_NS(DMA_BUF);
      | ^~~~~~~~~~~~~~~~
make[8]: *** [/usr/lib/modules/6.13.0-rc2-1-mainline-um5606-dirty/build/scripts/Makefile.build:194: amdxdna_gem.o] Error 1
```

houlz0507: Ok, you are right. It was DMA_BUF with rc1 and changed to "DMA_BUF" in rc2. In this case, please add something like:
#if LINUX_VERSION_CODE < KERNEL_VERSION(6, 13, 0)
MODULE_IMPORT_NS(DMA_BUF)
#else 
MODULE_IMPORT_NS("DMA_BUF")
#endif

ThatOneCalculator: @houlz0507 added :)

houlz0507 (approved):

houlz0507:
@maxzhen, could you review the drm header change?

maxzhen:
We'd like to use the drm.h from linux-libc-dev since that will for sure match the kernel headers. Could you install linux-libc-dev before you build?

ThatOneCalculator:
That package does not exist on Arch Linux.

ThatOneCalculator:
Also, as far as I'm aware, the first part of that change `#ifdef __KERNEL__` will make sure that `drm/drm.h` is used, where it would be most necessary that `drm.h` matches the exact `drm.h` in the kernel headers.

houlz0507:
> Also, as far as I'm aware, the first part of that change `#ifdef __KERNEL__` will make sure that `drm/drm.h` is used, where it would be most necessary that `drm.h` matches the exact `drm.h` in the kernel headers.

I would suggest to copy the latest drm.h to the same fol

ThatOneCalculator:
Manually copying around header files doesn't seem like the best idea for consistent builds, especially for packing on Arch's AUR, where packages are often built from source

houlz0507:
> Manually copying around header files doesn't seem like the best idea for consistent builds, especially for packing on Arch's AUR, where packages are often built from source

Oh, I clicked wrong button and did not finish my suggestion.  Here is what I think:
1. copy latest drm.h to driver src/include/uapi/drm_local
2. in amdxdna_accel.h: 
#ifdef __KERNEL__
#include <uapi/drm/drm.h>
#else
#include "drm.h"
#endif

The reasons I suggest this are:
1. This does not add dependence package for building user space libraries. 
2.  Linux kernel will not break existing user space API. Thus future kernel updating should be compatible.

The user space app/library is not built on target machine. In another word, the packages does not deliver the user space source code to target machine and build on it.

ThatOneCalculator:
You mean copying the drm.h file from my system into the source code?

Also,

> The user space app/library is not built on target machine.

This isn't necessarily true – for AUR distribution, it's a standard practice to have the user build the user space app in library on their machine. Same with Gentoo.

houlz0507:
> You mean copying the drm.h file from my system into the source code?
yes, from linux source code. Let's see if @maxzhen would agree or not. :)
> 
> Also,
> 
> > The user space app/library is not built on target machine.
> 
> This isn't necessarily true – for AUR distribution, it's a standard practice to have the user build the user space app in library on their machine. Same with Gentoo.
Sure. And the suggested way works fine for both because it does not include drm.h from other packages.

maxzhen:
Agreed that if we make a copy of drm.h, we have one less pkg dependency.

ThatOneCalculator:
Sounds good – will tackle this tomorrow morning!

ThatOneCalculator:
@houlz0507 @maxzhen added!

maxzhen:
@ThatOneCalculator , we reviewed the last change and had some internal discussion. it seems that it is not ideal to check in files not copyright'ed by AMD. we decided to just use libdrm/drm.h as you suggested in your original PR. could you revert your change back? we'll also update our internal build environment after we merge your change. sorry for the back-and-forth discussion.

ThatOneCalculator:
@maxzhen all good! Part of the process lol, reverted the commits pertaining to adding the header files :)

maxzhen (approved):

mamin506:
ok to test

ThatOneCalculator:
Awesome, thank you!

PR #328, opened 2024-12-16
EDGEML-8777: Support firmware log buffer

VinitAmd:
[Why]
  Log buffer support is required to enhance debugging support on NPU stack

[How]
  1. allocate log buffer and share with firmware via 'start event trace' message Config start_event_trace request param, send via mgmt channel. Handle the send buffer resp.
  2. receive notifications about log buffer fullness via interrupt from firmware. Handle channel interrupt process read FW lg buffer and update log metadata head/tail.
  
Signed-off-by: Vinit <vinitkumar.shukla@amd.com>

### DIFF
--- a/src/driver/amdxdna/aie2_debugfs.c
+++ b/src/driver/amdxdna/aie2_debugfs.c
@@ -263,6 +263,41 @@ static int aie2_dpm_level_get(struct seq_file *m, void *unused)
 
 AIE2_DBGFS_FOPS(dpm_level, aie2_dpm_level_get, aie2_dpm_level_set);
 
+static ssize_t aie2_event_trace_write(struct file *file, const char __user *ptr,
+				      size_t len, loff_t *off)
+{
+	struct amdxdna_dev_hdl *ndev = file_to_ndev_rw(file);
+	bool state;
+	int ret;
+
+	ret = kstrtobool_from_user(ptr, len, &state);
+	if (ret) {
+		XDNA_ERR(ndev->xdna, "Invalid input value: %d", state);
+		return ret;
+	}
+
+	mutex_lock(&ndev->xdna->dev_lock);
+	aie2_assign_event_trace_state(ndev, state);
+	mutex_unlock(&ndev->xdna->dev_lock);
+
+	return len;
+}
+
+static int aie2_event_trace_show(struct seq_file *m, void *unused)
+{
+	struct amdxdna_dev_hdl *ndev = m->private;
+
+	if (aie2_is_event_trace_enable(ndev))
+		seq_puts(m, "Event trace is enabled\n");
+	else
+		seq_puts(m, "Event trace is disabled\n"
+						"echo 1 > To enable event trace\n");
+
+	return 0;
+}
+
+AIE2_DBGFS_FOPS(event_trace, aie2_event_trace_show, aie2_event_trace_write);
+
 static int test_case01(struct amdxdna_dev_hdl *ndev)
 {
 	int ret;
@@ -573,6 +608,7 @@ const struct {
 	AIE2_DBGFS_FILE(telemetry_error_info, 0400),
 	AIE2_DBGFS_FILE(telemetry_profiling, 0400),
 	AIE2_DBGFS_FILE(telemetry_debug, 0400),
+	AIE2_DBGFS_FILE(event_trace, 0600),
 };
 
 void aie2_debugfs_init(struct amdxdna_dev *xdna)
--- a/src/driver/amdxdna/aie2_event_trace.c
+++ b/src/driver/amdxdna/aie2_event_trace.c
@@ -1,0 +1,368 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ * Copyright (C) 2025, Advanced Micro Devices, Inc.
+ */
+
+#include <linux/kthread.h>
+#include <linux/kernel.h>
+#include <linux/dma-mapping.h>
+#include <drm/drm_cache.h>
+#include "aie2_msg_priv.h"
+#include "aie2_pci.h"
+#include "amdxdna_trace.h"
+#include "amdxdna_mailbox.h"
+
+struct event_trace_req_buf {
+	struct amdxdna_dev_hdl   *ndev;
+	struct workqueue_struct  *wq;
+	struct work_struct       work;
+	u8                       *kern_log_buf;
+	u8                       *buf;
+	u64                      dram_buffer_address;
+	u64                      resp_timestamp;
+	u64                      sys_start_time;
+	u32                      dram_buffer_size;
+	int                      log_ch_irq;
+	bool                     enabled;
+};
+
+struct trace_event_metadata {
+	u64 tail_offset;
+	u64 head_offset;
+	u32 padding[12];
+};
+
+struct trace_event_log_data {
+	u64 counter;
+	u16 payload_hi;
+	u16 type;
+	u32 payload_low;
+};
+
+static void clear_event_trace_msix(struct amdxdna_dev_hdl *ndev)
+{
+	/* Clear the log buffer interrupt */
+	writel(0, (void *)((u64)ndev->mbox_base + (u64)LOG_BUF_MB_IOHUB_PTR));
+}
+
+static int aie2_is_event_trace_supported_on_dev(struct amdxdna_dev_hdl *ndev)
+{
+	struct pci_dev *pdev = to_pci_dev(ndev->xdna->ddev.dev);
+
+	XDNA_DBG(ndev->xdna, "Dev id: 0x%x, Dev rev: 0x%x\n", pdev->device, pdev->revision);
+	return (pdev->device == 0x17f0 && pdev->revision >= 0x10);
+}
+
+static u32 aie2_get_trace_event_content(struct event_trace_req_buf *trace_req_buf)
+{
+	struct amdxdna_dev_hdl *ndev = trace_req_buf->ndev;
+	struct trace_event_metadata *trace_metadata;
+	u8 *kern_buf = trace_req_buf->kern_log_buf;
+	u8 *sys_buf = trace_req_buf->buf;
+	u32 head_ptr, tail_ptr, tail_ptr_wrap;
+	u32 log_size = 0, offset = 0;
+	u32 total_log_size = 0;
+
+	WARN_ON(LOG_RB_SIZE <= 0);
+	trace_metadata = (struct trace_event_metadata *)(sys_buf + LOG_RB_SIZE);
+	head_ptr = (u32)(trace_metadata->head_offset % LOG_RB_SIZE);
+	tail_ptr = (u32)(trace_metadata->tail_offset);
+	tail_ptr_wrap = tail_ptr % LOG_RB_SIZE;
+
+	/* Update the Ring Buffer head pointer */
+	trace_metadata->head_offset = tail_ptr;
+
+	do {
+		if (tail_ptr_wrap > head_ptr)
+			log_size = tail_ptr_wrap - head_ptr;
+		else if (tail_ptr_wrap < head_ptr)
+			log_size = LOG_RB_SIZE - head_ptr;
+		else
+			return 0;
+
+		if (log_size > LOG_RB_SIZE) {
+			XDNA_ERR(ndev->xdna, "log_size > LOG_RB_SIZE");
+			return 0;
+		}
+		/* Copy the ring buffer content to kernel buffer */
+		memcpy(kern_buf + offset, (u8 *)(sys_buf + head_ptr), log_size);
+
+		offset += log_size;
+		total_log_size += log_size;
+		head_ptr = (head_ptr + log_size) % LOG_RB_SIZE;
+	} while (head_ptr < tail_ptr_wrap);
+
+	return total_log_size;
+}
+
+static void aie2_print_trace_event_log(struct amdxdna_dev_hdl *ndev)
+{
+	struct event_trace_req_buf *trace_req_buf;
+	struct trace_event_log_data *log_content;
+	u64 payload;
+	u32 log_size;
+
+	trace_req_buf = ndev->event_trace_req;
+	log_size = aie2_get_trace_event_content(trace_req_buf);
+	XDNA_DBG(ndev->xdna, "FW log size in bytes %u", log_size);
+
+	if (!log_size) {
+		XDNA_ERR(ndev->xdna, "No log data available");
+		return;
+	}
+
+	char *str = (char *)trace_req_buf->kern_log_buf;
+	char *end = str + log_size;
+	u64 fwTicks;
+
+	trace_req_buf->kern_log_buf[log_size] = 0;
+
+	while (str < end) {
+		log_content = (struct trace_event_log_data *)str;
+		payload = ((u64)log_content->payload_hi << 32) | log_content->payload_low;
+		fwTicks = log_content->counter - trace_req_buf->resp_timestamp;
+		fwTicks = fwTicks / 24 + ndev->event_trace_req->sys_start_time;
+		pr_debug("[NPU]::[%llu] type: 0x%04x payload:0x%016llx",
+			 fwTicks, log_content->type, payload);
+		str += MAX_ONE_TIME_LOG_INFO_LEN;
+	}
+}
+
+static void deffered_logging_work(struct work_struct *work)
+{
+	struct event_trace_req_buf *trace_rq = container_of(work, struct event_trace_req_buf, work);
+
+	aie2_print_trace_event_log(trace_rq->ndev);
+}
+
+static irqreturn_t log_buffer_irq_handler(int irq, void *data)
+{
+	struct amdxdna_dev_hdl *ndev = (struct amdxdna_dev_hdl *)data;
+
+	trace_mbox_irq_handle("LOG_BUFFER", irq);
+	clear_event_trace_msix(ndev);
+	queue_work(ndev->event_trace_req->wq, &ndev->event_trace_req->work);
+	return IRQ_HANDLED;
+}
+
+static int aie2_register_log_buf_irq_hdl(struct amdxdna_dev_hdl *ndev, u32 msi_idx)
+{
+	struct amdxdna_dev *xdna = ndev->xdna;
+	struct event_trace_req_buf *req_buf;
+	int ret;
+
+	req_buf = ndev->event_trace_req;
+	INIT_WORK(&req_buf->work, deffered_logging_work);
+	req_buf->wq = alloc_ordered_workqueue("LOG_BUFFER", 0);
+	if (!req_buf->wq) {
+		XDNA_ERR(xdna, "Failed to allocate workqueue");
+		ret = -ENOMEM;
+		goto free_dma_trace_buf;
+	}
+
+	ret = pci_irq_vector(to_pci_dev(xdna->ddev.dev), msi_idx);
+	if (ret < 0) {
+		XDNA_ERR(xdna, "failed to alloc irq vector %d", ret);
+		goto destroy_wq;
+	}
+	req_buf->log_ch_irq = ret;
+
+	ret = request_irq(req_buf->log_ch_irq, log_buffer_irq_handler, 0, "LOG_BUFFER", ndev);
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to register irq %d ret %d", msi_idx, ret);
+		goto destroy_wq;
+	}
+
+	req_buf->kern_log_buf = kcalloc(TRACE_EVENT_BUF_SIZE, sizeof(u8), GFP_KERNEL);
+	if (!req_buf->kern_log_buf) {
+		ret = -ENOMEM;
+		goto free_irq;
+	}
+	return 0;
+
+free_irq:
+	free_irq(req_buf->log_ch_irq, ndev);
+destroy_wq:
+	destroy_workqueue(req_buf->wq);
+free_dma_trace_buf:
+	dma_free_noncoherent(xdna->ddev.dev, req_buf->dram_buffer_size, req_buf->buf,
+			     (dma_addr_t)req_buf->dram_buffer_address, DMA_BIDIRECTIONAL);
+	req_buf->buf = NULL;
+	req_buf->dram_buffer_address = 0;
+	req_buf->dram_buffer_size = 0;
+	return ret;
+}
+
+static void aie2_deregister_log_buf_irq_hdl(struct amdxdna_dev_hdl *ndev)
+{
+	struct event_trace_req_buf *req_buf = ndev->event_trace_req;
+
+	cancel_work_sync(&req_buf->work);
+	/* print already accumulated FW logs */
+	aie2_print_trace_event_log(ndev);
+	destroy_workqueue(req_buf->wq);
+
+	free_irq(req_buf->log_ch_irq, ndev);
+	kfree(req_buf->kern_log_buf);
+}
+
+static int aie2_event_trace_alloc(struct amdxdna_dev_hdl *ndev)
+{
+	struct event_trace_req_buf *req_buf = ndev->event_trace_req;
+	struct amdxdna_dev *xdna = ndev->xdna;
+
+	req_buf->buf = dma_alloc_noncoherent(xdna->ddev.dev, TRACE_EVENT_BUF_SIZE,
+					     (dma_addr_t *)&req_buf->dram_buffer_address,
+					     DMA_BIDIRECTIONAL, GFP_KERNEL);
+
+	if (!req_buf->buf)
+		return -ENOMEM;
+
+	req_buf->dram_buffer_size = TRACE_EVENT_BUF_SIZE;
+	XDNA_DBG(ndev->xdna, "Start event trace buf addr: 0x%llx size 0x%x",
+		 req_buf->dram_buffer_address, req_buf->dram_buffer_size);
+
+	return 0;
+}
+
+static void aie2_event_trace_free(struct amdxdna_dev_hdl *ndev)
+{
+	struct event_trace_req_buf *req_buf = ndev->event_trace_req;
+	struct amdxdna_dev *xdna = ndev->xdna;
+
+	dma_free_noncoherent(xdna->ddev.dev, req_buf->dram_buffer_size, req_buf->buf,
+			     (dma_addr_t)req_buf->dram_buffer_address, DMA_BIDIRECTIONAL);
+
+	req_buf->buf = NULL;
+	req_buf->dram_buffer_address = 0;
+	req_buf->dram_buffer_size = 0;
+}
+
+static int aie2_start_event_trace_send(struct amdxdna_dev_hdl *ndev)
+{
+	struct event_trace_req_buf *req_buf = ndev->event_trace_req;
+	struct amdxdna_dev *xdna = ndev->xdna;
+	int ret;
+
+	ret = aie2_event_trace_alloc(ndev);
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to allocate log buffer, ret %d", ret);
+		return ret;
+	}
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+	ret = aie2_start_event_trace(ndev, req_buf->dram_buffer_address,
+				     req_buf->dram_buffer_size);
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to start event trace, ret %d", ret);
+		aie2_event_trace_free(ndev);
+		return ret;
+	}
+
+	return 0;
+}
+
+static int aie2_stop_event_trace_send(struct amdxdna_dev_hdl *ndev)
+{
+	struct amdxdna_dev *xdna = ndev->xdna;
+	int ret;
+
+	if (!ndev->event_trace_req) {
+		XDNA_DBG(xdna, "Event tracing is not started");
+		return 0;
+	}
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+	ret = aie2_stop_event_trace(ndev);
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to stop event trace, ret %d", ret);
+		return ret;
+	}
+	aie2_event_trace_free(ndev);
+
+	return 0;
+}
+
+bool aie2_is_event_trace_enable(struct amdxdna_dev_hdl *ndev)
+{
+	if (ndev->event_trace_req)
+		return (ndev->event_trace_req->enabled);
+	return false;
+}
+
+void aie2_set_trace_timestamp(struct amdxdna_dev_hdl *ndev,  struct start_event_trace_resp *resp)
+{
+	ndev->event_trace_req->resp_timestamp = resp->current_timestamp;
+	ndev->event_trace_req->sys_start_time = ktime_get_ns() / 1000; /*Convert ns to us*/
+	aie2_register_log_buf_irq_hdl(ndev, resp->msi_idx);
+}
+
+void aie2_unset_trace_timestamp(struct amdxdna_dev_hdl *ndev)
+{
+	ndev->event_trace_req->resp_timestamp = 0;
+	ndev->event_trace_req->sys_start_time = 0;
+	aie2_deregister_log_buf_irq_hdl(ndev);
+}
+
+void aie2_assign_event_trace_state(struct amdxdna_dev_hdl *ndev, bool state)
+{
+	int err;
+
+	if (!aie2_is_event_trace_supported_on_dev(ndev)) {
+		XDNA_ERR(ndev->xdna, "Event trace is not supported on this device");
+		return;
+	}
+
+	if (!ndev->event_trace_req) {
+		XDNA_DBG(ndev->xdna, "Event trace req buffer is not allocated!");
+		return;
+	}
+
+	if (aie2_is_event_trace_enable(ndev) == state) {
+		XDNA_DBG(ndev->xdna, "Event trace state is already %d", state);
+		return;
+	}
+
+	if (!state) {
+		err = aie2_stop_event_trace_send(ndev);
+		if (err)
+			return;
+
+		goto done;
+	}
+
+	err = aie2_start_event_trace_send(ndev);
+	if (err)
+		return;
+
+done:
+	ndev->event_trace_req->enabled = state;
+	XDNA_DBG(ndev->xdna, "Event trace state: %d", state);
+}
+
+int aie2_event_trace_init(struct amdxdna_dev_hdl *ndev)
+{
+	struct event_trace_req_buf *req_buf;
+
+	req_buf = kzalloc(sizeof(*req_buf), GFP_KERNEL);
+	if (!req_buf)
+		return -ENOMEM;
+
+	req_buf->ndev = ndev;
+	req_buf->enabled = false;
+	ndev->event_trace_req = req_buf;
+
+	return 0;
+}
+
+void aie2_event_trace_fini(struct amdxdna_dev_hdl *ndev)
+{
+	if (!ndev->event_trace_req)
+		return;
+
+	if (aie2_is_event_trace_enable(ndev))
+		aie2_assign_event_trace_state(ndev, false);
+
+	kfree(ndev->event_trace_req);
+	ndev->event_trace_req = NULL;
+}
--- a/src/driver/amdxdna/aie2_message.c
+++ b/src/driver/amdxdna/aie2_message.c
@@ -232,6 +232,40 @@ int aie2_query_firmware_version(struct amdxdna_dev_hdl *ndev,
 	return 0;
 }
 
+int aie2_start_event_trace(struct amdxdna_dev_hdl *ndev, dma_addr_t addr, u32 size)
+{
+	DECLARE_AIE2_MSG(start_event_trace, MSG_OP_START_EVENT_TRACE);
+	int ret;
+
+	req.dram_buffer_address = addr;
+	req.dram_buffer_size = size;
+	req.event_trace_dest = EVENT_TRACE_DEST_DRAM;
+	req.event_trace_categories = 0xFFFFFFFF;
+	req.event_trace_timestamp = EVENT_TRACE_TIMESTAMP_FW_CHRONO;
+
+	XDNA_DBG(ndev->xdna, "send start event trace msg");
+	ret = aie2_send_mgmt_msg_wait(ndev, &msg);
+	if (ret)
+		return ret;
+
+	aie2_set_trace_timestamp(ndev, &resp);
+	return 0;
+}
+
+int aie2_stop_event_trace(struct amdxdna_dev_hdl *ndev)
+{
+	DECLARE_AIE2_MSG(stop_event_trace, MSG_OP_STOP_EVENT_TRACE);
+	int ret;
+
+	XDNA_DBG(ndev->xdna, "send stop event trace msg");
+	ret = aie2_send_mgmt_msg_wait(ndev, &msg);
+	if (ret)
+		return ret;
+
+	aie2_unset_trace_timestamp(ndev);
+	return 0;
+}
+
 int aie2_create_context(struct amdxdna_dev_hdl *ndev, struct amdxdna_ctx *ctx,
 			struct xdna_mailbox_chann_info *info)
 {
--- a/src/driver/amdxdna/aie2_msg_priv.h
+++ b/src/driver/amdxdna/aie2_msg_priv.h
@@ -40,6 +40,8 @@ enum aie2_msg_opcode {
 	MSG_OP_SET_RUNTIME_CONFIG          = 0x10A,
 	MSG_OP_GET_RUNTIME_CONFIG          = 0x10B,
 	MSG_OP_REGISTER_ASYNC_EVENT_MSG    = 0x10C,
+	MSG_OP_START_EVENT_TRACE           = 0x10F,
+	MSG_OP_STOP_EVENT_TRACE            = 0x110,
 	MSG_OP_MAX_DRV_OPCODE,
 	MSG_OP_GET_PROTOCOL_VERSION        = 0x301,
 	MSG_OP_MAX_OPCODE
@@ -379,6 +381,51 @@ struct async_event_msg_resp {
 	enum aie2_msg_status	status;
 	enum async_event_type	type;
 } __packed;
+
+/* Start of event tracing data struct */
+#define TRACE_EVENT_BUF_SIZE				0x2000
+#define TRACE_EVENT_BUF_METADATA_SIZE			0x40
+#define MAX_ONE_TIME_LOG_INFO_LEN			16
+/* FIXME: To be deleted */
+#define MPNPU_IOHUB_INT_27_ALIAS			0xD7008
+#define LOG_BUF_MB_IOHUB_PTR				MPNPU_IOHUB_INT_27_ALIAS
+#define LOG_RB_SIZE	(TRACE_EVENT_BUF_SIZE - TRACE_EVENT_BUF_METADATA_SIZE)
+
+enum event_trace_destination {
+	EVENT_TRACE_DEST_DEBUG_BUS,
+	EVENT_TRACE_DEST_DRAM,
+	EVENT_TRACE_DEST_COUNT
+};
+
+enum event_trace_timestamp {
+	EVENT_TRACE_TIMESTAMP_FW_CHRONO,
+	EVENT_TRACE_TIMESTAMP_CPU_CCOUNT,
+	EVENT_TRACE_TIMESTAMP_COUNT
+};
+
+struct start_event_trace_req {
+	u32 event_trace_categories;
+	enum event_trace_destination event_trace_dest;
+	enum event_trace_timestamp event_trace_timestamp;
+	u64 dram_buffer_address;
+	u32 dram_buffer_size;
+} __packed;
+
+struct start_event_trace_resp {
+	enum aie2_msg_status status;
+	u32 msi_idx;
+	u64 current_timestamp;
+} __packed;
+
+struct stop_event_trace_req {
+	u32 place_holder;
+} __packed;
+
+struct stop_event_trace_resp {
+	enum aie2_msg_status status;
+} __packed;
+
+/* End of event tracing data structs */
 
 #define MAX_CHAIN_CMDBUF_SIZE 0x1000
 #define slot_cf_has_space(offset, payload_size) \
--- a/src/driver/amdxdna/aie2_pci.c
+++ b/src/driver/amdxdna/aie2_pci.c
@@ -625,6 +625,10 @@ skip_pasid:
 		goto async_event_free;
 	}
 
+	ret = aie2_event_trace_init(ndev);
+	if (ret)
+		XDNA_DBG(xdna, "Event trace init failed, ret %d", ret);
+
 	release_firmware(fw);
 	return 0;
 
@@ -647,6 +651,7 @@ static void aie2_fini(struct amdxdna_dev *xdna)
 	struct pci_dev *pdev = to_pci_dev(xdna->ddev.dev);
 	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
 
+	aie2_event_trace_fini(ndev);
 	aie2_hw_stop(xdna);
 	aie2_error_async_events_free(ndev);
 #ifdef AMDXDNA_DEVEL
--- a/src/driver/amdxdna/aie2_pci.h
+++ b/src/driver/amdxdna/aie2_pci.h
@@ -68,6 +68,8 @@
 
 struct amdxdna_ctx_priv;
 struct xrs_action_load;
+struct event_trace_req_buf;
+struct start_event_trace_resp;
 
 enum aie2_smu_reg_idx {
 	SMU_CMD_REG = 0,
@@ -104,7 +106,7 @@ enum psp_reg_idx {
 };
 
 enum dpm_level {
-	DPM_LEVEL_0=0,
+	DPM_LEVEL_0 = 0,
 	DPM_LEVEL_1,
 	DPM_LEVEL_2,
 	DPM_LEVEL_3,
@@ -256,6 +258,7 @@ struct amdxdna_dev_hdl {
 	struct mailbox			*mbox;
 	struct mailbox_channel		*mgmt_chann;
 	struct async_events		*async_events;
+	struct event_trace_req_buf	*event_trace_req;
 
 	u32				dev_status;
 	u32				hwctx_cnt;
@@ -357,6 +360,14 @@ void aie2_error_async_events_free(struct amdxdna_dev_hdl *ndev);
 int aie2_error_async_events_send(struct amdxdna_dev_hdl *ndev);
 int aie2_error_async_msg_thread(void *data);
 
+/* aie2_event_trace.c */
+bool aie2_is_event_trace_enable(struct amdxdna_dev_hdl *ndev);
+int aie2_event_trace_init(struct amdxdna_dev_hdl *ndev);
+void aie2_event_trace_fini(struct amdxdna_dev_hdl *ndev);
+void aie2_set_trace_timestamp(struct amdxdna_dev_hdl *ndev, struct start_event_trace_resp *resp);
+void aie2_unset_trace_timestamp(struct amdxdna_dev_hdl *ndev);
+void aie2_assign_event_trace_state(struct amdxdna_dev_hdl *ndev, bool state);
+
 /* aie2_message.c */
 int aie2_suspend_fw(struct amdxdna_dev_hdl *ndev);
 int aie2_resume_fw(struct amdxdna_dev_hdl *ndev);
@@ -370,6 +381,8 @@ int aie2_query_aie_version(struct amdxdna_dev_hdl *ndev, struct aie_version *ver
 int aie2_query_aie_metadata(struct amdxdna_dev_hdl *ndev, struct aie_metadata *metadata);
 int aie2_query_firmware_version(struct amdxdna_dev_hdl *ndev,
 				struct amdxdna_fw_ver *fw_ver);
+int aie2_start_event_trace(struct amdxdna_dev_hdl *ndev, dma_addr_t addr, u32 size);
+int aie2_stop_event_trace(struct amdxdna_dev_hdl *ndev);
 int aie2_create_context(struct amdxdna_dev_hdl *ndev, struct amdxdna_ctx *ctx,
 			struct xdna_mailbox_chann_info *info);
 int aie2_destroy_context(struct amdxdna_dev_hdl *ndev, struct amdxdna_ctx *ctx);
--- a/src/driver/amdxdna/Kbuild
+++ b/src/driver/amdxdna/Kbuild
@@ -32,6 +32,7 @@ amdxdna-y := \
 	aie2_error.o \
 	aie2_debugfs.o \
 	aie2_message.o \
+	aie2_event_trace.o \
 	aie2_pm.o \
 	aie2_pci.o \
 	npu1_regs.o \


src/driver/amdxdna/aie2_msg_priv.h
```diff
@@ -387,6 +389,53 @@ struct async_event_msg_resp {
...
+};
+
+struct start_event_trace_req {
+    uint32_t event_trace_categories;
+    enum event_trace_destination event_trace_dest;
+    enum event_trace_timestamp event_trace_timestamp;
+	//DRAM log buffer address and size
+    uint64_t dram_buffer_address;
+    uint32_t dram_buffer_size;
+};
```
mamin506:
```suggestion
struct start_event_trace_req {
    uint32_t event_trace_categories;
    enum event_trace_destination event_trace_dest;
    enum event_trace_timestamp event_trace_timestamp;
	//DRAM log buffer address and size
    uint64_t dram_buffer_address;
    uint32_t dram_buffer_size;
} __packed;
```

src/driver/amdxdna/aie2_msg_priv.h
```diff
@@ -387,6 +389,53 @@ struct async_event_msg_resp {
...
+
+struct start_event_trace_resp {
+    enum aie2_msg_status status;
+    uint32_t msi_idx;
+    uint64_t current_timestamp;
+};
```
mamin506:
```suggestion
struct start_event_trace_resp {
    enum aie2_msg_status status;
    uint32_t msi_idx;
    uint64_t current_timestamp;
} __packed;
```

src/driver/amdxdna/aie2_msg_priv.h
```diff
@@ -387,6 +389,53 @@ struct async_event_msg_resp {
...
+    enum aie2_msg_status status;
+};
+
+struct set_event_trace_categories_req {
+     uint32_t event_trace_categories;
+};
+
+struct set_event_trace_categories_resp {
+    enum aie2_msg_status status;
+};
```
mamin506: For all *_req and *_resp struct, "__packed" is needed.

BTW, I don't see "stop_event_trace_req". Did I miss something?

mamin506 (commented):
"__packed" is missing for the request and response structs.

src/driver/amdxdna/aie2_msg_priv.h
vengutta18: @mamin506  I think it's still a Draft PR, Vinit will be posting more updates as he keeps working on it.

maxzhen: Right, no need to review until the PR is migrated to a formal one, which means that:
1. The unit test has passed
2. CI passed (no regression and checkpatch clean)

VinitAmd: Hi team, thanks for review initiation.
Yes, the code is still in daft phase, pushed code for review in case of debug support with global team.
Will add comments if any info required.

VinitAmd: > For all *_req and *_resp struct, "__packed" is needed.
> _Thanks, updated and resp issue is resolved due to size mismatch at FW. _

> BTW, I don't see "stop_event_trace_req". Did I miss something?
> _Will add further._

mamin506:
ok to test

VinitAmd:
=== 3 tests failed on Linux_npu_tests run.
    ipu_suspend :  aie2_state_write:NPU suspend \xf8,\x94\x84\xff\xff\xff\xff\xb8%n\x84\xff\xff\xff\xff\x08\x1c̅\xff\xff\xff\xff failed
    ipu_resume   :  aie2_state_write: NPU resume \xff\xf8,\x94\x84\xff\xff\xff\xff(\xbcw\x8eW\xa8\xff\xff\x08\x1c̅\xff\xff\xff\xff failed
    ipu_check_header_hash
Already ticket is raised for FW team [FWDEV-103771](https://ontrack-internal.amd.com/browse/FWDEV-103771) .

mamin506:
@VinitAmd , please fix the pipeline failure. The driver is not able to load. Please try your driver on a NPU1 machine. Please contact @vengutta18 for the machine if you don't have one.

src/driver/amdxdna/amdxdna_mailbox_helper.h
```diff
@@ -37,6 +37,23 @@ struct xdna_notify {
 		.notify_cb = xdna_msg_cb,				\
 	}
 
+#define DECLARE_XDNA_STOP_EVENT_TRACE_MSG(name, op, status)	\
```
mamin506: This is just a dup of DECLARE_XDNA_MSG_COMMON. 
Delete this new macro and use DECLARE_XDNA_MSG_COMMON.

src/driver/amdxdna/amdxdna_mailbox.h
```diff
@@ -133,6 +134,8 @@ void xdna_mailbox_stop_channel(struct mailbox_channel *mailbox_chann);
 int xdna_mailbox_send_msg(struct mailbox_channel *mailbox_chann,
 			  struct xdna_mailbox_msg *msg, u64 tx_timeout);
 
+void clear_logbuff_irq(uint32_t mailbox_int_addr, struct amdxdna_dev_hdl *ndev);
+
```
mamin506: This looks bad. 
If aie2_event_trace.c needs a way to access register, use writel() directly. Add a mailbox API is insane.

src/driver/amdxdna/amdxdna_mailbox.c
```diff
@@ -493,6 +494,13 @@ static irqreturn_t mailbox_irq_handler(int irq, void *p)
 	return IRQ_HANDLED;
 }
 
+void clear_logbuff_irq(uint32_t mailbox_int_addr, struct amdxdna_dev_hdl *ndev)
+{
+	struct mailbox_channel *mb_chann = ndev->mgmt_chann;
+	/* Clear IOHUB register */
+	mailbox_reg_write(mb_chann, mailbox_int_addr, 0);
+}
+
```
mamin506: Delete this from mailbox please.

src/driver/amdxdna/amdxdna_mailbox.c
```diff
@@ -25,6 +25,7 @@
 #include <linux/irqreturn.h>
 #include "amdxdna_trace.h"
 #include "amdxdna_mailbox.h"
+#include "aie2_pci.h"
```
mamin506: Delete.

mamin506 (changes requested):
I'm not expect any changes for amdxdna_mailbox* files for this feature.
Please cleanup the amdxdna_mailbox* files.

After this is fix, I will review other code. 
Please follow https://github.com/amd/xdna-driver?tab=readme-ov-file#checkpatch to setup your workspace.
The coding style needs to pass checkpatch.pl test.

VinitAmd:
> I'm not expect any changes for amdxdna_mailbox* files for this feature. Please cleanup the amdxdna_mailbox* files.
> 
> After this is fix, I will review other code. Please follow https://github.com/amd/xdna-driver?tab=readme-ov-file#checkpatch to setup your workspace. The coding style needs to pass checkpatch.pl test.

Thanks, I was looking for this info

src/driver/amdxdna/amdxdna_mailbox_helper.h
VinitAmd: There is difference, pls check line no; 24 "struct name##_req req = { 0 };" request is initialized with 0. Whereas ```
struct stop_event_trace_req {
    uint32_t :32;
} __packed; 
```
is place holder cannot be initialized, so need a different definition.

File: src/driver/amdxdna/aie2_msg_priv.h (hunk 1 above)
Comment (VinitAmd) [reply src/driver/amdxdna/aie2_msg_priv.h]: done

File: src/driver/amdxdna/aie2_msg_priv.h (hunk 3 above)
Comment (VinitAmd) [reply src/driver/amdxdna/aie2_msg_priv.h]: done

File: src/driver/amdxdna/amdxdna_mailbox_helper.h (hunk 4 above)
Comment (mamin506) [reply src/driver/amdxdna/amdxdna_mailbox_helper.h]: I don't know why this place holder cannot be initiated. There are some other requests like stop_event_trace_req, just a place holder, in aie2_msg_priv.h. They are all good. Do you see any error if you use DECLARE_XDNA_MSG_COMMON?

Comment (VinitAmd) [reply src/driver/amdxdna/amdxdna_mailbox_helper.h]: amdxdna_mailbox_helper.h:24:41: error: excess elements in struct initializer [-Werror]
   24 |         struct name##_req       req = { 0 }; <--------this is the error because it is trying to initialize "uint32_t :32;" memory, which is placeholder only used for padding.
 Since this definition is from mgmt_msg.h file not modifying struct stop_event_trace_req. Changes applied in definition.
 
 _**"in aie2_msg_priv.h. They are all good"**_ --> there are no anonymous variable in struct other than this event so working fine.

**Comment by VinitAmd** [comment]
> > I'm not expect any changes for amdxdna_mailbox* files for this feature. Please cleanup the amdxdna_mailbox* files.
> > After this is fix, I will review other code. Please follow https://github.com/amd/xdna-driver?tab=readme-ov-file#checkpatch to setup your workspace. The coding style needs to pass checkpatch.pl test.
> 
> Thanks, I was looking for this info

Code formatting done with checkpatch.pl

File: src/driver/amdxdna/aie2_msg_priv.h [hunk 8]
```diff
@@ -387,6 +389,64 @@ struct async_event_msg_resp {
...
+} __packed;
+
+struct start_event_trace_resp {
+	enum aie2_msg_status status;
+	uint32_t msi_idx;
+	uint64_t current_timestamp;
+} __packed;
+
+struct stop_event_trace_req {
+	uint32_t:32;
```
mamin506: Linux reviewer was asking not use bitfield.
```suggestion
	u32		place_holder;
```

src/driver/amdxdna/amdxdna_mailbox_helper.h
mamin506: See my comment for "struct stop_event_trace_req".

src/driver/amdxdna/aie2_msg_priv.h
VinitAmd: updated.

src/driver/amdxdna/amdxdna_mailbox.c
VinitAmd: done

src/driver/amdxdna/amdxdna_mailbox.c
VinitAmd: done

src/driver/amdxdna/amdxdna_mailbox.h
VinitAmd: updated.

src/driver/amdxdna/amdxdna_mailbox_helper.h
VinitAmd: updated.

mamin506:
retest this please

mamin506:
ok to test

mamin506:
retest this please

amd-akshatah:
retest this please

sreedharamd:
ok to test

amd-akshatah:
retest this please

src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -263,6 +263,45 @@ static int aie2_dpm_level_get(struct seq_file *m, void *unused)
 
 AIE2_DBGFS_FOPS(dpm_level, aie2_dpm_level_get, aie2_dpm_level_set);
 
+//write debufs for event trace
```
vengutta18: Kernel style comments please.

src/driver/amdxdna/aie2_pci.h
```diff
@@ -250,11 +250,13 @@ struct amdxdna_dev_hdl {
 	u32				npuclk_freq;
 	u32				hclk_freq;
 	bool				force_preempt_enabled;
+	uint32_t			event_trace_enabled;
```
vengutta18: can it be bool?

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,288 @@
...
+#include <linux/kthread.h>
+#include <linux/kernel.h>
+#include <linux/dma-mapping.h>
+#include <drm/drm_cache.h>
+#include "aie2_msg_priv.h"
+#include "aie2_pci.h"
+#include "amdxdna_trace.h"
+#include "amdxdna_mailbox.h"
+
+uint8_t g_fwLogBuf[TRACE_EVENT_BUFFER_SIZE];
```
vengutta18: Please allocate this using kzalloc and clean up when it's no longer needed. Also variable name to fw_log_buf

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,288 @@
...
+struct trace_event_log_data {
+	uint64_t counter;
+	uint16_t payload_hi;
+	uint16_t type;
+	uint32_t payload_low;
+};
+
+static void clear_event_trace_msix(struct amdxdna_dev_hdl *ndev)
+{
+	// Clear the log buffer interrupt
```
vengutta18: comments ditto, kernel style.

src/driver/amdxdna/aie2_pci.h
```diff
@@ -356,6 +358,17 @@ void aie2_error_async_events_free(struct amdxdna_dev_hdl *ndev);
...
+/* aie2_event.c */
+int aie2_event_trace_alloc(struct amdxdna_dev_hdl *ndev);
+void aie2_event_trace_free(struct amdxdna_dev_hdl *ndev);
+int aie2_stop_event_trace_send(struct amdxdna_dev_hdl *ndev);
+int aie2_stop_event_trace(struct amdxdna_dev_hdl *ndev);
+int aie2_start_event_trace_send(struct amdxdna_dev_hdl *ndev);
+int aie2_start_event_trace(struct amdxdna_dev_hdl *ndev, dma_addr_t addr, u32 size);
+void aie2_set_trace_timestamp(struct amdxdna_dev_hdl *ndev, uint32_t *resp);
+void aie2_assign_event_trace_state(struct amdxdna_dev_hdl *ndev, uint32_t state);
+int aie2_is_event_trace_supported_on_dev(struct amdxdna_dev_hdl *ndev);
```
vengutta18: Do we need all the functions here, only keep the functions that are used by other files,rest of them make it static inside the aie2_event_trace.c

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,288 @@
...
+		return;
+	}
+
+	ndev->event_trace_enabled = state;
+	if (state) {
+		int ret = aie2_start_event_trace_send(ndev);
+
+		if (ret) {
+			XDNA_ERR(ndev->xdna, "Send start event trace failed, ret %d", ret);
+			// Currently this feature is supported on limited HW's,
```
vengutta18: multi-line comment style

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,288 @@
...
+		XDNA_ERR(ndev->xdna, "Event trace is not supported on this device");
+		return;
+	}
+
+	if (ndev->event_trace_enabled == state) {
+		XDNA_DBG(ndev->xdna, "Event trace state is already %d", state);
+		return;
+	}
+
+	ndev->event_trace_enabled = state;
```
vengutta18: Could you assign the state once the trace is enabled successfully,  It could fail at line no. 55!

src/driver/amdxdna/aie2_message.c
```diff
@@ -253,6 +256,28 @@ int aie2_query_firmware_version(struct amdxdna_dev_hdl *ndev,
...
+{
+	DECLARE_AIE2_MSG(start_event_trace, MSG_OP_START_EVENT_TRACE);
+
+	req.dram_buffer_address = addr;
+	req.dram_buffer_size = size;
+	req.event_trace_dest = EVENT_TRACE_DEST_DRAM;
+	req.event_trace_categories = 0xFFFFFFFF;
+	req.event_trace_timestamp = EVENT_TRACE_TIMESTAMP_FW_CHRONO;
+
+	XDNA_INFO(ndev->xdna, "send start event trace msg");
```
vengutta18: make it XDNA_DBG

src/driver/amdxdna/aie2_message.c
```diff
@@ -253,6 +256,28 @@ int aie2_query_firmware_version(struct amdxdna_dev_hdl *ndev,
...
+	return aie2_send_mgmt_msg_wait(ndev, &msg);
+}
+
+int aie2_stop_event_trace(struct amdxdna_dev_hdl *ndev)
+{
+	DECLARE_AIE2_MSG(stop_event_trace, MSG_OP_STOP_EVENT_TRACE);
+
+	XDNA_INFO(ndev->xdna, "send stop event trace msg");
```
vengutta18: ditto

src/driver/amdxdna/aie2_pci.c
```diff
@@ -645,9 +645,15 @@ static void aie2_fini(struct amdxdna_dev *xdna)
...
 	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+	int is_event_trace_supported = aie2_is_event_trace_supported_on_dev(ndev);
+
+	if (ndev->dev_status >= AIE2_DEV_START && is_event_trace_supported)
+		aie2_stop_event_trace_send(ndev);
 
 	aie2_hw_stop(xdna);
 	aie2_error_async_events_free(ndev);
+	if (is_event_trace_supported)
+		aie2_event_trace_free(ndev);
```
vengutta18: Couple of things here, are trace_alloc and trace_free are in sync?
what happens during the driver unload or suspend/resume?

vengutta18 (changes requested):

src/driver/amdxdna/aie2_pci.h
VinitAmd: There may be a option to enable event trace but disable log printing for debug purpose , so kept it uint32.
if we want all the time encode log too added to dmesg, it can be bool

src/driver/amdxdna/aie2_event_trace.c
VinitAmd: is there any guide available for kernel comment style ? checkPatch.pl doesn't fix the problem , it only shows the problem.
if no warning/error while  checkPatch.pl run hard to fix problem without guide.

src/driver/amdxdna/aie2_pci.c
VinitAmd: 1. are trace_alloc and trace_free are in sync?---> yes both are in sync
2. what happens during the driver unload ---> when driver unload , aie2_fini is called and which triggers stop event trace and free even trace resource.
3. suspend/resume -> Currently there is FW issue when PG applied FW loosing event trace buffer information, causing crash in resume ([FWDEV-103771](https://ontrack-internal.amd.com/browse/FWDEV-103771)

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,305 @@
...
+
+	ndev->event_trace_req->resp_timestamp = trace_resp->current_timestamp;
+	ndev->event_trace_req->sys_start_time = ktime_get_ns()/1000; /*Convert ns to us*/
+}
+
+static uint32_t aie2_get_trace_event_content(struct event_trace_req_buf *trace_req_buf, uint8_t *kern_buf)
+{
+	u8 *sysBuf;
+	uint32_t rd_ptr, wr_ptr, wr_ptr_wrap;
+	// Log size is the size of the log content in the trace buff in bytes.
```
vengutta18: Could you fix the comment style in all places?

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,305 @@
...
+	if (rb_size == 0)
+		return 0;
+
+	trace_metadata = (struct trace_event_metadata *)(sysBuf + rb_size);
+	if (!trace_metadata) {
+		XDNA_ERR(ndev->xdna, "FW trace buffer metadata is null!");
+		return 0;
+	}
+
+	// Get the ring buffer read and write pointers, update the ring buffer content size
```
vengutta18: ditto

src/driver/amdxdna/aie2_pci.c
vengutta18: can we just call aie2_assign_event_trace_state with state disable.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,322 @@
...
+		char *end = ((char *)str + log_size);
+
+		ndev->fw_log_buf[log_size] = 0;
+
+		while (str < end) {
+			log_content = (struct trace_event_log_data *)str;
+			payload = (uint64_t)((uint64_t)(log_content->payload_hi) << 32 | log_content->payload_low);
+			fwTicks = log_content->counter - trace_req_buf->resp_timestamp;
+			fwTicks = fwTicks/24 + ndev->event_trace_req->sys_start_time;
+			XDNA_INFO(ndev->xdna, "[%llu][FW] type: 0x%04x payload:0x%016llx", fwTicks, log_content->type, payload);
```
mamin506: Talked with @vengutta18 , this should be XDNA_DBG().

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,322 @@
...
+	if (!ndev) {
+		XDNA_INFO(ndev->xdna, "xdna dev is null !");
+		return IRQ_NONE;
+	}
+
+	trace_mbox_irq_handle("LOG_BUFFER", irq);
+	clear_event_trace_msix(ndev);
+
+	if (ndev->event_trace_enabled)
+		aie2_print_trace_event_log(ndev);
```
mamin506: Do NOT do heavy lifting in an IRQ handler. Use workqueue please.

mamin506 (changes requested):

src/driver/amdxdna/aie2_pci.c
```diff
@@ -647,6 +647,8 @@ static void aie2_fini(struct amdxdna_dev *xdna)
 	struct pci_dev *pdev = to_pci_dev(xdna->ddev.dev);
 	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
 
+	/* disable FW event logging if not stopped */
+	aie2_assign_event_trace_state(ndev, false);
```
vengutta18: aie2_event_trace_fini()

src/driver/amdxdna/aie2_pci.h
```diff
@@ -251,15 +251,18 @@ struct amdxdna_dev_hdl {
...
 	/* Mailbox and the management channel */
 	struct mailbox			*mbox;
 	struct mailbox_channel		*mgmt_chann;
 	struct async_events		*async_events;
+	struct event_trace_req_buf	*event_trace_req;
 
 	u32				dev_status;
 	u32				hwctx_cnt;
 	u32				hwctx_limit;
+	uint8_t				*fw_log_buf;
```
vengutta18: we can have the buffer inside the event_trace_req

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	XDNA_DBG(ndev->xdna, "Dev id: 0x%x, Dev rev: 0x%x\n", pdev->device, pdev->revision);
+	return (pdev->device == 0x17f0 && pdev->revision >= 0x10);
+}
+
+static uint32_t aie2_get_trace_event_content(struct event_trace_req_buf *trace_req_buf, uint8_t *kern_buf)
+{
+	u8 *sysBuf;
+	uint32_t rd_ptr, wr_ptr, wr_ptr_wrap;
```
vengutta18: u32 instead of uint32_t,
could you replace in all other places

src/driver/amdxdna/aie2_pci.c
vengutta18: You can sync this by calling aie2_event_trace_start in aie2_start() where you do the required initialization 
and clean up in aie2_event_trace_fini()

src/driver/amdxdna/aie2_event_trace.c
vengutta18: Similarly for u64

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,344 @@
...
+			str += MAX_ONE_TIME_LOG_INFO_LEN;
+		}
+	}
+}
+
+static void deffered_logging_work(struct work_struct *work)
+{
+	struct event_trace_req_buf *trace_rq = container_of(work, struct event_trace_req_buf, work);
+
+	if (!trace_rq)
```
vengutta18: Don't need the null checks.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,344 @@
...
+		return;
+
+	aie2_print_trace_event_log(trace_rq->ndev);
+}
+
+static irqreturn_t log_buffer_irq_handler(int irq, void *data)
+{
+	struct amdxdna_dev_hdl *ndev = (struct amdxdna_dev_hdl *)data;
+
+	if (!ndev) {
```
vengutta18: Ditto, don't need the null checks, if ndev is null this wouldn't have been called.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,344 @@
...
+static void aie2_print_trace_event_log(struct amdxdna_dev_hdl *ndev)
+{
+	uint32_t log_size;
+	uint64_t payload;
+	struct event_trace_req_buf *trace_req_buf;
+	struct trace_event_log_data *log_content;
+
+	trace_req_buf = ndev->event_trace_req;
+
+	if (!trace_req_buf) {
```
vengutta18: don't need null checks. You need null checks probably for the data that's passed externally to you.
This is all the driver data, so you don't need null checks.

Also, you check only when you are allocating, in all the remaining places it is supposed to exist, otherwise something wrong with the code, and let the kernel OOPS it so that you can see the actual issue.
If you just return it hides the real issue.

src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -263,6 +263,40 @@ static int aie2_dpm_level_get(struct seq_file *m, void *unused)
...
 AIE2_DBGFS_FOPS(dpm_level, aie2_dpm_level_get, aie2_dpm_level_set);
 
+static ssize_t aie2_event_trace_write(struct file *file, const char __user *ptr,
+				size_t len, loff_t *off)
+{
+	struct amdxdna_dev_hdl *ndev = file_to_ndev_rw(file);
+	u32 state;
+	int ret;
+
+	ret = kstrtouint_from_user(ptr, len, 2, &state);
```
vengutta18: state is being passed as boolean in line no:280,
So could you update the variable to
bool state;
you could use: kstrtobool_from_user(ptr, len, &state);

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,342 @@
...
+{
+	/* Clear the log buffer interrupt */
+	writel(0, (void *)((u64)ndev->mbox_base + (u64)LOG_BUF_MB_IOHUB_PTR));
+}
+
+static int aie2_is_event_trace_supported_on_dev(struct amdxdna_dev_hdl *ndev)
+{
+	struct pci_dev *pdev = to_pci_dev(ndev->xdna->ddev.dev);
+
+	if (!pdev) {
```
vengutta18: I don't think this check is necessary. pdev won't be NULL here.
if pdev is really NULL, Let it crash and fix the issue.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,339 @@
...
+	} else {
+		XDNA_ERR(xdna, "Failed to allocate and register event trace");
+	}
+
+	return ret;
+}
+
+void aie2_assign_event_trace_state(struct amdxdna_dev_hdl *ndev, bool state)
+{
+	if (!ndev) {
```
vengutta18: ditto

vengutta18 (commented):
Minor comments.

src/driver/amdxdna/aie2_message.c
```diff
@@ -43,6 +43,9 @@ aie2_send_mgmt_msg_wait_offset(struct amdxdna_dev_hdl *ndev,
 		XDNA_ERR(xdna, "command opcode 0x%x failed, status 0x%x",
 			 msg->opcode, *hdl->data);
 		ret = -EINVAL;
+	} else {
+		if (msg->opcode == MSG_OP_START_EVENT_TRACE)
+			aie2_set_trace_timestamp(ndev, hdl->data);
 	}
```
mamin506: This is the common code path, not the place to handle an individual OPCODE.
Please move this logic to aie2_start_event_trace().

src/driver/amdxdna/aie2_message.c
```diff
@@ -232,6 +235,28 @@ int aie2_query_firmware_version(struct amdxdna_dev_hdl *ndev,
...
+	DECLARE_AIE2_MSG(start_event_trace, MSG_OP_START_EVENT_TRACE);
+
+	req.dram_buffer_address = addr;
+	req.dram_buffer_size = size;
+	req.event_trace_dest = EVENT_TRACE_DEST_DRAM;
+	req.event_trace_categories = 0xFFFFFFFF;
+	req.event_trace_timestamp = EVENT_TRACE_TIMESTAMP_FW_CHRONO;
+
+	XDNA_DBG(ndev->xdna, "send start event trace msg");
+	return aie2_send_mgmt_msg_wait(ndev, &msg);
```
mamin506:
```suggestion
	ret = aie2_send_mgmt_msg_wait(ndev, &msg);
	if (ret)
	    return ret;
	    
	aie2_set_trace_timestamp(ndev, &resp);
	return 0;
```

src/driver/amdxdna/aie2_msg_priv.h
```diff
@@ -40,6 +40,8 @@ enum aie2_msg_opcode {
 	MSG_OP_SET_RUNTIME_CONFIG          = 0x10A,
 	MSG_OP_GET_RUNTIME_CONFIG          = 0x10B,
 	MSG_OP_REGISTER_ASYNC_EVENT_MSG    = 0x10C,
+	MSG_OP_START_EVENT_TRACE          = 0x10F,
+	MSG_OP_STOP_EVENT_TRACE          = 0x110,
```
mamin506: Alignment for "=".

src/driver/amdxdna/aie2_msg_priv.h
```diff
@@ -380,6 +382,71 @@ struct async_event_msg_resp {
...
 } __packed;
 
+/* Start of event tracing data struct */
+
+#define LOG_BUFFER_IRQ 27
+#define TRACE_EVENT_BUFFER_SIZE	0x2000
+#define TRACE_EVENT_BUFFER_METADATA_SIZE	0x40
+#define MAX_ONE_TIME_LOG_INFO_LEN	16
+#define MPIPU_IOHUB_INT_27_ALIAS 0xD7008
+#define LOG_BUF_MB_IOHUB_PTR MPIPU_IOHUB_INT_27_ALIAS
```
mamin506: MPIPU_IOHUB_INT_27_ALIAS and LOG_BUFFER_IRQ are device, npu4, specified. But aie2_msg_priv.h is the place for driver/firmware interface. Please do below changes,

1. Rename "MPIPU_" to "MPNPU_".
2. Move LOG_BUFFER_IRQ and MPNPU_IOHUB_INT_27_ALIAS definitions to npu1_regs.c and npu4_family.h
3. MPNPU_IOHUB_INT_27_ALIAS is the offset of a register. Please ref npu1_regs.c and npu4_family.h to see how to define registers address. (The absolute register address is required for legal review and scan)
4. Extend "struct amdxdna_dev_priv" to store these device static data. For example, you can do,

In aie2_pci.h,

struct aie2_log_buf {
    u32     msix_id;
    u32     iohub_reg_off;
};

struct amdxdna_dev_priv {
...
    struct aie2_log_buf     log_buf;
};

src/driver/amdxdna/aie2_msg_priv.h
```diff
@@ -380,6 +382,71 @@ struct async_event_msg_resp {
...
+} __packed;
+
+struct set_event_trace_categories_req {
+	u32 event_trace_categories;
+};
+
+struct set_event_trace_categories_resp {
+	enum aie2_msg_status status;
+};
+
```
mamin506: These are unused. Delete them please.

src/driver/amdxdna/aie2_msg_priv.h
```diff
@@ -380,6 +382,71 @@ struct async_event_msg_resp {
...
+	struct start_event_trace_req trace_req;
+	struct workqueue_struct *wq;
+	struct work_struct work;
+	u64 resp_timestamp;
+	u64 sys_start_time;
+	int log_ch_irq;
+	u8 *fw_log_buf;
+	u8 *buf;
+};
+
```
mamin506: This struct is for aie2_event_trace.c internal used. It is better to be invisible to other part of the driver.
Move this to aie2_event_trace.c.
In aie2_pci.h, forward declare "struct event_trace_req_buf" is enough.

src/driver/amdxdna/aie2_msg_priv.h
```diff
@@ -380,6 +382,71 @@ struct async_event_msg_resp {
...
+	u32 event_trace_categories;
+};
+
+struct set_event_trace_categories_resp {
+	enum aie2_msg_status status;
+};
+
+struct event_trace_req_buf {
+	struct amdxdna_dev_hdl *ndev;
+	struct start_event_trace_req trace_req;
```
mamin506: You only need dram_buffer_address and dram_buffer_size. Please replace struct start_event_trace_req by them.

vengutta18 (approved):

src/driver/amdxdna/aie2_pci.c
```diff
@@ -625,6 +625,10 @@ static int aie2_init(struct amdxdna_dev *xdna)
 		goto async_event_free;
 	}
 
+	ret = aie2_event_trace_init(ndev);
+	if (ret)
+		XDNA_ERR(xdna, "Event trace init failed, ret %d", ret);
+
```
mamin506: If this error can be ignore, please used XDNA_DBG(). Otherwise, please goto rewind code and return error.
One thing more, if init failed, aie2_assign_event_trace_state() should check ndev->event_trace_req is NULL or not. This is a bug.

src/driver/amdxdna/aie2_pci.h
```diff
@@ -251,11 +251,13 @@ struct amdxdna_dev_hdl {
 	u32				npuclk_freq;
 	u32				hclk_freq;
 	bool				force_preempt_enabled;
+	bool				event_trace_enabled;
```
mamin506: Move this to struct event_trace_req_buf. 
aie2_event_trace.c can provide an function like "bool aie2_is_event_trace_enable(struct amdxdna_dev_hdl *ndev)" to check this value.

src/driver/amdxdna/aie2_pci.h
```diff
@@ -357,6 +359,14 @@ void aie2_error_async_events_free(struct amdxdna_dev_hdl *ndev);
 int aie2_error_async_events_send(struct amdxdna_dev_hdl *ndev);
 int aie2_error_async_msg_thread(void *data);
 
+/* aie2_event.c */
+int aie2_event_trace_init(struct amdxdna_dev_hdl *ndev);
+void aie2_event_trace_fini(struct amdxdna_dev_hdl *ndev);
+int aie2_start_event_trace(struct amdxdna_dev_hdl *ndev, dma_addr_t addr, u32 size);
+int aie2_stop_event_trace(struct amdxdna_dev_hdl *ndev);
```
mamin506: Move to "/* aie2_message.c */" section in this file.

src/driver/amdxdna/aie2_pci.h
```diff
@@ -357,6 +359,14 @@ void aie2_error_async_events_free(struct amdxdna_dev_hdl *ndev);
 int aie2_error_async_events_send(struct amdxdna_dev_hdl *ndev);
 int aie2_error_async_msg_thread(void *data);
 
+/* aie2_event.c */
```
mamin506: aie2_event_trace.c

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,329 @@
...
+		ndev->event_trace_enabled = state;
+	}
+	clear_event_trace_msix(ndev);
+}
+
+int aie2_event_trace_init(struct amdxdna_dev_hdl *ndev)
+{
+	struct event_trace_req_buf *req_buf;
+
+	req_buf = kzalloc(sizeof(struct event_trace_req_buf), GFP_KERNEL);
```
mamin506: Linux suggest coding style for kalloc() kind of functions.
```suggestion
	req_buf = kzalloc(sizeof(*req_buf), GFP_KERNEL);
```

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,329 @@
...
+{
+	if (!ndev->event_trace_req)
+		return;
+
+	if (ndev->event_trace_enabled) {
+		aie2_assign_event_trace_state(ndev, false);
+	} else {
+		kfree(ndev->event_trace_req);
+		ndev->event_trace_req = NULL;
+	}
```
mamin506: Is this a memleak?
```suggestion
	if (ndev->event_trace_enabled) {
		aie2_assign_event_trace_state(ndev, false);
	}
	kfree(ndev->event_trace_req);
	ndev->event_trace_req = NULL;
```

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,329 @@
...
+
+	if (state) {
+		int ret = aie2_start_event_trace_send(ndev);
+
+		if (ret) {
+			XDNA_ERR(ndev->xdna, "Send start event trace failed, ret %d", ret);
+			/*
+			 * Currently this feature is supported on limited HW's,
+			 * driver loading should not fail if FW logging not supported.
+			 */
```
mamin506: Maybe the wrong place for this comment ?? This function is not about driver loading

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,329 @@
...
+			ndev->event_trace_enabled = state;
+			XDNA_DBG(ndev->xdna, "Event trace state: %d", state);
+		}
+	} else {
+		if (ndev->dev_status >= AIE2_DEV_START) {
+			aie2_stop_event_trace_send(ndev);
+			aie2_event_trace_free(ndev);
+		} else {
+			XDNA_DBG(ndev->xdna, "Event trace is not started");
+		}
```
mamin506: As I commented in aie2_pci.c, even AIE2_DEV_START is set. Event trace might still not being initialized.
Here we need to check ndev->event_trace_req, which is allocated in this c file.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,329 @@
...
+					trace_req_buf->trace_req.dram_buffer_size);
+	} else {
+		XDNA_ERR(xdna, "Failed to allocate and register event trace");
+	}
+
+	return ret;
+}
+
+void aie2_assign_event_trace_state(struct amdxdna_dev_hdl *ndev, bool state)
+{
```
mamin506: A general Linux coding style comment.
Linux driver dislike below code block,
``` c
ret = my_foo(arg1, arg2);
if (!ret) {
   // do something in success
} else {
   // handle error
}
...
```

Linux suggested coding style.
``` c
ret = my_foo(arg1, arg2);
if (ret) {
   // Print error message if needed (Linux not allow print error for ENOMEM)
   // return error or goto rewind section
}

// do something in success

return 0;

rewind:
...
return ret;
```

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,329 @@
...
+	} else {
+		if (ndev->dev_status >= AIE2_DEV_START) {
+			aie2_stop_event_trace_send(ndev);
+			aie2_event_trace_free(ndev);
+		} else {
+			XDNA_DBG(ndev->xdna, "Event trace is not started");
+		}
+		ndev->event_trace_enabled = state;
+	}
+	clear_event_trace_msix(ndev);
```
mamin506: This looks weird.

1. Why we need to clear_event_trace_msix(), when error happen?
2. Why we need to clear_event_trace_msix(), when enable or disable? 

log_buffer_irq_handler() should be the only place that clear interrupt.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,329 @@
...
+	free_irq(req_buf->log_ch_irq, ndev);
+	kfree(req_buf);
+	ndev->event_trace_req = NULL;
+}
+
+static int aie2_start_event_trace_send(struct amdxdna_dev_hdl *ndev)
+{
+	int ret;
+	struct amdxdna_dev *xdna = ndev->xdna;
+	struct event_trace_req_buf *trace_req_buf = ndev->event_trace_req;
```
mamin506: Linux coding style require reverse XMAS tree declarations. Same for other functions.
```suggestion
	struct event_trace_req_buf *trace_req_buf = ndev->event_trace_req;
	struct amdxdna_dev *xdna = ndev->xdna;
	int ret;
```

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,329 @@
...
+
+	if (!trace_req_buf) {
+		XDNA_DBG(xdna, "Event trace req buffer is not allocated!");
+		return -ENOMEM;
+	}
+
```
mamin506: This check is not needed.
1. This function is static. It is invisible to other part of the driver and it is not directly called by user interface (debugfs).
2. If everything works fine, trace_req_buf is allocated by aie2_event_trace_init() in this function.

So, do NOT check trace_req_buf here and just let kernel Oops. So that you know this is a bug in this file.

mamin506 (changes requested):
Still reviewing aie2_event_trace.c. Please address comments for other files.

src/driver/amdxdna/aie2_msg_priv.h
VinitAmd: This is defined in Driver/FW interface header file mgmt_msg.h, will be used in coming implementation, still want to remove ?

src/driver/amdxdna/aie2_pci.c
VinitAmd: If this error can be ignore, please used XDNA_DBG(). 
--> This can be ignored.
aie2_assign_event_trace_state() should check ndev->event_trace_req is NULL or not.
---> Yes above function should not check event_trace_req  is null or not, since event_trace_req  is place holder for log buffer, it should be checked before logbuf allocation at aie2_start_event_trace_send.c at 255.
This is a bug.
---> Can you pls add info on what bug ?

src/driver/amdxdna/aie2_event_trace.c
VinitAmd: If there is -ENOMEM error for event_trace_req_buf alloc should this cause kernel crash?
This is good to have feature not essential module to run ?

src/driver/amdxdna/aie2_msg_priv.h
VinitAmd: LOG_BUFFER_IRQ  is replaced with msi_idx param in start_event_trace_resp. 
in the next FW version update MPIPU_IOHUB_INT_27_ALIAS will also be part of start_event_trace_resp as new u64 bit field, please approve with current change.

src/driver/amdxdna/aie2_msg_priv.h
mamin506: Yes. If this is for something in the future, please add them later.

src/driver/amdxdna/aie2_msg_priv.h
mamin506: Oh, this sounds great. Please put a comment "/* FIXME: To be deleted */".

src/driver/amdxdna/aie2_pci.c
mamin506: I see in the new code this is XDNA_DBG(). It looks good to me. The bug is also being fixed in the new code.

If this pointer can be NULL, you should check "if (!ndev->event_trace_req)" in below APIs,
aie2_is_event_trace_enable()
aie2_set_trace_timestamp()
aie2_assign_event_trace_state()
aie2_event_trace_fini()

There is no need to do this checking in those static functions in aie2_event_trace.c. Because the static function must be called by one of above.

src/driver/amdxdna/aie2_event_trace.c
mamin506: First of all, this is a static function. It is called by other function in this file. For now, maybe this code is okay.
My suggestion is target for long term maintenance. Do this check in the beginning of aie2_assign_event_trace_state(). Then it protects all of the following static functions. You don't need to think about if you should check if you add a new static function in the future.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+int aie2_event_trace_init(struct amdxdna_dev_hdl *ndev)
+{
+	struct event_trace_req_buf *req_buf;
+
+	req_buf = kzalloc(sizeof(*req_buf), GFP_KERNEL);
+	if (!req_buf)
+		return -ENOMEM;
+
+	req_buf->ndev = ndev;
+	ndev->event_trace_req = req_buf;
```
mamin506: It will be good to initial event_trace_enabled to false explicitly.
```suggestion
	req_buf->ndev = ndev;
	req_bug->event_trace_enabled = false;
	ndev->event_trace_req = req_buf;
```

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+		}
+	} else {
+		if (aie2_is_event_trace_enable(ndev)) {
+			aie2_stop_event_trace_send(ndev);
+			aie2_event_trace_free(ndev);
+		}
+	}
+
+	ndev->event_trace_req->event_trace_enabled = state;
+	XDNA_DBG(ndev->xdna, "Event trace state: %d", state);
```
mamin506: 1. In the "else" branch, there is no need to check aie2_is_event_trace_enable(ndev) again. It is checked in line 279.
2. If aie2_start_event_trace_send() failed, it still go to line 298 to set enabled as true. This doesn't look correct.
3. If aie2_start_event_trace_send() failed, it should release all the resources that it allocated. This function is not the place to call aie2_event_trace_free(). 

```suggestion
	if (!state) {
	        /* Stop event trace */
	        aie2_stop_event_trace_send(ndev);
	        aie2_event_trace_free(ndev);

		goto done:
	}

	/* Start event trace */
	err = aie2_start_event_trace_send(ndev);
	if (err) {
		XDNA_ERR(ndev->xdna, "Send start event trace failed, ret %d", ret);
		return;
	}

done:
	ndev->event_trace_req->event_trace_enabled = state;
	XDNA_DBG(ndev->xdna, "Event trace state: %d", state);
```

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+	struct workqueue_struct  *wq;
+	struct work_struct       work;
+	u8                       *fw_log_buf;
+	u8                       *buf;
+	u64                      dram_buffer_address;
+	u64                      resp_timestamp;
+	u64                      sys_start_time;
+	u32                      dram_buffer_size;
+	int                      log_ch_irq;
+	bool                     event_trace_enabled;
```
mamin506: Linux has 100 columns limit. Use a shorter and still clear name is better.
```suggestion
	bool                     enabled;
```

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+	dma_free_noncoherent(xdna->ddev.dev, req_buf->dram_buffer_size, req_buf->buf,
+			(dma_addr_t)req_buf->dram_buffer_address, DMA_BIDIRECTIONAL);
+
+	kfree(ndev->event_trace_req->fw_log_buf);
+	free_irq(req_buf->log_ch_irq, ndev);
+}
+
+static int aie2_start_event_trace_send(struct amdxdna_dev_hdl *ndev)
+{
+	struct event_trace_req_buf *trace_req_buf = ndev->event_trace_req;
```
mamin506: I see naming inconsistent. struct event_trace_req_buf variables/arguments named "req_buf" or "trace_req_buf". Please cleanup, "req_buf" is a good name.

```suggestion
	struct event_trace_req_buf *req_buf = ndev->event_trace_req;
```

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+}
+
+static void aie2_event_trace_free(struct amdxdna_dev_hdl *ndev)
+{
+	struct event_trace_req_buf *req_buf = ndev->event_trace_req;
+	struct amdxdna_dev *xdna = ndev->xdna;
+
+	if (!req_buf)
+		return;
+
```
mamin506: Don't check this in static function.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+	ret = aie2_event_trace_alloc(ndev);
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to allocate log buffer, ret %d", ret);
+		return ret;
+	}
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+	ret = aie2_start_event_trace(ndev, trace_req_buf->dram_buffer_address,
+				trace_req_buf->dram_buffer_size);
+	return 0;
```
mamin506: If ret is error, still return 0? 

If aie2_start_event_trace() failed, call aie2_event_trace_free() to release resources allocated by aie2_event_trace_alloc().

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+
+	req_buf->buf = dma_alloc_noncoherent(xdna->ddev.dev, TRACE_EVENT_BUFFER_SIZE,
+					(dma_addr_t *)&req_buf->dram_buffer_address, DMA_BIDIRECTIONAL, GFP_KERNEL);
+
+	if (!req_buf->buf)
+		return -ENOMEM;
+
+	req_buf->dram_buffer_size = TRACE_EVENT_BUFFER_SIZE;
+	XDNA_DBG(ndev->xdna, "Start event trace buf addr: 0x%llx size 0x%x", req_buf->dram_buffer_address,
+			 req_buf->dram_buffer_size);
```
mamin506: The alignment doesn't look good. Can you run checkpatch.pl? See https://gitenterprise.xilinx.com/TurinAI/amd-aie#checkpatch

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+	if (!ndev->event_trace_req) {
+		XDNA_DBG(xdna, "Event tracing is not started");
+		return 0;
+	}
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+	return aie2_stop_event_trace(ndev);
+}
+
+static int aie2_event_trace_alloc(struct amdxdna_dev_hdl *ndev)
```
mamin506, line 209: aie2_event_trace_alloc() and aie2_event_trace_free() are not symmetric.
 
 free() release more resources than alloc().

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+
+static irqreturn_t log_buffer_irq_handler(int irq, void *data)
+{
+	struct amdxdna_dev_hdl *ndev = (struct amdxdna_dev_hdl *)data;
+
+	trace_mbox_irq_handle("LOG_BUFFER", irq);
+	clear_event_trace_msix(ndev);
+
+	if (aie2_is_event_trace_enable(ndev))
+		queue_work(ndev->event_trace_req->wq, &ndev->event_trace_req->work);
```
mamin506: If event trace is disabled, can this IRQ handler still active?

src/driver/amdxdna/aie2_message.c
```diff
@@ -232,6 +232,34 @@ int aie2_query_firmware_version(struct amdxdna_dev_hdl *ndev,
...
+	return 0;
+}
+
+int aie2_stop_event_trace(struct amdxdna_dev_hdl *ndev)
+{
+	DECLARE_AIE2_MSG(stop_event_trace, MSG_OP_STOP_EVENT_TRACE);
+
+	XDNA_DBG(ndev->xdna, "send stop event trace msg");
+	return aie2_send_mgmt_msg_wait(ndev, &msg);
+}
```
mamin506: This and aie2_start_event_trace() are not symmetric.
In aie2_start_event_trace(), it does
1. send start event trace message
2. set trace timestamp which allocate workqueue and interrupt resoruces

Then in aie2_stop_event_trace(), it should do
1. send stop event trace message
2. unset trace timestamp and release workqueue and interrupt resources

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+		XDNA_ERR(xdna, "Failed to register irq %d ret %d", msi_idx, ret);
+		goto destroy_wq;
+	}
+
+	req_buf->fw_log_buf = kzalloc((sizeof(u8) * TRACE_EVENT_BUFFER_SIZE), GFP_KERNEL);
+	if (!req_buf->fw_log_buf) {
+		XDNA_ERR(xdna, "Failed to allocate kernel log buffer");
+		ret = -ENOMEM;
+		goto destroy_wq;
+	}
```
mamin506: 1. Use kcalloc() for "sizeof() * item size" kind of memory allocation. This is from linux community.
2. No error print for -ENOMEM.
3. Suggestion, move this to the beginning of this function. In linux driver, usually, if a function needs to allocation memory, always do it in the beginning.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+	req_buf = ndev->event_trace_req;
+	INIT_WORK(&req_buf->work, deffered_logging_work);
+	req_buf->wq = alloc_ordered_workqueue("LOG_BUFFER", 0);
+	if (!req_buf->wq) {
+		XDNA_ERR(xdna, "Failed to allocate workqueue");
+		ret = -ENOMEM;
+		goto free_dma_trace_buf;
+	}
+
+	req_buf->log_ch_irq = pci_irq_vector(to_pci_dev(xdna->ddev.dev), msi_idx);
```
mamin506: pci_irq_vector() can return error. The error needs to be handled.

https://elixir.bootlin.com/linux/v6.14-rc3/source/drivers/pci/msi/api.c#L301

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+
+		trace_req_buf->fw_log_buf[log_size] = 0;
+
+		while (str < end) {
+			log_content = (struct trace_event_log_data *)str;
+			payload = ((u64)log_content->payload_hi << 32) | log_content->payload_low;
+			fwTicks = log_content->counter - trace_req_buf->resp_timestamp;
+			fwTicks = fwTicks / 24 + ndev->event_trace_req->sys_start_time;
+			printk(KERN_DEBUG "[IPU]::[%llu] type: 0x%04x payload:0x%016llx",
+					fwTicks, log_content->type, payload);
```
mamin506: checkpatch.pl should report error for this. Linux doesn't allow driver to directly call printk. Please use XDNA_DBG().

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+
+static u32 aie2_get_trace_event_content(struct event_trace_req_buf *trace_req_buf)
+{
+	struct amdxdna_dev_hdl *ndev = trace_req_buf->ndev;
+	struct trace_event_metadata *trace_metadata;
+	u8 *sysBuf = trace_req_buf->buf;
+	u8 *kern_buf = trace_req_buf->fw_log_buf;
+	u32 rd_ptr, wr_ptr, wr_ptr_wrap;
+	u32 total_log_size = 0, log_size = 0;
+	u32 rb_size, offset = 0;
```
mamin506: Reverse XMAS tree please.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+
+	rb_size = TRACE_EVENT_BUFFER_SIZE - TRACE_EVENT_BUFFER_METADATA_SIZE;
+	if (rb_size == 0)
+		return 0;
```
mamin506: The rb_size is a constant. Can we define a macro for it, Instead of rely on compiler to optimize this?
The size is a constant then no need to check the size. If someone change the constant to zero in this code base, you want a hint in the dmesg log. Use WARN_ON().

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+
+	trace_metadata = (struct trace_event_metadata *)(sysBuf + rb_size);
+	rd_ptr = (u32)(trace_metadata->head_offset % rb_size);
+	wr_ptr = (u32)(trace_metadata->tail_offset);
+	wr_ptr_wrap = wr_ptr % rb_size;
+
+	/* Update the Ring Buffer read pointer */
+	trace_metadata->head_offset = wr_ptr;
+
+	do {
```
mamin506: 1. Why the comment say it is "read pointer" but the variable name is "wr_ptr" (write pointer)?
2. Once head_offset is updated. The firmware can continue to write to the ring buffer. right? But driver read ringbuf after update head_offset. No race condition?

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,326 @@
...
+	struct event_trace_req_buf *trace_req_buf;
+	struct trace_event_log_data *log_content;
+	u64 payload;
+	u32 log_size;
+
+	trace_req_buf = ndev->event_trace_req;
+	log_size = aie2_get_trace_event_content(trace_req_buf);
+	XDNA_DBG(ndev->xdna, "FW log size in bytes %u", log_size);
+
+	if (log_size) {
```
mamin506: Use below coding style to avoid indentation.
```suggestion
	if (!log_size)
	    return;
	    
	// Read and print log
```

mamin506 (changes requested):

src/driver/amdxdna/aie2_event_trace.c
VinitAmd: 1. Why the comment say it is "read pointer" but the variable name is "wr_ptr" (write pointer)?
  -----> head_offset is always a read pointer, when first round of log is read, the head_offset should be updated with tail_offset which is wr_ptr for FW so that for next round head_offest should start from next read position where FW written the content.
  
2. FW always writes to tail_offset , driver reads from head offset. The buffer size is maintained in such away there would be no race condition.
pls follow https://confluence.amd.com/pages/viewpage.action?spaceKey=IPUFW&title=Strix+Logging for more info.
though there is ambiguity in some text (FW team will update).

src/driver/amdxdna/aie2_event_trace.c
VinitAmd: checkpatch.pl doesnt report error but a warning, there are 2 problems in using   XDNA_DBG().
1. XDNA_DBG() appends module name in the log which not desired FW log requires [IPU] as module.
2. Keeping it DBG, kernel ignores these logs doesn't capture in dmesg because of debug level.
Actually, these logs are the required result of this feature which is missed in XDNA_DBG().
So kept as printk, Though I agree to your suggestion, we are thinking for long term solution when FW event tracing and logging both are implemented and enabled. Will add a comment and update this in next change.

src/driver/amdxdna/aie2_event_trace.c
VinitAmd: agree on 1 & 2, IMO 3rd change is not reasonable here.
kernel log buffer is only required when workqueue and Irq stablished else this allocation and cleanup in failure case 
is overhead. pls add your comment if otherwise?

src/driver/amdxdna/aie2_event_trace.c
VinitAmd: Agree to 1 &2 but overall code flow modified so the above suggested change is not applicable completely.

src/driver/amdxdna/aie2_event_trace.c
mamin506: Our expectation is 0 ERRORs, WARNINGs and CHECKs. 

XDNA_DBG() is dev_dbg() based. If you don't want device debug print format, please use [pr_debug](https://elixir.bootlin.com/linux/v6.14-rc3/C/ident/pr_debug).

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,366 @@
...
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to register irq %d ret %d", msi_idx, ret);
+		goto destroy_wq;
+	}
+
+	req_buf->fw_log_buf = kcalloc(TRACE_EVENT_BUF_SIZE, sizeof(u8), GFP_KERNEL);
+	if (!req_buf->fw_log_buf) {
+		ret = -ENOMEM;
+		goto destroy_wq;
+	}
```
mamin506: If this failed, free_irq() is needed.

mamin506 (changes requested):

src/driver/amdxdna/aie2_event_trace.c
mamin506: Okay, as we talk offline. Please don't use "wr_" and "rd_", use head_ptr and tail_ptr instead.
For 2, if this is the design of ringbuf, I'm okay with that.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -0,0 +1,368 @@
...
+static int aie2_stop_event_trace_send(struct amdxdna_dev_hdl *ndev)
+{
+	struct amdxdna_dev *xdna = ndev->xdna;
+	int ret;
+
+	if (!ndev->event_trace_req) {
+		XDNA_DBG(xdna, "Event tracing is not started");
+		return 0;
+	}
+
```
mamin506, line 274: The condition will always be false. It was checked in aie2_assign_event_trace_state() and return early if this condition is true. But just a minor issue. Please delete this in your next PR.

mamin506 (approved):

PR #334, opened 2024-12-20
Addition of xrt_smi_configuration.json to xdna tools

aktondak:
This PR adds the xrt_smi_configuration.json to xdna tools and adds a new query object in ishim to provide the path to it at run-time.
This is required for xrt-smi architectural change that we plan to do for https://jira.xilinx.com/browse/VITIS-13804.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,3 +45,12 @@ install(FILES ${amdxdna_tools}
   DESTINATION xrt/${XDNA_COMPONENT}
   COMPONENT ${XDNA_COMPONENT}
   )
+
+set(amdxdna_xrt_smi_config
+  ${CMAKE_CURRENT_SOURCE_DIR}/tools/xrt_smi_config.json
+  )
+install(FILES ${amdxdna_xrt_smi_config}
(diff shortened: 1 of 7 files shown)


AMDGithubSCIMAdmin:

maxzhen (commented):
The xrt_smi_config.json needs a copyright? It looks more like source code, not some tools for repo maintenance. Should we move it to src/tools?

maxzhen:
ok to test

aktondak:
> The xrt_smi_config.json needs a copyright? It looks more like source code, not some tools for repo maintenance. Should we move it to src/tools?

That makes sense. It should be moved to src/tools. Let me make that change
For copyright, jsons dont directly support commenting directly so I've not added it right now. Can you point me to any other json which handles copyrights correctly?

maxzhen:
> > The xrt_smi_config.json needs a copyright? It looks more like source code, not some tools for repo maintenance. Should we move it to src/tools?
> 
> That makes sense. It should be moved to src/tools. Let me make that change For copyright, jsons dont directly support commenting so I've not added it right now. Can you point me to any other json which handles copyrights correctly?

Ah, so we need to treat it as binary in terms of copyright. You need to add the path to https://github.com/amd/xdna-driver/blob/main/WHENCE

maxzhen (approved):

maxzhen:
retest this please

aktondak:
retest this please

PR #349, opened 2025-01-13
add umq_version for npu.info

xdavidz:
@maxzhen this is needed for npu3 driver umq, but we have to get this in the core.

### DIFF
--- a/tools/info.json
+++ b/tools/info.json
@@ -36,12 +36,9 @@
 			"pci_revision_id": "11",
 			"version": "1.0.4.180",
 			"fw_name": "npu.sbin"
-		},
-                {
-			"device": "npu3",
-			"version": "20250108-6b53eb8f",
-			"fw_name": "app.elf"
-                }
-
-	]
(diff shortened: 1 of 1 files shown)


maxzhen (approved):

mamin506 (changes requested):
Please don't put this "app.elf" and metadata in "firmwares" array. It doesn't look like other firmwares. Create a separate array.

xdavidz:
> Please don't put this "app.elf" and metadata in "firmwares" array. It doesn't look like other firmwares. Create a separate array.

If creating a separate array, would "umq_version" ok? or you prefer the name as "version"?

mamin506:
> > Please don't put this "app.elf" and metadata in "firmwares" array. It doesn't look like other firmwares. Create a separate array.
> 
> If creating a separate array, would "umq_version" ok? or you prefer the name as "version"?

umq_version is OK.

mamin506 (approved):

PR #352, opened 2025-01-14
Some code refine and fix

mamin506:
wait 2 seconds before stop channel on aie2 
before release hardware resource, destroy drm sched entity


mamin506:
Talk with Max. I will force merge this, then fix the pipeline.

PR #353, opened 2025-01-14
Fixing umq log metadata to be inside umq src files

hlaccabu:


src/shim/hwctx.cpp
```diff
@@ -244,33 +244,14 @@ delete_ctx_on_device()
...
-  }
-}
-
 void
 hw_ctx::
 init_log_buf()
 {
-  size_t column_size = 1024;
-  auto log_buf_size = m_num_cols * column_size + sizeof(m_metadata);
+  auto log_buf_size = m_num_cols * 1024;
```
xdavidz: this is incorrect size. Let me think how to handle this.

maxzhen: Moving entire init_log_buf and related stuff to umq hwctx?

hlaccabu: @maxzhen We are thinking of doing that. I will update this PR to remove init_log_buf from parent hwctx.cpp/.h if we decide that.

src/shim/umq/hwctx.cpp
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2023-2024, Advanced Micro Devices, Inc. All rights reserved.
 
+#include "bo.h"
```
xdavidz: we don't need to include bo.h?

hlaccabu: Ah thank you, that was leftover before I changed some of the parent class variables to protected.

mamin506:
retest this please

maxzhen (approved):

xdavidz (approved):

PR #354, opened 2025-01-15
hwctx to ctx; fwctx to hwctx

mamin506:


src/driver/amdxdna/aie2_solver.h
```diff
@@ -81,8 +81,8 @@ struct clk_list_info {
 };
 
 struct xrs_action_ops {
-	int (*load_hwctx)(struct amdxdna_hwctx *hwctx, struct xrs_action_load *action);
-	int (*unload_hwctx)(struct amdxdna_hwctx *hwctx);
+	int (*load_ctx)(struct amdxdna_ctx *ctx, struct xrs_action_load *action);
```
maxzhen: Does it make sense to keep the original name for this call back API? We are allocating ctx on firmware, right?

maxzhen (approved):

src/driver/amdxdna/aie2_solver.h
mamin506: Oh, good catch. This is a over kill.

PR #364, opened 2025-01-20
Ctx limit and module parameters

mamin506:
1. Add hardware context limit and context limit for each device.
2. Add module parameters to config the above limitation.

### DIFF
--- a/src/driver/amdxdna/aie2_pci.h
+++ b/src/driver/amdxdna/aie2_pci.h
@@ -257,7 +257,8 @@ struct amdxdna_dev_hdl {
 	struct async_events		*async_events;
 
 	u32				dev_status;
-	u32				ctx_num;
+	u32				hwctx_num;
+	u32				hwctx_limit;
 };
 
 #define DEFINE_BAR_OFFSET(reg_name, bar, reg_addr) \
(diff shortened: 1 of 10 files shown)


src/driver/amdxdna/aie2_pm.c
```diff
@@ -68,7 +68,7 @@ int aie2_pm_set_mode(struct amdxdna_dev_hdl *ndev, enum amdxdna_power_mode_type
 
 	switch (target) {
 	case POWER_MODE_TURBO:
-		if (ndev->ctx_num) {
```
maxzhen, line 71: it might make more sense to use ctx_cnt here.

maxzhen (changes requested):

src/driver/amdxdna/aie2_pci.h
```diff
@@ -257,7 +257,8 @@ struct amdxdna_dev_hdl {
 	struct async_events		*async_events;
 
 	u32				dev_status;
-	u32				ctx_num;
+	u32				hwctx_num;
```
maxzhen: call it hwctx_cnt? it is consistent with ctx_cnt.

src/driver/amdxdna/aie2_pm.c
mamin506: fixed

src/driver/amdxdna/aie2_pci.h
mamin506: done

maxzhen (approved):

PR #367, opened 2025-01-27
Adding support for arm64 arch compilation

hlaccabu:


src/shim/kmq/bo.cpp
```diff
@@ -25,6 +27,14 @@ flag_to_type(uint64_t bo_flags)
   return AMDXDNA_BO_INVALID;
 }
 
+void flush_cache_line(const char *cur) {
```
mamin506: flush_cache_line() is also defined in umq/hwq.cpp. Maybe you want to move this to a header file in shim/ then this two files just include it?

src/shim/umq/hwq.cpp
```diff
@@ -1,24 +1,31 @@
...
+  const int LINESIZE = 64;
+#if defined(__x86_64__) || defined(_M_X64)
+  _mm_clflush(cur);
+#else
+  __builtin___clear_cache((char *)cur, (char *)(cur + LINESIZE));
+#endif
+}
+
+void clflush_data(const void *data, size_t len) {
+  const int LINESIZE = 64;
```
mamin506, line 11: LINESIZE is also used in flush_cache_lin() and are both 64. Use macro is better.

src/shim/kmq/bo.cpp
```diff
@@ -25,6 +27,14 @@ flag_to_type(uint64_t bo_flags)
...
+  const int LINESIZE = 64;
```
mamin506: The x86_64 branch doesn't use this value. If it is ARM specific, move it to the ARM code.

mamin506 (changes requested):

mamin506:
Please update the copyright of bo.h.

src/shim/bo.h
```diff
@@ -16,9 +16,22 @@
...
 
 namespace shim_xdna {
 
+  const int LINESIZE = 64;
+  
+  void flush_cache_line(const char *cur) {
+#if defined(__x86_64__) || defined(_M_X64)
+  _mm_clflush(cur);
+#else
+  __builtin___clear_cache((char *)cur, (char *)(cur + LINESIZE));
```
houlz0507: This is instruction cache only, correct? I think we are flushing data.  And is it only supported on ARM and MIPS? Also if the cache flush function supports address range, please use it outside the loop. (change clflush_data instead)

houlz0507 (changes requested):

src/shim/bo.h
hlaccabu: Ah yes, __builtin___clear_cache() is only for instruction cache. I will look for a different alternative

hlaccabu: good catch

src/shim/bo.h
```diff
@@ -16,9 +16,29 @@
...
+#endif
 
 namespace shim_xdna {
 
+  const int LINESIZE = 64;
+  
+  void flush_cache_line(const char *cur) {
+#if defined(__x86_64__) || defined(_M_X64)
+  _mm_clflush(cur);
+#else
```
houlz0507: probably #elif defined (__ aarch64 __)?

amd-akshatah:
ok to test

houlz0507 (approved):

src/shim/bo.h
```diff
@@ -16,9 +16,29 @@
...
+  asm volatile(
+    "DC CIVAC, %[addr]\n"  // Clean and invalidate data cache
+    "DSB SY\n"             // Data Synchronization Barrier
+    "ISB SY\n"             // Instruction Synchronization Barrier
+    :
+    : [addr] "r" (cur)
+    : "memory"
+  );
+#endif
+  }
```
mamin506: Coding style...

```suggestion
inline void flush_cache_line(const char *cur)
{
#if defined(__x86_64__) || defined(_M_X64)
  _mm_clflush(cur);
#elif defined(__aarch64__)
  asm volatile(
    "DC CIVAC, %[addr]\n"  // Clean and invalidate data cache
    "DSB SY\n"             // Data Synchronization Barrier
    "ISB SY\n"             // Instruction Synchronization Barrier
    :
    : [addr] "r" (cur)
    : "memory"
  );
#endif
}
```

src/shim/bo.h
```diff
@@ -16,9 +16,29 @@
...
 #include <string>
 #include <atomic>
+#if defined(__x86_64__) || defined(_M_X64)
+  #include <x86intrin.h>
+#endif
 
 namespace shim_xdna {
 
+  const int LINESIZE = 64;
+
```
mamin506: Delete this.

src/shim/umq/hwq.cpp
```diff
@@ -1,24 +1,18 @@
...
-		_mm_clflush(cur);
-		cur += LINESIZE;
-	} while (cur <= (const char *)lastline);
+void clflush_data(const void *data, size_t len) {
+  const char *cur = (const char *)data;
+  uintptr_t lastline = (uintptr_t)(cur + len - 1) | (shim_xdna::LINESIZE - 1);
+  do {
+    shim_xdna::flush_cache_line(cur);
+    cur += shim_xdna::LINESIZE;
+  } while (cur <= (const char *)lastline);
```
mamin506: Coding style.

```suggestion
void clflush_data(const void *data, size_t len)
{
  const int LINESIZE = 64;
  const char *cur = (const char *)data;
  uintptr_t lastline = (uintptr_t)(cur + len - 1) | (LINESIZE - 1);
  do {
    shim_xdna::flush_cache_line(cur);
    cur += LINESIZE;
  } while (cur <= (const char *)lastline);
```

mamin506 (changes requested):

mamin506:
retest this please

src/shim/umq/hwq.cpp
mamin506: This is not being resolved yet. When define a function in shim code, use braces in a new line.
void foo()
{
   // something
}

hlaccabu: Apologies, I thought your previous suggested changes had that in there.

mamin506 (approved):

mamin506:
retest this please

mamin506:
retest this please

mamin506:
retest this please

amd-akshatah:
ok to test

amd-akshatah:
retest this please

mamin506:
retest this please

xdavidz (approved):
Looking good to me too.

PR #368, opened 2025-01-28
Implement xrt-smi query to report preemption counters

NishadSaraf:
Implement xrt-smi query to report preemption counters.

image

image

PS: Error "Subcommand not found" is due to the recent changes in xrt, unrelated to this change.

### DIFF
--- a/src/include/uapi/drm_local/amdxdna_accel.h
+++ b/src/include/uapi/drm_local/amdxdna_accel.h
@@ -406,7 +406,7 @@ struct amdxdna_drm_query_sensor {
  * @context_id: The ID for this context.
  * @start_col: The starting column for the partition assigned to this context.
  * @num_col: The number of columns in the partition assigned to this context.
- * @pad: Structure padding.
+ * @nwctx_id: Hardware context ID.
  * @pid: The Process ID of the process that created this context.
  * @command_submissions: The number of commands submitted to this context.
  * @command_completions: The number of commands completed by this context.
@@ -420,7 +420,7 @@ struct amdxdna_drm_query_ctx {
 	__u32 context_id;
 	__u32 start_col;
 	__u32 num_col;
-	__u32 pad;
+	__u32 hwctx_id;
 	__s64 pid;
 	__u64 command_submissions;
 	__u64 command_completions;
--- a/src/shim/device.cpp
+++ b/src/shim/device.cpp
@@ -387,6 +387,80 @@ struct preemption
   }
 };
 
+struct rtos_telemetry
+{
+  struct amdxdna_drm_query_telemetry {
+    uint32_t major;
+    uint32_t minor;
+    uint64_t reserved[352];
+    uint64_t layer_boundary_count[16];
+    uint64_t frame_boundary_count[16];
+    uint64_t reserved1[126];
+  };
+
+  using result_type = query::rtos_telemetry::result_type;
+
+  static constexpr uint32_t NPU_RTOS_MAX_USER_ID_COUNT = 16;
+  static constexpr uint16_t NPU4_DEVICE_ID = 0x17f0;
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
+  {
+    amdxdna_drm_query_ctx* data;
+    const uint32_t output_size = 256 * sizeof(*data);
+    result_type output;
+
+    auto device_id = sysfs_fcn<uint16_t>::get(get_pcidev(device), "", "device");
+    if (device_id != NPU4_DEVICE_ID)
+        return output;
+
+    std::vector<char> payload(output_size);
+    amdxdna_drm_get_info query_ctx = {
+      .param = DRM_AMDXDNA_QUERY_HW_CONTEXTS,
+      .buffer_size = output_size,
+      .buffer = reinterpret_cast<uintptr_t>(payload.data())
+    };
+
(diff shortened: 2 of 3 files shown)


src/include/uapi/drm_local/amdxdna_accel.h
```diff
@@ -413,6 +413,7 @@ struct amdxdna_drm_query_sensor {
 /**
  * struct amdxdna_drm_query_ctx - The data for single context.
  * @context_id: The ID for this context.
+ * @user_id: Firmware context ID.
```
mamin506: In driver, there are context and hardware context. I suggest we use hwctx_id.
```suggestion
 * @hwctx_id: Hardware context ID.
```

src/include/uapi/drm_local/amdxdna_accel.h
```diff
@@ -427,6 +428,7 @@ struct amdxdna_drm_query_sensor {
  */
 struct amdxdna_drm_query_ctx {
 	__u32 context_id;
+	__u32 user_id;
```
mamin506: Please move this to the end of the struct not adding in the middle.

src/include/uapi/drm_local/amdxdna_accel.h
```diff
@@ -507,6 +509,22 @@ struct amdxdna_drm_query_firmware_version {
...
+ * @frame_boundary_count: frame boundary preemption counts indexed with user ID
+ */
+struct amdxdna_drm_query_telemetry {
+	__u32 major;
+	__u32 minor;
+	__u64 reserved[352];
+	__u64 layer_boundary_count[16];
+	__u64 frame_boundary_count[16];
+};
+
```
mamin506: I don't see "struct amdxdna_drm_query_telemetry" is used in the driver. Please move this to device.cpp.

mamin506 (changes requested):
Some minor change is needed. Not reviewed SHIM changes.

src/shim/device.cpp
```diff
@@ -387,6 +387,72 @@ struct preemption
...
+    amdxdna_drm_get_info query_telemetry = {
+      .param = DRM_AMDXDNA_QUERY_TELEMETRY,
+      .buffer_size = sizeof(telemetry),
+      .buffer = reinterpret_cast<uintptr_t>(&telemetry)
+    };
+
+    pci_dev_impl.ioctl(DRM_IOCTL_AMDXDNA_GET_INFO, &query_telemetry);
+
+    if (telemetry.major != NPU_TELEMETRY_VERSION_MAJOR || telemetry.minor != NPU_TELEMETRY_VERSION_MINOR) {
+      memset(&telemetry, 0, sizeof(telemetry));
```
maxzhen: why do you need to clear the buffer?

src/include/uapi/drm_local/amdxdna_accel.h
NishadSaraf: would utilizing the pad field be a better option?

src/shim/device.cpp
NishadSaraf: This was done to align with the Windows driver implementation for the Phoenix devices. Since we don't have preemption support in Phoenix, the expectation is that querying the preemption counter should return zeros instead of an exception.

src/include/uapi/drm_local/amdxdna_accel.h
mamin506: Oh. yes. "pad" is a better option.

src/shim/device.cpp
maxzhen: You want to clear the buffer, send it down to driver and return to xrt-smi with whatever is returned from driver. On PHX, since there is no preemption support, driver should have already put all zeros in the buffer, no?

mamin506 (approved):
The driver change looks good.

src/shim/device.cpp
mamin506: XDNA driver do put all zeros in the buffer.
1. buff = dma_alloc_noncoherent()
2. memset(buff, 0, aligned_buffer_size)
3. query telemetry data from firmware
4. copy_to_user(args->buffer, buff, args->buffer_size)

NishadSaraf: As @mamin506 mentioned, the driver does initialize the buffer to all 0s. However, the firmware memsets it to all 0xFF values by default.

mamin506:
retest this please

src/shim/device.cpp
```diff
@@ -387,6 +387,80 @@ struct preemption
...
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
+  {
+    amdxdna_drm_query_ctx* data;
+    const uint32_t output_size = 256 * sizeof(*data);
+    result_type output;
+
+    auto device_id = sysfs_fcn<uint16_t>::get(get_pcidev(device), "", "device");
+    if (device_id != NPU4_DEVICE_ID)
```
maxzhen, line 414: Should this be if (device_id == NPU1_DEVICE_ID)?

maxzhen (changes requested):

src/shim/device.cpp
maxzhen: Talked to Nishad offline and the current implementation is good.

maxzhen (approved):

PR #373, opened 2025-02-03
avoid to use enum on ioctl()

mamin506:
Just a clean up, it is better to use define rather than enum on ioctl interface.


maxzhen (approved):

mamin506:
retest this please

PR #378, opened 2025-02-10
Refine sum_exe with safe guard codes

gyang1099:


gyang1099:
retest this please

gyang1099:
retest this please

maxzhen (changes requested):
This looks like a race condition in PMFW. It can't be fixed in driver. There is no guarantee that the SMU will still be in idle state after driver's read and before its first write.

mamin506:
@maxzhen , this is not a fix. The purpose of this PR is to let driver be able to show this situation to whoever debugging SMU and help them quickly narrow down.

It will be better to add "#ifdef AMDXDNA_DEVEL" macro for this. In the first line of the comment say "This is not a fix in the driver. It is just an internal debug helper". What do you think?

mamin506 (approved):

PR #403, opened 2025-03-04
Fixing coverity scan issues CID#69559, 69585, 163674

hlaccabu:


xdavidz (approved):
it looks ok to me. Please have Max review this too.

maxzhen (commented):
could you explain the change in pcidrv.cpp? The global variable x may not be initialized before create_pcidev()?

maxzhen:
> > could you explain the change in pcidrv.cpp? The global variable x may not be initialized before create_pcidev()?
> 
> Sort of, the global variable x depends on initialization of another global variable (driver_list) inside the XRT file that defines register_driver. The order of initialization is undefined, so there is a potential issue if we try to initialize X before driver_list is initialized. the exact error is: Initialization or destruction ordering is unspecified (GLOBAL_INIT_ORDER) constructor_uses_global_object: The constructor of global object ::x itself makes use of global object ::driver_list::list defined in another compilation unit. The order of construction is unspecified, so ::x might be created before ::driver_list::list is available. This issue also is raised in the file pcidrv_virtio.cpp which I now also changed.

The plug-in lib is dlopen()'ed by the xrt lib, so it is certain that driver_list is initialized before anything inside plug-in lib. so, i think this is a false alarm?

xdavidz:
> > > could you explain the change in pcidrv.cpp? The global variable x may not be initialized before create_pcidev()?
> > 
> > 
> > Sort of, the global variable x depends on initialization of another global variable (driver_list) inside the XRT file that defines register_driver. The order of initialization is undefined, so there is a potential issue if we try to initialize X before driver_list is initialized. the exact error is: Initialization or destruction ordering is unspecified (GLOBAL_INIT_ORDER) constructor_uses_global_object: The constructor of global object ::x itself makes use of global object ::driver_list::list defined in another compilation unit. The order of construction is unspecified, so ::x might be created before ::driver_list::list is available. This issue also is raised in the file pcidrv_virtio.cpp which I now also changed.
> 
> The plug-in lib is dlopen()'ed by the xrt lib, so it is certain that driver_list is initialized before anything inside plug-in lib. so, i think this is a false alarm?

@hlaccabu you can put Max's comment into the coverity system and mark this specific issue/warning as "false alarm". I guess there are not just one but several of these having the same type of false alarms.

maxzhen (approved):

PR #405, opened 2025-03-05
Update README to recommend HWE stack

jgmelber:
Point users to use pre-packaged Linux kernels rather than build from source if they can.


maxzhen:
ok to test

PR #408, opened 2025-03-07
Reset power mode during hw stop

vengutta18:


houlz0507:
Which exactly part actually fixed the issue? disable clkgating or set dpm_level to 0?

vengutta18:
> Which exactly part actually fixed the issue? disable clkgating or set dpm_level to 0?

It's disabling the clkgating, which is fixing the issue.

houlz0507:
> > Which exactly part actually fixed the issue? disable clkgating or set dpm_level to 0?
> 
> It's disabling the clkgating, which is fixing the issue.

To handle the suspend/resume case, it needs to change the setting temporarily. Driver need to set it back to previous settings after resume.

vengutta18:
> > > Which exactly part actually fixed the issue? disable clkgating or set dpm_level to 0?
> > 
> > 
> > It's disabling the clkgating, which is fixing the issue.
> 
> To handle the suspend/resume case, it needs to change the setting temporarily. Driver need to set it back to previous settings after resume.

That's a good point.  I was thinking about the driver unload only, so I didn't retain the power mode.
If the expectation is to retain the power mode (turbo mode here) during suspend/resume, then we need to set the power mode back to turbo during resume. I'll check if we can cache it and set it during the resume! 
Also, I was thinking if turbo mode doesn't let the NPU suspend, this issue should have been present during suspend/resume as well.

src/driver/amdxdna/aie2_pci.c
```diff
@@ -429,15 +430,15 @@ static int aie2_hw_start(struct amdxdna_dev *xdna)
 		goto destroy_mbox;
 	}
 
-	ret = aie2_pm_init(ndev);
+	ret = aie2_mgmt_fw_init(ndev);
```
houlz0507, line 433: Why does it need to switch the order of pm_init() and mgmt_fw_init()?

vengutta18: So In susped/resume path, during resume mgmt_fw_init -> xdna_reset -> npu_suspend is called, this again not supported when turbo mode is enabled, so that's why I had to swap it.

houlz0507 (approved):

src/driver/amdxdna/aie2_pm.c
```diff
@@ -57,17 +53,24 @@ int aie2_pm_init(struct amdxdna_dev_hdl *ndev)
...
+{
+	struct amdxdna_dev *xdna = ndev->xdna;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&ndev->aie2_lock));
+
+	if(aie2_pm_set_mode(ndev, POWER_MODE_DEFAULT, false))
+		XDNA_ERR(ndev->xdna, "Can not set to default power mode");
+}
+
+int aie2_pm_set_mode(struct amdxdna_dev_hdl *ndev, int target, bool cache_pw_mode)
```
mamin506: Please keep the definition of `int aie2_pm_set_mode(struct amdxdna_dev_hdl *ndev, int target)` unchange.
I don't want the caller outside of aie2_pm.c be able to control cache pw_mode or not.

I suggest to add below internal function in aie2_pm.c.
`static int pm_set_mode(struct amdxdna_dev_hdl *ndev, int target, bool cache_pw_mode)`

aie2_pm_init() and aie2_pm_fini() can call pm_set_mode() and control the behavior of cache pw_mode or not.

The function aie2_pm_set_mode() can implement as below,
```
int aie2_pm_set_mode(struct amdxdna_dev_hdl *ndev, int target)
{
        
        drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&ndev->aie2_lock));
       	if (ndev->pw_mode == target)
		return 0;
	return pm_set_mode(ndev, target, true);
}
```

mamin506 (changes requested):

mamin506 (approved):
Please just fix the indentation issues. Linux required tab not spaces for indentation.

maxzhen (approved):

gyang1099 (approved):

PR #434, opened 2025-03-26
VITIS-13930 Implement driver side for xrt_resource_raw

hlaccabu:


src/driver/amdxdna/aie2_pci.c
```diff
@@ -1044,6 +1044,36 @@ static int aie2_get_force_preempt_state(struct amdxdna_client *client,
...
+static int aie2_get_resource_info(struct amdxdna_client *client,
+				  struct amdxdna_drm_get_info *args)
+{
+	struct amdxdna_drm_get_resource_info res_info = {};
+	struct amdxdna_dev *xdna = client->xdna;
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+	const struct amdxdna_dev_priv *pdev = ndev->priv;
+	int ret = 0;
+
+	res_info.ipu_clk_max = pdev->dpm_clk_tbl[ndev->max_dpm_level - 2].hclk;
```
hlaccabu: @mamin506 I indexed at max_dpm_level - 2 because the last element of all the dpm_clk_tbls in the npu_reg files is 0, is that correct?

src/driver/amdxdna/aie2_pci.c
```diff
@@ -1044,6 +1044,36 @@ static int aie2_get_force_preempt_state(struct amdxdna_client *client,
...
+	res_info.ipu_task_max = pdev->hwctx_limit;
+	res_info.ipu_task_curr = ndev->hwctx_cnt;
+	u32 total_col = ndev->total_col;
+	u32 hclk_freq = ndev->hclk_freq;
+
+	if (pdev->protocol_major == 0x5 && pdev->protocol_minor == 0x5) { // For PHX
+		res_info.ipu_tops_max = 2 * total_col;
```
hlaccabu: @mamin506 the values for tops are supposed to be doubles, but when I'd try to do double math I would get compile error "SSE register return with SSE disabled"

src/driver/amdxdna/aie2_pci.c
mamin506: By current design, the last element is 0 in dpm_clk_tbls. But ndev->max_dpm_level should point to the max non-zero value. I don't think "ndev->max_dmp_level - 2" is correct.

src/driver/amdxdna/aie2_pci.c
```diff
@@ -1044,6 +1044,36 @@ static int aie2_get_force_preempt_state(struct amdxdna_client *client,
...
+	struct amdxdna_dev *xdna = client->xdna;
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+	const struct amdxdna_dev_priv *pdev = ndev->priv;
+	int ret = 0;
+
+	res_info.ipu_clk_max = pdev->dpm_clk_tbl[ndev->max_dpm_level - 2].hclk;
+	res_info.ipu_task_max = pdev->hwctx_limit;
+	res_info.ipu_task_curr = ndev->hwctx_cnt;
+	u32 total_col = ndev->total_col;
+	u32 hclk_freq = ndev->hclk_freq;
```
mamin506: Do NOT declare local variable in the middle of the code. Unless it is a new block. I know this is suggested in C++. But this not the case in Linux.

src/driver/amdxdna/aie2_pci.c
```diff
@@ -1044,6 +1044,36 @@ static int aie2_get_force_preempt_state(struct amdxdna_client *client,
...
 }
 
+static int aie2_get_resource_info(struct amdxdna_client *client,
+				  struct amdxdna_drm_get_info *args)
+{
+	struct amdxdna_drm_get_resource_info res_info = {};
+	struct amdxdna_dev *xdna = client->xdna;
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+	const struct amdxdna_dev_priv *pdev = ndev->priv;
+	int ret = 0;
```
mamin506: Reverse X'Mas tree for variable declare.

src/driver/amdxdna/aie2_pci.c
```diff
@@ -1044,6 +1044,36 @@ static int aie2_get_force_preempt_state(struct amdxdna_client *client,
...
+	} else {
+		res_info.ipu_tops_max = 4096 * total_col * pdev->dpm_clk_tbl
+		[ndev->max_dpm_level - 2].hclk / 1000000;
+		res_info.ipu_tops_curr = 4096 * hclk_freq * total_col / 1000000;
+	}
+
+	if (copy_to_user(u64_to_user_ptr(args->buffer), &res_info, sizeof(res_info)))
+		ret = -EFAULT;
+
+	return ret;
```
mamin506:
```suggestion
	if (copy_to_user(u64_to_user_ptr(args->buffer), &res_info, sizeof(res_info)))
		return -EFAULT;

	return 0;
```

src/driver/amdxdna/aie2_pci.c
mamin506: I want to avoid code like,
if (this dev) // do this;
else // do that;

Let me talk with you offline.

mamin506 (changes requested):

src/driver/amdxdna/aie2_pci.c
mamin506: As we discussed, please reference aie2_pm_init() and add hw_ops->get_tops() callback.

src/driver/amdxdna/aie2_pm.c
```diff
@@ -115,6 +115,31 @@ int aie2_pm_init(struct amdxdna_dev_hdl *ndev)
...
+
+int npu4_get_tops(struct amdxdna_dev_hdl *ndev, u64 *max, u64 *curr)
+{
+	const struct amdxdna_dev_priv *pdev = ndev->priv;
+	u64 total_col, hclk_freq;
+
+	total_col = ndev->total_col;
+	hclk_freq = ndev->hclk_freq;
+	*max = 4096 * total_col * pdev->dpm_clk_tbl[ndev->max_dpm_level].hclk / 1000000;
+	*curr = 4096 * hclk_freq * total_col / 1000000;
```
mamin506: Looks like both formula has same component. Can you try?
```suggestion
        topc = 4096 * total_col / 1000000;
	*max = topc * pdev->dpm_clk_tbl[ndev->max_dpm_level].hclk;
	*curr = topc * hclk_freq;
```

src/driver/amdxdna/aie2_pm.c
```diff
@@ -115,6 +115,31 @@ int aie2_pm_init(struct amdxdna_dev_hdl *ndev)
 	return 0;
 }
 
+int npu1_get_tops(struct amdxdna_dev_hdl *ndev, u64 *max, u64 *curr)
+{
+	u64 total_col, hclk_freq;
+
+	total_col = ndev->total_col;
+	hclk_freq = ndev->hclk_freq;
+	*max = 2 * ndev->total_col;
```
mamin506:
```suggestion
	*max = 2 * total_col;
```

src/driver/amdxdna/aie2_pm.c
```diff
@@ -115,6 +115,31 @@ int aie2_pm_init(struct amdxdna_dev_hdl *ndev)
...
+	*curr = hclk_freq * 2 * total_col / 1028;
```
mamin506:
```suggestion
	*curr = *max * hclk_freq / 1028;
```

src/driver/amdxdna/aie2_pci.c
```diff
@@ -1044,6 +1044,29 @@ static int aie2_get_force_preempt_state(struct amdxdna_client *client,
 	return 0;
 }
 
+static int aie2_get_resource_info(struct amdxdna_client *client,
+				  struct amdxdna_drm_get_info *args)
+{
+	struct amdxdna_drm_get_resource_info res_info = {};
+	struct amdxdna_dev *xdna = client->xdna;
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+	const struct amdxdna_dev_priv *pdev = ndev->priv;
```
mamin506: pdev is for "PCIe device" in the driver. Use "priv" instead.
```suggestion
	const struct amdxdna_dev_priv *priv = ndev->priv;
```

BTW, the declare is still not a reverse X'Mas tree. Please update it as well.

mamin506 (changes requested):

src/include/uapi/drm_local/amdxdna_accel.h
```diff
@@ -514,6 +530,7 @@ struct amdxdna_drm_get_info {
 #define	DRM_AMDXDNA_GET_POWER_MODE		9
 #define	DRM_AMDXDNA_QUERY_TELEMETRY		10
 #define	DRM_AMDXDNA_GET_FORCE_PREEMPT_STATE	11
+#define	DRM_AMDXDNA_QUERY_RESOURCE_INFO	12
```
mamin506: Please pay attention on the alignment for the digits as well. Thanks.

src/driver/amdxdna/aie2_pm.c
mamin506: I'm also thinking, the "topc" and "max tops" are constant value for a device instance. We can calculate it once in the aie2_pm_init(). Is it better?

src/driver/amdxdna/aie2_pci.c
```diff
@@ -1044,6 +1044,33 @@ static int aie2_get_force_preempt_state(struct amdxdna_client *client,
 	return 0;
 }
 
+static int aie2_get_resource_info(struct amdxdna_client *client,
+				  struct amdxdna_drm_get_info *args)
+{
+	struct amdxdna_drm_get_resource_info res_info = {};
```
mamin506: I forgot this line. Please only initialize local variable if necessary. This is the Linux coding style.
In this function, all members of res_info was assigned to a meaningful value.
So, do not initialize res_info when declare.

mamin506 (approved):

PR #453, opened 2025-04-03
[Draft PR] Linux driver update to support FW debug enhancements.[No review]

VinitAmd:
[Why]
  DRAM logging support is required to enhance FW debugging.

[What]
  Add log info to iomapped DRAM buffer from FW, read at
  driver side on FW triggered interrupts.


EugeneHiew:

src/driver/amdxdna/aie2_msg_priv.h
```diff
@@ -435,6 +438,62 @@ struct stop_event_trace_resp {
...
+    RUNTIME_CONFIGURATION_LOGGING_DEST_UTL = 0,
+    RUNTIME_CONFIGURATION_LOGGING_DEST_FIXED,
+    RUNTIME_CONFIGURATION_LOGGING_DEST_DEBUG_REGS,
+    RUNTIME_CONFIGURATION_LOGGING_DEST_STB,
+    RUNTIME_CONFIGURATION_LOGGING_DEST_DRAM,
+    MAX_RUNTIME_CONFIGURATION_LOGGING_DEST
+};
+
+struct config_logging_dram_buf_req {
+	u64 dram_buffer_address;
```
vengutta18: you could use dma_addr_t here! to avoid typecasting later?!

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,404 @@
...
+
+static void aie2_print_log_buffer_data(struct amdxdna_dev_hdl *ndev)
+{
+
+        struct logging_req_buf *log_req_buf;
+        struct log_msg_header *log_content;
+        u32 log_size;
+
+        log_req_buf = ndev->logging_req;
+        log_size = aie2_get_log_content(log_req_buf);
```
vengutta18: We discussed on this, looks like dma_sync_single_for_cpu(..) is a must since it's non-coherent memory!
Also just use FROM_DEVICE instead of BIDIRECTIONAL_FLAG. I don't think we write to that buffer!
Could you double check that to be sure?

vengutta18: Looks like in the event trace flow we aren't doing any dma_sync it's still functional as expected,
If you rule out the dma_sync_*, maybe something else is the cause!

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,404 @@
...
+
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to set runtime log level, ret %d", ret);
+		return ret;
+	}
+	XDNA_INFO(xdna, "Set runtime log level, ret %d", ret);
+
+	ret = aie2_set_runtime_cfg(ndev, RUNTIME_CONFIGURATION_LOGGING_DESTINATION,
+				   RUNTIME_CONFIGURATION_LOGGING_DEST_DRAM);
+
```
vengutta18, line 365: The key difference I see when compared with the event trace is that, setting this runtime cfg, make sure you have the correct values across working and non-working cases.

src/driver/amdxdna/aie2_logging.c
VinitAmd: > We discussed on this, looks like dma_sync_single_for_cpu(..) is a must since it's non-coherent memory! Also just use FROM_DEVICE instead of BIDIRECTIONAL_FLAG. I don't think we write to that buffer! Could you double check that to be sure?

Driver updates the buffer metadata header for next read/write hence BIDIRECTIONAL is required.

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,538 @@
...
+{
+	struct amdxdna_dev_hdl *ndev = req_buf->ndev;
+	struct log_buffer_metadata *log_metadata;
+
+	u8 *buf = req_buf->buf;
+	u32 try_cnt = 0;
+
+	log_metadata = (struct log_buffer_metadata *)(buf + rb_size);
+
+	while (try_cnt < 10 && (*head == *tail)) {
```
VinitAmd: This is WA , need to debug and find out if FW intr triggered why tail_ptr is not updated in buf metadata at host side ?

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,538 @@
...
+				   RUNTIME_CONFIGURATION_LOGGING_DEST_DRAM);
+
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to set runtime log destination, ret %d", ret);
+		aie2_free_log_buf(ndev);
+		return ret;
+	}
+
+	drm_clflush_virt_range(req_buf->buf, req_buf->dram_buffer_size);
+	return 0;
```
eble-amd: This flush seems too late.  If there were something in the CPU cache at this point, it would overwrite whatever values the firmware initialized when it was granted the buffer.  I would flush the cache after allocating the buffer and before granting the buffer to the firmware.

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,538 @@
...
+		msg_size = (header->argc) * sizeof(u32);
+		msg_size = ((msg_size / 8) + (msg_size % 8 ? 1 : 0)) * 8 + header_size;
+		str += msg_size;
+		XDNA_INFO(ndev->xdna, "[NPU FW][%u]:%s", header->serial_num, msg);
+		str = aie2_get_valid_msg_header(str, end);
+	}
+}
+
+static void deffered_logging_work(struct work_struct *work)
+{
```
eble-amd: `deffered` → `deferred`

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,538 @@
...
+
+	log_metadata = (struct log_buffer_metadata *)(sys_buf + log_rb_size);
+	head_ptr = log_metadata->head;
+	tail_ptr = make_64bit(log_metadata->tail_lo, log_metadata->tail_hi);
+
+	if (head_ptr == tail_ptr)
+		sync_cpu_buf(req_buf, &head_ptr, &tail_ptr, log_rb_size);
+
+	head_ptr_wrap = head_ptr % log_rb_size;
+	tail_ptr_wrap = tail_ptr % log_rb_size;
```
eble-amd: Suggestion: Code clarity would be improved by calling these something other than `_ptr`.  These are not pointers into any address space.  Rather, the 64-bit values are sequence numbers or counts of bytes produced and consumed, and the remainders after division are indices or offsets into the log ring.

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,538 @@
...
+	log_size = tail_ptr - head_ptr;
```
eble-amd: There is an unlikely but possible case that should be handled before using `tail_ptr`: the device has updated the lower 32 bits, the change in value needs to be carried into the upper 32 bits, but the device has not yet updated the upper 32 bits.

I recommend detecting this case and correcting the value before proceeding, making the assumption that the change in value just needs to carry 1 into the upper half of the value.  (I recommend against detecting it and retrying in a loop.)

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,538 @@
...
+
+	log_metadata->head = tail_ptr;
+
+	/* Handle buffer overflow case, dump all log w.r.t timestamp */
+	if (log_size > log_rb_size) {
+		XDNA_DBG(ndev->xdna, "log_size is %u, buffer overflow!", log_size);
+		u32 part_log = log_rb_size - tail_ptr_wrap;
+
+		memcpy((u8 *)kern_buf, (u8 *)sys_buf + tail_ptr_wrap, part_log);
+		memcpy((u8 *)(kern_buf + part_log), (u8 *)sys_buf, tail_ptr_wrap);
```
eble-amd: Shouldn't the cache be invalidated before calls to `memcpy` (see below also) so that the CPU sees the latest values written by the device?  (Immediately afterward could be fine too, as long as there is high confidence that nothing else will pull these regions back into the cache before the next time around.)

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,538 @@
...
+		return log_size;
+	}
+	/* General case when tail > head and with in log buff size */
+	memcpy((u8 *)kern_buf, (u8 *)(sys_buf + head_ptr_wrap), log_size);
+	return log_size;
+}
+
+static char *aie2_get_valid_msg_header(char *start, char *end)
+{
+	while (start && *start != 0xc0 && start < end)
```
eble-amd: Check `start < end` before dereferencing `start`.

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,538 @@
...
+		start += sizeof(char) * 8;
+	return start;
```
eble-amd: Suggestion: Code clarity would be improved by giving a common name to `8` wherever it is used as the alignment of messages within the log ring.

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,538 @@
...
+
+	if (!log_size)
+		return;
+
+	char *str = (char *)req_buf->kern_log_buf;
+	char *end = (char *)str + log_size;
+
+	str = aie2_get_valid_msg_header(str, end);
+	while (str && str < end) {
+		struct log_msg_header *header = (struct log_msg_header *)str;
```
eble-amd, line 173: To improve robustness in cases of memory corruption or defective firmware, verify that the entire header falls within the limit of `end` before reading it.  (This appears to verify only that the header starts before `end`.  If there is something elsewhere that makes this OK which I have missed, an explanatory comment would be a good idea.)

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,538 @@
...
+		char *msg = (char *)(str + header_size);
+		u32 msg_size;
+
+		if (!header || !msg)
+			return;
+
+		msg_size = (header->argc) * sizeof(u32);
+		msg_size = ((msg_size / 8) + (msg_size % 8 ? 1 : 0)) * 8 + header_size;
+		str += msg_size;
+		XDNA_INFO(ndev->xdna, "[NPU FW][%u]:%s", header->serial_num, msg);
```
eble-amd: To improve robustness in cases of memory corruption or defective firmware, verify that the entire message body falls within the limit of `end` before reading it.  (This appears to verify only that the message starts before `end`.  If there is something elsewhere that makes this OK which I have missed, an explanatory comment would be a good idea.)

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,538 @@
...
+		str = aie2_get_valid_msg_header(str, end);
```
eble-amd: To improve robustness in cases of memory corruption or defective firmware, verify that the message body is null-terminated before interpreting it with `%s`.  I believe that the developer of the Windows driver decided just to write a null character at the expected place unconditionally first, which seems fine to me given that this buffer is a private copy of the log ring rather than the actual log ring.  I can imagine other acceptable solutions.

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,538 @@
...
+	ret = aie2_set_runtime_cfg(ndev, RUNTIME_CONFIGURATION_LOGGING_DESTINATION,
+				   RUNTIME_CONFIGURATION_LOGGING_DEST_FIXED);
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to remove FW logging config, ret %d", ret);
+		return ret;
+	}
+	XDNA_DBG(xdna, "Set runtime cfg logging dest fixed success!");
+
+	aie2_remove_log_buf_irq(ndev);
+	aie2_free_log_buf(ndev);
```
eble-amd: Before releasing these resources, please revoke them from the firmware by sending MSG_OP_CONFIG_LOGGING_DRAM_BUF with a size of zero.

src/driver/amdxdna/aie2_logging.c
eble-amd: Actually, invalidating afterward is not good enough.  The CPU might prefetch memory beyond the end of these ranges that are being copied here, leaving it in cache for the next time.  The cache needs to be invalidated before reading.

src/driver/amdxdna/aie2_logging.c
VinitAmd: Yes, validation is done just before ln:190, we can discuss further.

src/driver/amdxdna/aie2_msg_priv.h
```diff
@@ -44,8 +44,10 @@ enum aie2_msg_opcode {
 	MSG_OP_START_EVENT_TRACE           = 0x10F,
 	MSG_OP_STOP_EVENT_TRACE            = 0x110,
 	MSG_OP_UPDATE_PROPERTY             = 0x113,
-	MSG_OP_GET_APP_HEALTH              = 0x114,
-	MSG_OP_ADD_HOST_BUFFER             = 0x115,
+	//MSG_OP_GET_APP_HEALTH              = 0x114,
```
zmotiXilinx, line 47: is the intend to disable APP_HEALTH? These codes doesn't match with IPU firmware. Please review.

src/driver/amdxdna/aie2_logging.c
VinitAmd: will verify this case in test phase.

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,539 @@
...
+		char log_msg[256] = {0};
+		u32 msg_size;
+
+		if (!header || !msg)
+			return;
+
+		msg_size = (header->argc) * sizeof(u32);
+		memcpy(log_msg, msg, msg_size + 1);
+		log_msg[msg_size + 1] = 0;
+		pr_debug("[NPU FW]:%s", log_msg);
```
eble-amd: When everything in the system is working as designed, one or more of the last four bytes of the message body will certainly be a zero.  The idea to force termination here is just for robustness in unexpected cases like defective firmware or memory corruption.  Therefore, instead of adding a terminator _after_ the message body, why not just set the last available byte to zero?  That would save the cost of making room for another byte (though you would need to handle the unexpected case `msg_size == 0` specially).

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,539 @@
...
+		char *msg = (char *)(str + header_size);
+		char log_msg[256] = {0};
+		u32 msg_size;
+
+		if (!header || !msg)
+			return;
+
+		msg_size = (header->argc) * sizeof(u32);
+		memcpy(log_msg, msg, msg_size + 1);
+		log_msg[msg_size + 1] = 0;
```
eble-amd: off by one
```c
memcpy(log_msg, msg, msg_size);
log_msg[msg_size] = 0;
```

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,539 @@
...
+		struct log_msg_header *header = (struct log_msg_header *)str;
+		char *msg = (char *)(str + header_size);
+		char log_msg[256] = {0};
+		u32 msg_size;
+
+		if (!header || !msg)
+			return;
+
+		msg_size = (header->argc) * sizeof(u32);
+		memcpy(log_msg, msg, msg_size + 1);
```
eble-amd: `header->argc` is an 8-bit field.  If it is not checked, `msg_size` could be up to 1024, which could overrun `log_msg`.  (Remember the 1 extra byte too.)

src/driver/amdxdna/aie2_logging.c
eble-amd: To clarify:
* When everything is working as designed, writing a zero to the last byte of the message body as it sits in kern_log_buf changes nothing.  There is no functional downside.
* When something has gone wrong, writing a zero to the last byte of the message body as it sits in kern_log_buf might change the value, but it was probably garbage anyway, so losing it is more than fair in exchange for stopping the interpretation of `%s` from overrunning the source.

I expect this suggested approach to perform better than making a copy of the whole message just for the sake of appending a `0` byte following the message body.

VinitAmd:
ok to test

VinitAmd:
trigger jenkins

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,532 @@
...
+		return ret;
+	}
+
+	ret = aie2_set_runtime_cfg(ndev, RUNTIME_CONFIGURATION_LOGGING_FORMAT,
+				   RUNTIME_CONFIGURATION_LOGGING_FORMAT_FULL);
+
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to set runtime log format, ret %d", ret);
+		return ret;
+	}
```
eble-amd: This early return and the next one appear to leave the log buffer allocated according to both this driver and the device firmware.  I can imagine a reason that this might be intended.  If it is intended, I recommend adding code comments explaining it for the sake of future maintainers, otherwise they might deem it an oversight and change it.

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,532 @@
...
+	XDNA_DBG(xdna, "Set runtime log level: %u, ret %d", req_buf->log_level, ret);
+
+	ret = aie2_set_runtime_cfg(ndev, RUNTIME_CONFIGURATION_LOGGING_DESTINATION,
+				   RUNTIME_CONFIGURATION_LOGGING_DEST_DRAM);
+
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to set runtime log destination, ret %d", ret);
+		aie2_free_log_buf(ndev);
+		return ret;
```
eble-amd: Unlike the early returns above, this frees the buffer from the perspective of this driver without first attempting to communicate the fact to the device firmware.  This should possibly be changed.  Either way, maintainability could be improved by handling errors consistently or trying to answer maintainers' questions in advance.

eble-xilinx (approved):
@VinitAmd, I focused my review time on interactions with the device firmware.  I have no concerns about merging this PR and improving things in a second pass.  Thanks for attending to my suggestions.

src/driver/amdxdna/aie2_logging.c
VinitAmd: This one is intended, will discuss in meeting along with runtime config error handling.

eble-amd: I see that you moved some of these into a new function.  That clarifies what is happening and addresses my concern about maintainability.

mamin506 (approved):
Reviewed aie2_message.c, aie2_msg_priv.h, aie2_pci.c and aie2_pci.h changes, LGTM.

src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -293,6 +293,91 @@ static int aie2_event_trace_show(struct seq_file *m, void *unused)
...
+
+	while (token) {
+		key = strsep(&token, "=");
+		val = token;
+
+		if (key && val) {
+			if (strcmp(key, "enable") == 0) {
+				ret = kstrtouint(val, 10, &enable);
+				if (ret) {
+					XDNA_INFO(ndev->xdna, "invalid enable value %u",
```
vengutta18: nit: xdna_dbg/err?

src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -293,6 +293,91 @@ static int aie2_event_trace_show(struct seq_file *m, void *unused)
...
+						  enable);
+					ret = -EINVAL;
+					goto out;
+				}
+			} else if (strcmp(key, "size") == 0) {
+				buf_size = memparse(val, NULL);
+			} else if (strcmp(key, "loglevel") == 0) {
+				ret = kstrtouint(val, 0, &loglevel);
+				if (ret) {
+					XDNA_INFO(ndev->xdna, "invalid log level %u",
```
vengutta18: ditto

src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -293,6 +293,91 @@ static int aie2_event_trace_show(struct seq_file *m, void *unused)
...
+						  loglevel);
+					ret = -EINVAL;
+					goto out;
+				}
+			} else {
+				XDNA_INFO(ndev->xdna, "invalid key %s format", key);
```
vengutta18: ditto

src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -293,6 +293,91 @@ static int aie2_event_trace_show(struct seq_file *m, void *unused)
...
+				ret = -EINVAL;
+				goto out;
+			}
+		}
+		token = strsep(&kbuf, " ");
+	}
+
+	XDNA_DBG(ndev->xdna, "Dram logging config: %u, %u, %u",
+		 enable, buf_size, loglevel);
+	if (enable && !buf_size) {
```
vengutta18, line 351: do we need this, if we just have enable 1, don't we pick the default buffer size and move on?

or else just add a print that we are not setting the dram logging since buf_size needs to be a valid value. because the previous print implies we are actually enabling it.

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,544 @@
...
+	ret = aie2_configure_dram_logging(ndev, req_buf->dram_buffer_address,
+					  req_buf->dram_buffer_size);
+	if (ret) {
+		XDNA_ERR(xdna, "Failed to configure FW logging, ret %d", ret);
+		aie2_free_log_buf(ndev);
+		return ret;
+	}
+
+	ret = aie2_apply_default_runtime_cfg(ndev);
+	if (ret) {
```
vengutta18: can we do the same goto approach like: "aie2_configure_log_buf_irq" to do the error handling?

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,544 @@
...
+		if (err)
+			return;
+
+		goto done;
+	}
+
+	err = aie2_configure_and_start_logging(ndev);
+	if (err)
+		return;
+
```
vengutta18, line 419: 408 - 418

can we have simple if else and check err once after the if else block?

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,544 @@
...
+	return 0;
+}
+
+void aie2_dram_logging_fini(struct amdxdna_dev_hdl *ndev)
+{
+	if (!ndev->logging_req)
+		return;
+
+	if (aie2_is_dram_logging_enable(ndev))
```
vengutta18, line 539: We are doing the state check inside aie2_update_logging_state, this is not needed.

src/driver/amdxdna/aie2_logging.c
```diff
@@ -0,0 +1,544 @@
...
+		return ret;
+	}
+
+	ret = aie2_apply_default_runtime_cfg(ndev);
+	if (ret) {
+		/*
+		 * Initial runtime config setting is failed, send detach logger
+		 * msg to FW and release buffer.
+		 */
+		(void)aie2_configure_dram_logging(ndev, req_buf->dram_buffer_address, 0);
```
vengutta18: what's the significance of this (void) ?

src/driver/amdxdna/aie2_logging.c
VinitAmd: No, this is not only state checking . aie2_update_logging_state only validates if same state is assigned again don't do anything. This is executed when logging is enabled and driver is uninstalled or device shutdown, driver should safely release resources and send msg to fw for detach logger.

src/driver/amdxdna/aie2_logging.c
VinitAmd: driver is not interested in return value of api at the moment, so don't save return value only avoid warning from compiler.

mamin506:
ok to test

PR #472, opened 2025-04-13
Add SKIP_KMOD/nokmod build option (fixes #342)

### LINKED ISSUE #342: Building the shim without the kernel driver
i dont want to build the kernel driver. i only need to build the shim library so that xrt could work with AMD NPU.

How do i do that?

xanderlent:
This allows the project to be built without building the kernel module.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,7 @@
 cmake_minimum_required(VERSION 3.19.0)
 project("amd-xdna")
 set(PROJECT_DESCRIPTION "AMD XDNA driver and SHIM for Linux")
+option(SKIP_KMOD "Skip building the kernel module" OFF)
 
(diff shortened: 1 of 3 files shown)


EugeneHiew:

xanderlent:
As an alternative and much cleaner approach, I'd propose moving the non-dkms `src/driver/tools` into `tools/` and having SKIP_KMOD (or whatever we decide on the option name) skip the entire driver subdirectory in `src/CMakeLists.txt`.

The only reason I didn't do that in the original version of this PR is because the `src/driver/tools` seem to be maintained with the kernel module.

maxzhen:
ok to test

xdavidz:
retest this please

mamin506:
retest this please

maxzhen:
Thanks for adding this support, @xanderlent !

PR #475, opened 2025-04-14
Update umq log metadata

hlaccabu:


src/shim/umq/hwctx.cpp
```diff
@@ -61,16 +61,16 @@ fini_log_buf(void)
 
 void
 hw_ctx_umq::
-set_metadata(int num_cols, size_t size, uint64_t bo_paddr, enum umq_log_flag flag)
```
xuhz, line 64: this is for the LOG_BUFFER setting.
we just add the define for UMQ_LOG_BUFFER in this PR, the what is the flag passed in to this call?

I guess we are still using UMQ_DEBUG_BUFFER?

hlaccabu: https://github.com/amd/xdna-driver/blob/54b246208ac3545ecb7eda7113a147db06128951/src/shim/umq/hwctx.cpp#L49

hlaccabu: This is the line where we pass the flag UMQ_DEBUG_BUFFER right now. Does that answer your question?

xuhz: that should be UMQ_LOG_BUFFER

xuhz (approved):

xdavidz (approved):

PR #478, opened 2025-04-16
Add functions to check if a messge op or runtime cfg is supported

mamin506:
1. Add functions to check if a message opcode or runtime cfg is supported
   - The function is working with a per device whitelist. For new opcode and runtime config, register it in the whitelist in npu*_regs.c
   - The opcode or runtime config function needs to decide what needs to be done. Silently skip, error or warning, etc.
2. Refactor recent preemption and update property messages based on 1)
3. Some bug fix.


maxzhen (approved):

mamin506:
retest this please

src/driver/amdxdna/aie2_pci.c
```diff
@@ -1131,7 +1103,7 @@ static int aie2_get_force_preempt_state(struct amdxdna_client *client,
 	}
 
 	ndev = xdna->dev_handle;
-	force.state = ndev->force_preempt_enabled;
+	force.state = ndev->force_preempt_enabled ? 1 : 0;
```
NishadSaraf, line 1106: It is best to explicitly have literals suffixed here and similar places with `U` to ensure type safety.

src/driver/amdxdna/aie2_msg_priv.h
```diff
@@ -545,25 +557,29 @@ struct legacy_config_cu_resp {
...
+	u32 resv;
 } __packed;
 
 struct update_property_resp {
 	enum aie2_msg_status status;
 } __packed;
 
+/* Do NOT put any firmware defined struct, enum etc. start from here */
+struct msg_op_ver {
+	u32			minor;
```
NishadSaraf: I'd name this and one below as `fw_minor`, but I'll leave it up to you to decide.

src/driver/amdxdna/aie2_message.c
```diff
@@ -95,26 +149,95 @@ int aie2_get_runtime_cfg(struct amdxdna_dev_hdl *ndev, u32 type, u64 *value)
...
+{
+	int ret;
+
+	ret = aie2_runtime_update_prop(ndev, ctx, UPDATE_PROPERTY_TIME_QUOTA, us);
+	if (ret == -EOPNOTSUPP) {
+		XDNA_DBG(ndev->xdna, "update time quota not support, skipped");
+		return 0;
+	}
+
+	XDNA_DBG(ndev->xdna, "%s execution time quantum updated to %d us",
```
NishadSaraf: This debug print may be incorrect when `aie2_runtime_update_prop()` fails with other error code other than `-EOPNOTSUPP`

NishadSaraf (changes requested):
Thanks for the change. Just a few minor comments.

src/driver/amdxdna/aie2_pci.c
mamin506: no, force.state is u32, use 0 and 1 is good and clear. I will just keep this unchange.

src/driver/amdxdna/aie2_msg_priv.h
mamin506: Good suggestion, fixed.

src/driver/amdxdna/aie2_message.c
mamin506: Fixed.

src/driver/amdxdna/aie2_pci.c
NishadSaraf: Actually, force.state is u8 same as force_preempt_enabled which makes this ternary operation redundant.

NishadSaraf (approved):
We might not need to explicitly convert boolean values to 1 or 0 anymore since we're using some newer C features, like flexible array notation, which C89 compilers don't even support, as far as I know. This could lead to bigger discussions, but for now, approving the change. We can circle back to this if needed. :)

PR #479, opened 2025-04-16
return -ENOSPC when user context buffer is smaller than needed

mamin506:


src/driver/amdxdna/aie2_pci.c
```diff
@@ -1282,9 +1282,10 @@ static int aie2_query_ctx_status_array(struct amdxdna_client *client,
 	WARN_ON(ctx_limit > AMDXDNA_MAX_NUM_ELEMENT);
 	ctx_cnt = aie2_rq_active_context(&xdna->dev_handle->ctx_rq);
 	if (args->num_element < ctx_cnt) {
-		XDNA_DBG(xdna, "Invalid number of elements, max ctx %d", ctx_limit);
+		XDNA_DBG(xdna, "Not enough space. Total ctx %d, got %d",
+			 ctx_cnt, args->num_element);
```
hlaccabu, line 1286: Should we also include ctx_limit in debug statement?

mamin506: If debug is enable, the reason of seeing this debug print just related to `args->num_element` and `ctx_cnt`. ctx_limit is not related.

If ctx_limit, which can be change by a module parameter, has something wrong, the WARN_ON() will report unexpected ctx_limit.

PR #486, opened 2025-04-19
Fail early for invalid payload length and minor optimization

NishadSaraf:
- Fail early for invalid payload length.
- For non-command chaining code path, only copy the required length of payload.

### DIFF
--- a/src/driver/amdxdna/aie2_message.c
+++ b/src/driver/amdxdna/aie2_message.c
@@ -708,8 +708,11 @@ int aie2_execbuf(struct amdxdna_ctx *ctx, struct amdxdna_sched_job *job,
 	op = amdxdna_cmd_get_op(cmd_abo);
 	switch (op) {
 	case ERT_START_CU:
-		if (unlikely(payload_len > sizeof(req.ebuf.payload)))
-			XDNA_DBG(xdna, "Invalid ebuf payload len: %d", payload_len);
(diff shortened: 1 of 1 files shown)


maxzhen (approved):

src/driver/amdxdna/aie2_message.c
```diff
@@ -707,19 +707,26 @@ int aie2_execbuf(struct amdxdna_ctx *ctx, struct amdxdna_sched_job *job,
 
 	op = amdxdna_cmd_get_op(cmd_abo);
 	switch (op) {
-	case ERT_START_CU:
-		if (unlikely(payload_len > sizeof(req.ebuf.payload)))
-			XDNA_DBG(xdna, "Invalid ebuf payload len: %d", payload_len);
+	case ERT_START_CU: {
```
mamin506: In the Linux coding style, all the example of "case" in switch statement doesn't use curly bracket. The coding style doesn’t forbid the use of curly brackets. But to keep the coding style consistent in our code, please don't use curly bracket in this case.

We can use curly bracket on demand, for example, the rest of the cases here use curly bracket for local variable "payload".

mamin506 (approved):
The code looks good to me. Just a minor suggestion about coding style.

src/driver/amdxdna/aie2_message.c
NishadSaraf: Fixed

PR #487, opened 2025-04-21
The newer mailbox_info doesn't have iohub address, it will be empty as 0 and no need to write/read it.

xdavidz:
also enable the stress testing in xrt_test

### DIFF
--- a/src/driver/amdxdna/amdxdna_mailbox.c
+++ b/src/driver/amdxdna/amdxdna_mailbox.c
@@ -162,6 +162,23 @@ static u32 mailbox_reg_read(struct mailbox_channel *mb_chann, u32 mbox_reg)
 	return readl(ringbuf_addr);
 }
 
+static inline void mailbox_irq_acknowledge(struct mailbox_channel *mb_chann)
+{
+	if (mb_chann->iohub_int_addr)
+		mailbox_reg_write(mb_chann,
+				  mb_chann->iohub_int_addr,
+				  0);
+}
+
+static inline u32 mailbox_irq_status(struct mailbox_channel *mb_chann)
+{
+	if (mb_chann->iohub_int_addr)
+		return mailbox_reg_read(mb_chann,
+					mb_chann->iohub_int_addr);
+
+	return 0;
+}
+
 static inline void
 mailbox_set_headptr(struct mailbox_channel *mb_chann, u32 headptr_val)
 {
@@ -422,7 +439,7 @@ static void mailbox_rx_worker(struct work_struct *rx_work)
 	}
 
 again:
-	mailbox_reg_write(mb_chann, mb_chann->iohub_int_addr, 0);
+	mailbox_irq_acknowledge(mb_chann);
 
 	while (1) {
 		/*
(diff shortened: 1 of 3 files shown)


houlz0507 (approved):

src/driver/amdxdna/amdxdna_mailbox.c
```diff
@@ -162,6 +162,23 @@ static u32 mailbox_reg_read(struct mailbox_channel *mb_chann, u32 mbox_reg)
 	return readl(ringbuf_addr);
 }
 
+static inline void mailbox_irq_acknowledge(struct mailbox_channel *mb_chann)
+{
+	if (mb_chann->iohub_int_addr)
+		mailbox_reg_write(mb_chann,
+				  mb_chann->iohub_int_addr,
+				  0);
```
mamin506:
```suggestion
	if (!mb_chann->iohub_int_addr)
	    return;
	
	mailbox_reg_write(mb_chann, mb_chann->iohub_int_addr, 0);
```

src/driver/amdxdna/amdxdna_mailbox.c
```diff
@@ -490,15 +507,14 @@ static void mailbox_polld_handle_chann(struct mailbox_channel *mb_chann)
 	if (mb_chann->bad_state)
 		return;
 
-	iohub = mailbox_reg_read(mb_chann, mb_chann->iohub_int_addr);
+	iohub = mailbox_irq_status(mb_chann);
 	if (!iohub)
```
mamin506: This change doesn't work with the mb_chann->iohub_int_addr == NULL code path. The iohub will always be 0.
This function will always return without doing anything. It makes turbo mode not working when mb_chann->iohub_int_addr == NULL.

test/xrt_test/xrt_test.cpp
```diff
@@ -432,8 +432,8 @@ std::vector<test_case> test_list {
   test_case{ "npu3 xrt remote_barrier", TEST_xrt_umq_remote_barrier, {} },
   test_case{ "npu3 xrt nop", TEST_xrt_umq_nop, {} },
   test_case{ "npu3 xrt df_bw", TEST_xrt_umq_df_bw, {} },
-  test_case{ "npu3 xrt stress - start", TEST_xrt_stress_start, {32} },
-  test_case{ "npu3 xrt stress - hwctx", TEST_xrt_stress_hwctx, {2} }, //upto 2 now
+  test_case{ "npu3 xrt stress - start", TEST_xrt_stress_start, {128} },
+  test_case{ "npu3 xrt stress - hwctx", TEST_xrt_stress_hwctx, {32} }, //upto 2 now
```
mamin506: The "//upto 2 now" comment can be deleted.

mamin506 (changes requested):

test/xrt_test/xrt_test.cpp
xdavidz: fixed, thanks

mamin506 (approved):

PR #496, opened 2025-04-23
add a note for dma_alloc_noncoherent

mamin506:


mamin506:
retest this please

src/driver/amdxdna/aie2_error.c
```diff
@@ -310,6 +310,16 @@ int aie2_error_async_events_alloc(struct amdxdna_dev_hdl *ndev)
 	if (!events)
 		return -ENOMEM;
 
+	/*
+	 * Note: We test the behavior of dma_alloc_noncoherent() on 6.13 kernel.
```
maxzhen, line 314: The limitation is only applicable when IOMMU is off. We probably should clarify it.

amd-akshatah:
retest this please

src/driver/amdxdna/aie2_error.c
mamin506: The 4MB limit is true for IOMMU on and off. The allocation always goes into __alloc_frozen_pages_noprof() and failed if the order is larger than MAX_PAGE_ORDER.

maxzhen: Good to know!

PR #497, opened 2025-04-25
Adding npu3 test resnet50 and single col preemption

hlaccabu:


xdavidz:
@hlaccabu  you have manually test the xrt_test right?

xdavidz (approved):

PR #500, opened 2025-04-28
Handle event trace logging in buffer overflow case.

VinitAmd:
In case of buffer over flow event traces would be overridden by new logs,
but the event info should be logged with respect to log time counter.

### DIFF
--- a/src/driver/amdxdna/aie2_event_trace.c
+++ b/src/driver/amdxdna/aie2_event_trace.c
@@ -58,44 +58,47 @@ static int aie2_is_event_trace_supported_on_dev(struct amdxdna_dev_hdl *ndev)
 
 static u32 aie2_get_trace_event_content(struct event_trace_req_buf *trace_req_buf)
 {
+	u32 head_ptr, tail_ptr, head_ptr_wrap, tail_ptr_wrap;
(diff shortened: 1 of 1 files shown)


EugeneHiew:

mamin506:
ok to test

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -58,44 +58,47 @@ static int aie2_is_event_trace_supported_on_dev(struct amdxdna_dev_hdl *ndev)
...
-		memcpy(kern_buf + offset, (u8 *)(sys_buf + head_ptr), log_size);
-
-		offset += log_size;
-		total_log_size += log_size;
-		head_ptr = (head_ptr + log_size) % LOG_RB_SIZE;
-	} while (head_ptr < tail_ptr_wrap);
-
-	return total_log_size;
+	/* Handle buffer overflow case, dump all log w.r.t timestamp */
+	if (log_size > LOG_RB_SIZE) {
```
vengutta18, line 82: How do we hit this case? How come the tail_ptr going beyond the LOG_RB_SIZE?

vengutta18 (approved):
LGTM

PR #518, opened 2025-05-07
CR-1223916 Add support for get_info_array ioctl to device.cpp

rchane:
Add support for get_info_array ioctl to use when querying HW context information. This allows xrt-smi to get additional information (QOS, InstrBO) for each HW context and for the shim to be able to reallocate a buffer if the driver deems it necessary. In the event that get_info_array ioctl is not supported, we fall back upon the legacy get_info ioctl.

Tested on STX.

### DIFF
--- a/src/shim/device.cpp
+++ b/src/shim/device.cpp
@@ -234,26 +234,83 @@ struct partition_info
     if (key != key_type::aie_partition_info)
       throw xrt_core::query::no_such_key(key, "Not implemented");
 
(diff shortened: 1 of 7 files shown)


mamin506:
ok to test

maxzhen (changes requested):
Please don't add empty implementation in virtio platform code. Just let it fail loudly in parent class.

maxzhen (approved):

mamin506:
retest this please

PR #521, opened 2025-05-08
Enhance query telemetry implementation

NishadSaraf:
Enhance query telemetry implementation.

### DIFF
--- a/src/driver/amdxdna/aie2_pci.c
+++ b/src/driver/amdxdna/aie2_pci.c
@@ -1005,26 +1005,24 @@ out:
 static int aie2_query_telemetry(struct amdxdna_client *client,
 				struct amdxdna_drm_get_info *args)
 {
+	struct amdxdna_drm_query_telemetry_header header, *tmp;
 	struct amdxdna_dev *xdna = client->xdna;
-	struct amdxdna_dev_hdl *ndev;
+	size_t aligned_sz, offset;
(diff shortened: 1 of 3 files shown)


src/driver/amdxdna/aie2_pci.c
```diff
@@ -1005,43 +1005,70 @@ static int aie2_query_ctx_status(struct amdxdna_client *client,
...
-	int ret;
 
 	if (!access_ok(u64_to_user_ptr(args->buffer), args->buffer_size)) {
 		XDNA_ERR(xdna, "Failed to access buffer size %d", args->buffer_size);
 		return -EFAULT;
 	}
 
-	if (copy_from_user(&type, u64_to_user_ptr(args->buffer), sizeof(type))) {
+	header = (struct amdxdna_drm_query_telemetry_header *)u64_to_user_ptr(args->buffer);
+	if (copy_from_user(&type, &header->type, sizeof(type))) {
```
mamin506: "header" is a user pointer. "header->type" directly dereference it. This will work without issue on x86. But this is not allow in upstreaming point of view.

maxzhen (changes requested):
Had offline discussion with @NishadSaraf. The design needs some tweak.

maxzhen (approved):

mamin506 (approved):

PR #528, opened 2025-05-11
Add dynamic buffer size and event category.

VinitAmd:
Add support to set event trace buffer size and event category from debugfs.

### DIFF
--- a/src/driver/amdxdna/aie2_debugfs.c
+++ b/src/driver/amdxdna/aie2_debugfs.c
@@ -216,20 +216,60 @@ static int aie2_dpm_level_get(struct seq_file *m, void *unused)
 
 AIE2_DBGFS_FOPS(dpm_level, aie2_dpm_level_get, aie2_dpm_level_set);
 
-static ssize_t aie2_event_trace_write(struct file *file, const char __user *ptr,
+static ssize_t aie2_event_trace_write(struct file *file, const char __user *buf,
 				      size_t len, loff_t *off)
 {
 	struct amdxdna_dev_hdl *ndev = file_to_ndev_rw(file);
-	bool state;
-	int ret;
+	u32 enable = 0, buf_size = 0, event_type = 0;
+	char *kbuf, *token, *key, *val;
 
-	ret = kstrtobool_from_user(ptr, len, &state);
-	if (ret) {
-		XDNA_ERR(ndev->xdna, "Invalid input value: %d", state);
-		return ret;
+	kbuf = kzalloc(len + 1, GFP_KERNEL);
+	if (!kbuf)
+		return -ENOMEM;
+
+	if (copy_from_user(kbuf, buf, len)) {
+		kfree(kbuf);
+		return -EFAULT;
 	}
 
-	aie2_assign_event_trace_state(ndev, state);
+	kbuf[len] = '\0';
+	token = strsep(&kbuf, " ");
+
+	while (token) {
+		key = strsep(&token, "=");
+		val = token;
+		int ret;
+
+		if (key && val) {
+			if (strcmp(key, "enable") == 0) {
+				ret = kstrtouint(val, 10, &enable);
+				if (ret) {
+					XDNA_INFO(ndev->xdna, "invalid enable value %u", enable);
+					return ret;
+				}
+			} else if (strcmp(key, "size") == 0) {
+				buf_size = memparse(val, NULL);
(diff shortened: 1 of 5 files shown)


EugeneHiew:

maxzhen:
ok to test

src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -263,20 +264,33 @@ static int aie2_dpm_level_get(struct seq_file *m, void *unused)
...
-		return ret;
+	if (len <= 1 || len > EVENT_TRACE_INPUT_COUNT) {
+		XDNA_ERR(ndev->xdna, "Invalid input size: %lu", len);
+		return -EINVAL;
+	}
+
+	if (copy_from_user(cmd_input, buf, len))
+		return -EFAULT;
+
+	ret = sscanf(cmd_input, "%u %u %u", &temp_buff[0], &temp_buff[1],
```
vengutta18: I thnk this might be error prone, maybe we should have token parsing
like "enable=1 size=4K mask=0xFFF"

vengutta18: Maybe something like below: 

You can also compare the value of the sizes to the valid sizes by maintaining a list , same w.r.t to the masks.


 token = strsep(&buf, " ");
    while (token) {
        key = strsep(&token, "=");
        val = token;
        if (key && val) {
            if (strcmp(key, "enable") == 0)
                // parse and set 'enable' value
            else if (strcmp(key, "size") == 0)
                // parse and set 'size' value
            else if (strcmp(key, "mask") == 0)
                // parse and set 'mask' value
            // handle other keys
        }
        token = strsep(&buf, " ");
    }

mamin506:
retest this please

vengutta18 (approved):

mamin506 (approved):
Did not got chance to look at all the implementation details in aie2_event_trace.c and aie2_debugfs.c. Overall, the changes and the design are looking good to me.

src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -242,7 +282,10 @@ static int aie2_event_trace_show(struct seq_file *m, void *unused)
 		seq_puts(m, "Event trace is enabled\n");
 	else
 		seq_puts(m, "Event trace is disabled\n"
-						"echo 1 > To enable event trace\n");
+						"echo \"enable=1 size=1K eventtype=0xffffff\" > Follow given input format to enable\n"
+						"echo \"enable=[0, 1]\" > enable=1 to enable, enable=0 to disable\n"
+						"echo \"size=[1K, 2K, 4K...512K to 1M\" buffer size should be pow of 2\n"
+						"echo \"eventtype=[0x000000FF, 0x0000FFFF, 0x00FFFFFF, 0xFFFFFFFF\" > are supported now\n");
```
marcusr-amd, line 288: Is this comment out of date?

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -56,21 +57,47 @@ static int aie2_is_event_trace_supported_on_dev(struct amdxdna_dev_hdl *ndev)
...
+		return -EINVAL;
+	}
+
+	if (buf_size <= 1024 || buf_size > 1024 * 1024 ||
+	    (buf_size & (buf_size - 1)) != 0) {
+		XDNA_ERR(ndev->xdna, "Invalid buffer size[1K-1M]: %u", buf_size);
+		return -EINVAL;
+	}
+
+	if (category < 0xFF) {
```
marcusr-amd, line 75: Where does this limitation come from?
Event category should support any 32 bit number.

marcusr-amd: I guess you could argue that it doesn't make sense to be 0 if event tracing isn't enabled.  But then that begs the question if we need a separate enable or we should just use non-zero to be the enable.

src/driver/amdxdna/aie2_debugfs.c
marcusr-amd: And if we're nitpicking we should call it event category instead of event type.  As it is defined today there are 32 categories each of which can define 256 events for a total of 8k possible events.

VinitAmd: Pls follow my reply on teams

PR #535, opened 2025-05-13
enable runtime auto suspend/resume; refine the design

mamin506:
1. Enable runtime auto suspend in 5 seconds if there is no outstanding command
2. Resume FW / SMU when the first command is submitted instead of the old design which resume when client open
3. There is NO performance degradation with the shim_test
4. Comment out event trace in suspend/resume code path for now. I will sync up with @VinitAmd and @NishadSaraf  offline.


maxzhen (approved):

mamin506:
retest this please

vengutta18 (approved):

gyang1099 (approved):

PR #538, opened 2025-05-14
Remove suspend/resume from debugfs

vengutta18:
Autosuspend is enabled so remove suspend/resume from debugfs.


mamin506:
retest this please

PR #549, opened 2025-05-19
Handle auto suspend in event tracing.

VinitAmd:


EugeneHiew:

mamin506:
ok to test

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -336,6 +336,9 @@ static void aie2_update_event_trace_state(struct amdxdna_dev_hdl *ndev, bool sta
 		return;
 	}
 
+	if (req_buf->suspend)
+		goto done;
+
```
mamin506: This is not needed.

When a debugfs was opened, the device will resume. So this condition will always be false.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -393,13 +406,73 @@ void aie2_config_event_trace(struct amdxdna_dev_hdl *ndev, u32 enable,
...
+	if (!ndev->event_trace_req) {
+		XDNA_DBG(ndev->xdna, "Event trace req buffer is not allocated!");
+		return;
+	}
+
+	if (ndev->event_trace_req->suspend == suspend) {
+		XDNA_DBG(ndev->xdna, "Device already in suspend: %d state",
+			 suspend);
+		return;
+	}
```
mamin506: Do you see this condition to be true? This is more like a WARN_ON().
By design, the suspend/resume flow should only happen once. If you see this happen, let me know. I will debug the flow.

src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -264,11 +265,11 @@ static ssize_t aie2_event_trace_write(struct file *file, const char __user *buf,
 	}
 
 	XDNA_DBG(ndev->xdna, "Event trace config: %u, %u, %u",
```
vengutta18: You can update this to hex format 0x%08x instead of u for event_category?

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -16,6 +16,7 @@ struct event_trace_req_buf {
 	struct amdxdna_dev_hdl   *ndev;
 	struct workqueue_struct  *wq;
 	struct work_struct       work;
+
```
vengutta18: nit: remove it

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -319,14 +323,10 @@ static int aie2_stop_event_trace_send(struct amdxdna_dev_hdl *ndev)
...
+	struct event_trace_req_buf *req_buf = ndev->event_trace_req;
 	int err;
 
-	if (!aie2_is_event_trace_supported_on_dev(ndev)) {
-		XDNA_DBG(ndev->xdna, "Event trace is not supported on this device");
-		return;
-	}
-
-	if (!ndev->event_trace_req) {
+	if (!req_buf) {
```
vengutta18: can this happen?

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -393,13 +406,73 @@ void aie2_config_event_trace(struct amdxdna_dev_hdl *ndev, u32 enable,
...
+						bool suspend)
+{
+	int err;
+
+	if (!aie2_is_event_trace_supported_on_dev(ndev)) {
+		XDNA_DBG(ndev->xdna, "Event trace is not supported on this device");
+		return;
+	}
+
+	if (!ndev->event_trace_req) {
```
vengutta18: can this req_buf/event_trace_req ever be NULL after the init?

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -393,13 +406,73 @@ void aie2_config_event_trace(struct amdxdna_dev_hdl *ndev, u32 enable,
...
+
+	if (suspend) {
+		err = aie2_stop_event_trace_send(ndev);
+		if (err)
+			return;
+
+		goto done;
+	}
+
+	err = aie2_start_event_trace_send(ndev);
```
vengutta18: The idea of stopping the event trace for suspend is okay, but during resume shouldn't we use the original event trace state : "req_buf->enabled"  and enable/disable based on that?

src/driver/amdxdna/aie2_event_trace.c
VinitAmd: Primitive and user defined variables are separated with empty line, this looks like standard practice.
Just followed aie2_pci.h definitions :)

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -393,13 +403,67 @@ void aie2_config_event_trace(struct amdxdna_dev_hdl *ndev, u32 enable,
...
+
+	XDNA_DBG(ndev->xdna, "prev autosuspend state: %s new state: %s",
+		 get_current_state_to_char(ndev->event_trace_req->suspend),
+		 get_current_state_to_char(suspend));
+
+	/*
+	 * event trace handler should know the driver suspend state
+	 * before enable/disable via debugfs.
+	 */
+	ndev->event_trace_req->suspend = suspend;
```
mamin506: Why we need "ndev->event_trace_req->suspend"?
The comment is confusing to me. The check in enable/disable code path is deleted. 

It looks like the only usage of this is to print message. If this is the case, I suggest delete from the event_trace_req. We don't need to maintain another status.

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -393,13 +403,67 @@ void aie2_config_event_trace(struct amdxdna_dev_hdl *ndev, u32 enable,
...
 
-	XDNA_DBG(ndev->xdna, "buf size 0x%08x category 0x%08x",
-		 buff_size, category);
 	ndev->event_trace_req->dram_buffer_size = buff_size;
 	ndev->event_trace_req->event_trace_category = category;
 	aie2_assign_event_trace_state(ndev, true);
 }
 
+static void aie2_event_trace_handle_autosuspend(struct amdxdna_dev_hdl *ndev,
+						bool suspend)
```
mamin506: From the readability point of view, can we avoid using "single function + status" here?

Below coding style is easier to read and we can change any code path without worry impact the other one (long term maintenance),
aie2_event_trace_suspend()
{
  // acquire lock
  // sanity check
  // if event trace is enable, stop event trace
  // release lock
}

aie2_event_trace_resume()
{
  // acquire lock
  // sanity check
  // if event trace is enable, start event trace
  // release lock
}

In my test, the current code of this function is correct. But coupling suspend and resume logic in one function make it harder to read and error-prone for long term maintenance.

mamin506 (changes requested):
In my test, some debugfs usage is not working.

For example, 
`echo "enable=1" > event_trace`, shows error "-bash: echo: write error: Invalid argument"
`echo "enable=1 size=1K eventtype=0xffffff" > event_trace`, the dmesg shows error "aie2_validate_event_trace_config: Invalid buffer size[1K-1M]: 1024". But, the error does show that 1K is valid.

VinitAmd:
> In my test, some debugfs usage is not working.
> 
> For example, `echo "enable=1" > event_trace`, shows error "-bash: echo: write error: Invalid argument" `echo "enable=1 size=1K eventtype=0xffffff" > event_trace`, the dmesg shows error "aie2_validate_event_trace_config: Invalid buffer size[1K-1M]: 1024". But, the error does show that 1K is included, as '[' is used on the left endpoint.

---> Please check the code "eventtype" is not a key, instead "echo enable=1 size=1K category=0xffffff " .
---> `echo "enable=1" is also not key, 3 keys required.
This is caused error for you, update your local code base.

src/driver/amdxdna/aie2_event_trace.c
VinitAmd: Yes, kept it for debugging, missed to remove text.

src/driver/amdxdna/aie2_event_trace.c
VinitAmd: I have other opinion on the improvement side, will discuss.

vengutta18 (approved):

PR #562, opened 2025-05-29
The buffer for MGMT FW cannot span 64MB boundary

mamin506:
This is a fix for AIESW-3960. The FW required all buffers used on mgmt channel to be 4MB aligned and not span 64MB boundary.

The solution of this PR is not perfect. It simply over-allocated 4MB buffer with dma_alloc_noncoherent() API, without considering the actual needs of the buffer (It only needs a few pages). 

DMA APIs doesn't provide alignment to the caller. So, maybe still need more efforts to investigate what is the best approach.

This PR can be a quick fix for 1.5 release.


mamin506:
Sync with the FW team, the restrict is only not span 64MB boundary. Allocate 4MB for each buffer is an over-kill. I will push a new commit for a better solution.

NishadSaraf (approved):

PR #572, opened 2025-06-04
Invalidate cache before copying trace event messages.

VinitAmd:


EugeneHiew:

amd-kirkirov:
ok to test

mamin506:
ok to test

src/driver/amdxdna/aie2_event_trace.c
```diff
@@ -250,10 +263,13 @@ static int aie2_event_trace_alloc(struct amdxdna_dev_hdl *ndev)
...
 	req_buf->buf = dma_alloc_noncoherent(xdna->ddev.dev, req_buf->dram_buffer_size,
 					     &req_buf->dram_buffer_address,
-					     DMA_BIDIRECTIONAL, GFP_KERNEL);
+					     DMA_FROM_DEVICE, GFP_KERNEL);
+
 	if (!req_buf->buf)
 		return -ENOMEM;
 
+	req_buf->rb_head = 0;
+	drm_clflush_virt_range(req_buf->buf, req_buf->dram_buffer_size);
```
vengutta18, line 272: looks like this is explicit flush operation, I think we can use dma_alloc_coherent to let the DMA APIs handle the cache operations for us.

Can we use dma_alloc_coherent instead?

@VinitAmd  @mamin506

vengutta18 (changes requested):

PR #578, opened 2025-06-06
Adding npu3 pf driver to xrt_plugin package

hlaccabu:
Tested with npu3 sriov package install, successful.


xdavidz (changes requested):
@hlaccabu @mamin506 no, we don't want to add change in the open sourced git repo right?

@hlaccabu  please talk to Min offline and get this sorted.

PR #580, opened 2025-06-09
dumping app health data on the exec command data section when timedout

mamin506:
Add driver support to dumping health data into exec cmd packet.

The driver will collect context health report when TDR thread detected device is timedout.
When mailbox was destroyed, it will call commands callback with NULL data pointer. Then the context will dump health data into the command packet and this command will be marked as ERT_CMD_STATE_TIMEOUT. The user space runtime can get this health report.

If a context has multiple out-standing commands, only dump health data on the first command. For other commands, their status will be ERT_CMD_STATE_ABORT without health data.


maxzhen (approved):

mamin506:
retest this please

mamin506:
retest this please

PR #597, opened 2025-06-16
Add runtime cfg support to fw logging

VinitAmd:


EugeneHiew:

src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -380,6 +380,82 @@ static int aie2_dram_logging_show(struct seq_file *m, void *unused)
...
+				XDNA_ERR(ndev->xdna, "invalid log level %u",
+					 loglevel);
+				ret = -EINVAL;
+				goto out;
+			}
+
+			ret = aie2_set_log_level(ndev, loglevel);
+			if (ret) {
+				XDNA_ERR(ndev->xdna, "Failed to set log level: %d", ret);
+				ret = -EINVAL;
```
vengutta18: ret update not needed, we can pass the return of aie2_set_log_level, so you can remove this.

src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -380,6 +380,82 @@ static int aie2_dram_logging_show(struct seq_file *m, void *unused)
...
+{
+	struct amdxdna_dev_hdl *ndev = file_to_ndev_rw(file);
+	char *kbuf, *token, *key, *val;
+	u32 loglevel;
+	int ret;
+
+	kbuf = kzalloc(len + 1, GFP_KERNEL);
+	if (!kbuf)
+		return -ENOMEM;
+
```
mamin506: I had some concerns about directly use 'len' here.  Usually, if the size is from user, Linux driver should check if the size make sense or not. For example, check the upper limit of the size.

In this code, it looks like we are using the nature upper limit, which is 4MB, of kzalloc() with GFP_KERNEL.
@VinitAmd, please confirm 4MB - 1 bytes is the limit of 'len' in your design.

mamin506 (commented):
Overall this look good to me. I just review the debugfs and the pci.h change.

mamin506 (approved):

PR #599, opened 2025-06-17
Runner recipe migrations

aktondak:
run recipe migration update

Note : The throughput test was never tested with elf flow on phoenix. Run recipe if elf only thus this required some plumbing.
I ran on birman-16 and following is the result :

```
xrt-smi validate -r throughput
WARNING: User doesn't have admin permissions to set performance mode. Running validate in Default mode
Validate Device           : [0000:c3:00.1]
    Platform              : NPU Phoenix
    Power Mode            : Default
-------------------------------------------------------------------------------
Test 1 [0000:c3:00.1]     : throughput
    Details               : Average throughput: 24058.0 ops
    Test Status           : [PASSED]

```
which matches with non-elf phoenix run.


aktondak:
retest this please

aktondak:
retest this please

mamin506:
retest this please

PR #600, opened 2025-06-17
Add thread test for concurrency

hlaccabu:


mamin506:
retest this please

xdavidz:
> retest this please
@mamin506 
Since we are here. There are 2 multi_thread code in xrt_test.cpp and shim_test.cpp. Should we combine these 2 to 1 common file?

mamin506:
> > retest this please
> > @mamin506
> > Since we are here. There are 2 multi_thread code in xrt_test.cpp and shim_test.cpp. Should we combine these 2 to 1 common file?

It is great to have separate multi thread implementation. Coupling shim_test and xrt_test might not be a good idea for now.
We can consider share common code when these two test cases are run by a single pipeline.

PR #605, opened 2025-06-19
Latency test migration

aktondak:
This PR converts the latency test to run recipe format. The recipe json runs 1 kernel for 10,000 iterations and reports the latency calculated by runner.
This PR also updates the xrt submodule


mamin506:
retest this please

PR #609, opened 2025-06-20
Implement polling for pulling FW log continuously.

VinitAmd:


src/driver/amdxdna/aie2_logging.c
```diff
@@ -203,7 +219,11 @@ static irqreturn_t log_buffer_irq_handler(int irq, void *data)
 
 	trace_mbox_irq_handle("DRAM_LOG_BUFFER", irq);
 	clear_logging_msix(ndev);
+
+	mutex_lock(&ndev->logging_req->wq_lock);
```
vengutta18: you can't use mutex here, irq! Use spinlock if really needed

vengutta18: Also to avoid synchronization aspects check mod_delayed_work() once to see if you can use that here.

vengutta18: Also see if this polling and interrupt should co-exist or mutual exclusive! 
If they should co-exist you can probably use mod_delayed_work()

src/driver/amdxdna/aie2_logging.c
```diff
@@ -189,6 +193,18 @@ static void aie2_print_log_buffer_data(struct amdxdna_dev_hdl *ndev)
 	}
 }
 
+static void poll_timer_callback(struct timer_list *timer)
+{
+	struct logging_req_buf *req_buf;
+
+	req_buf = container_of(timer, struct logging_req_buf, poll_timer);
+
+	mutex_lock(&req_buf->wq_lock);
```
vengutta18: no mutex in timer callbacks too.
Actually we don't need timer maybe. Check other alternatives as suggested below.

vengutta18 (changes requested):
Few comments above.

src/driver/amdxdna/aie2_logging.c
VinitAmd: Ahh...surprisingly i didn't get BUG_ON(!in_atomic())

VinitAmd:
retest this please

src/driver/amdxdna/aie2_logging.c
```diff
@@ -203,7 +219,11 @@ static irqreturn_t log_buffer_irq_handler(int irq, void *data)
 
 	trace_mbox_irq_handle("DRAM_LOG_BUFFER", irq);
 	clear_logging_msix(ndev);
-	queue_work(ndev->logging_req->wq, &ndev->logging_req->work);
+
+	/* Only queue if not already pending */
+	if (!work_pending(&ndev->logging_req->work))
+		queue_work(ndev->logging_req->wq, &ndev->logging_req->work);
```
maxzhen: Is this race free? I'd expect the check and queue operations to be done as one atomic operation. What will happen if after you checked, the pending work is completed, then you'll miss a queue-work? What would go wrong if you just queue blindly?

src/driver/amdxdna/aie2_logging.c
```diff
@@ -189,6 +192,19 @@ static void aie2_print_log_buffer_data(struct amdxdna_dev_hdl *ndev)
...
 
+static void poll_timer_callback(struct timer_list *timer)
+{
+	struct logging_req_buf *req_buf;
+
+	req_buf = container_of(timer, struct logging_req_buf, poll_timer);
+
+	/* Only queue if not already pending */
+	if (!work_pending(&req_buf->work))
+		queue_work(req_buf->wq, &req_buf->work);
```
maxzhen: Similar questions as above.

src/driver/amdxdna/aie2_logging.c
VinitAmd: queue_work() is already thread-safe and internally synchronized. Linux kernel calls  queue_work_on(WORK_CPU_UNBOUND, wq, work); internally where it uses local_irq_save/restore and WORK_STRUCT_PENDING_BIT check .
What will happen if after you checked, the pending work is completed, then you'll miss a queue-work?
       -->There should be no log miss ..in next callback log would be read from prev head since head is not moved.

 What would go wrong if you just queue blindly?
      -->queue_work() is safe and will not queue the same work item twice, it is good practice to avoid unnecessary queueing call.

maxzhen: It is OK for queue_work to internally check pending and queue. It is atomic. It is not OK for you to do this check and queue without holding a lock since it is not atomic. Please remove the check.

maxzhen (changes requested):
Checking pending queue work, then queue work without lock is not OK.

src/driver/amdxdna/aie2_logging.c
vengutta18: @VinitAmd  @maxzhen 
I think mod_delayed_work takes care of canceling the pending queued work and the synchronization part of queuing the work in process/interrupt context.

I'd prefer #620
but open to suggestions.

VinitAmd: Please review https://github.com/amd/xdna-driver/pull/620

maxzhen (approved):

NishadSaraf (approved):

PR #611, opened 2025-06-21
hmm invalidate should wait forever

mamin506:
Based on dma_resv_wait_timeout() document, "Returns -ERESTARTSYS if interrupted, 0 if the wait timed out, or greater than zero on success."

aie2_hmm_invalidate() should only exit the loop on success.

### DIFF
--- a/src/driver/amdxdna/aie2_ctx.c
+++ b/src/driver/amdxdna/aie2_ctx.c
@@ -1066,17 +1066,11 @@ int aie2_cmd_wait(struct amdxdna_ctx *ctx, u64 seq, u32 timeout)
 void aie2_hmm_invalidate(struct amdxdna_gem_obj *abo, unsigned long cur_seq)
 {
 	struct drm_gem_object *gobj = to_gobj(abo);
-	long ret;
 
 	/*
 	 * Must wait forever, otherwise, memory was unmapped then FW might crash.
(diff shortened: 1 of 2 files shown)


src/driver/amdxdna/aie2_ctx.c
```diff
@@ -1065,12 +1065,18 @@ int aie2_cmd_wait(struct amdxdna_ctx *ctx, u64 seq, u32 timeout)
...
-	if (!ret || ret == -ERESTARTSYS)
-		XDNA_ERR(xdna, "Failed to wait for bo, ret %ld", ret);
+	/*
+	 * Must wait forever, otherwise, memory was unmapped then FW might crash.
+	 * In case FW not response, TDR will terminal context execution and unref all BOs.
+	 * This loop will eventually exit.
+	 */
+	while (1) {
+		ret = dma_resv_wait_timeout(gobj->resv, DMA_RESV_USAGE_BOOKKEEP,
+					    true, MAX_SCHEDULE_TIMEOUT);
```
houlz0507: Instead of while (1), I suggest dma_resv_wait_timeout(..., false, MAX_SCHEDULE_TIMEOUT);

maxzhen: Agreed

PR #613, opened 2025-06-23
ve2: add ctx_config() support

rajkumar-xilinx:
Added hardware context config support in ve2 driver. It supports the following buffers configuration at runtime.
-Debug buffer
-DTrace buffer
-FW log buffer

### DIFF
--- a/src/driver/amdxdna/ve2_hwctx.c
+++ b/src/driver/amdxdna/ve2_hwctx.c
@@ -613,3 +613,209 @@ void ve2_hwctx_fini(struct amdxdna_ctx *hwctx)
 	ve2_free_hsa_queue(xdna, &hwctx->priv->hwctx_hsa_queue);
 	kfree(hwctx->priv);
 }
+
+static int ve2_hwctx_config_dtrace_buf(struct amdxdna_ctx *hwctx, u64 base_paddr,
+				       struct uc_info_entry uc_info[], bool attach)
+{
+	struct device *aie_dev = hwctx->priv ? hwctx->priv->aie_dev : NULL;
+	struct amdxdna_dev *xdna = hwctx->client->xdna;
+	struct handshake hs = { 0 };
+	u32 uc_index = 0;
+	u32 prev_sz = 0;
+	int ret = 0;
+	u64 paddr;
+
+	if (!aie_dev) {
+		XDNA_ERR(xdna, "%s aie device handle not found", __func__);
+		return -EINVAL;
+	}
+
+	for (u32 col = 0; col < hwctx->num_col; col++, uc_index++) {
+		if (attach) {
+			if (uc_info[uc_index].size == 0)
+				continue;
+			paddr = base_paddr + prev_sz;
+			hs.trace.dtrace_addr_high = upper_32_bits((u64)paddr);
+			hs.trace.dtrace_addr_low = lower_32_bits((u64)paddr);
+			prev_sz += uc_info[uc_index].size;
+			XDNA_DBG(xdna, "col[%d].paddr high = %u, low = %u, size = %u\n", col,
+				 hs.trace.dtrace_addr_high, hs.trace.dtrace_addr_low,
+				 uc_info[uc_index].size);
+		}
+
+		ret = aie_partition_write_privileged_mem(aie_dev, CERT_HANDSHAKE_OFF(col) +
+							 offsetof(struct handshake,
+								  trace.dtrace_addr_high),
+							 sizeof(hs.trace), (void *)&hs.trace);
+		if (ret < 0)
+			return ret;
+	}
+
+	return ret;
+}
+
+static int ve2_hwctx_config_debug_buf(struct amdxdna_ctx *hwctx, u64 base_paddr,
+				      struct uc_info_entry uc_info[], bool attach)
+{
+	struct device *aie_dev = hwctx->priv ? hwctx->priv->aie_dev : NULL;
+	struct amdxdna_dev *xdna = hwctx->client->xdna;
+	struct handshake hs = { 0 };
+	u32 uc_index = 0;
+	u32 prev_sz = 0;
+	int ret = 0;
+	u64 paddr;
+
+	if (!aie_dev) {
+		XDNA_ERR(xdna, "%s aie device handle not found", __func__);
+		return -EINVAL;
+	}
+
+	for (u32 col = 0; col < hwctx->num_col; col++, uc_index++) {
+		if (attach) {
+			if (uc_info[uc_index].size == 0)
+				continue;
+			paddr = base_paddr + prev_sz;
+			hs.dbg_buf.dbg_buf_addr_high = upper_32_bits((u64)paddr);
+			hs.dbg_buf.dbg_buf_addr_low = lower_32_bits((u64)paddr);
+			hs.dbg_buf.size = uc_info[uc_index].size;
+			prev_sz += uc_info[uc_index].size;
+			XDNA_DBG(xdna, "col[%d].paddr high = %u, low = %u, size = %u\n", col,
+				 hs.dbg_buf.dbg_buf_addr_high, hs.dbg_buf.dbg_buf_addr_low,
+				 uc_info[uc_index].size);
+		}
+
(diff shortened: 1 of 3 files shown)


src/driver/amdxdna/ve2_hwctx.c
```diff
@@ -613,3 +613,209 @@ void ve2_hwctx_fini(struct amdxdna_ctx *hwctx)
...
+	u32 prev_sz = 0;
+	int ret = 0;
+	u64 paddr;
+
+	if (!aie_dev) {
+		XDNA_ERR(xdna, "%s aie device handle not found", __func__);
+		return -EINVAL;
+	}
+
+	for (u32 col = 0; col < hwctx->num_col; col++, uc_index++) {
```
chvamshi-xilinx: This for loop is repeated in all three functions. Can you create a common function and pass corresponding offset to update. You can even use one function for all these updates by providing an enum/offset

rajkumar-xilinx: Hi @chvamshi-xilinx , each loop has it's own handshake structure variables to update. We can't pass these variables to other function thru offset or any enum. I am copying that code below for quick ref:

In Loop1: 
`hs.trace.dtrace_addr_high = upper_32_bits((u64)paddr);
 hs.trace.dtrace_addr_low = lower_32_bits((u64)paddr);`

In Loop2:
`hs.dbg_buf.dbg_buf_addr_high = upper_32_bits((u64)paddr);
 hs.dbg_buf.dbg_buf_addr_low = lower_32_bits((u64)paddr);
 hs.dbg_buf.size = uc_info[uc_index].size;`

In Loop3:
`hs.log_addr_high = upper_32_bits((u64)paddr);
 hs.log_addr_low = lower_32_bits((u64)paddr);
 hs.log_buf_size = uc_info[uc_index].size;`

src/driver/amdxdna/ve2_hwctx.c
```diff
@@ -613,3 +613,209 @@ void ve2_hwctx_fini(struct amdxdna_ctx *hwctx)
...
+	struct handshake hs = { 0 };
+	u32 uc_index = 0;
+	u32 prev_sz = 0;
+	int ret = 0;
+	u64 paddr;
+
+	if (!aie_dev) {
+		XDNA_ERR(xdna, "%s aie device handle not found", __func__);
+		return -EINVAL;
+	}
```
mamin506: In my understanding, aie_dev is fundation for ve2 driver. Is this is NULL, it must be a BUG.
If my understanding is correct, I suggest to do one of below,
1. Do nothing, let it crash to expose the NULL aie_dev bug.
2. Use WARN_ON(!aie_dev), when this pointer might be NULL in your code path after driver is loaded

src/driver/amdxdna/ve2_hwctx.c
```diff
@@ -613,3 +613,209 @@ void ve2_hwctx_fini(struct amdxdna_ctx *hwctx)
 	ve2_free_hsa_queue(xdna, &hwctx->priv->hwctx_hsa_queue);
 	kfree(hwctx->priv);
 }
+
+static int ve2_hwctx_config_dtrace_buf(struct amdxdna_ctx *hwctx, u64 base_paddr,
+				       struct uc_info_entry uc_info[], bool attach)
+{
+	struct device *aie_dev = hwctx->priv ? hwctx->priv->aie_dev : NULL;
```
mamin506: doesn't make sense to me. Why hwctx->priv is NULL? This is something must be initialized in ctx_init() callback. If you do have reason for this line, that must be a bug needs to be fixed somewhere else.

src/driver/amdxdna/ve2_hwctx.c
mamin506: @chvamshi-xilinx  I like your idea.

@rajkumar-xilinx , maybe you can consider below coding blocks.
``` c

static void ve2_fill_handshake(/* arguments */)
{
    switch (buf_type) {
    case AMDXDNA_FW_BUF_DEBUG:
    // Fill handshake in BUF debug way...
    break;
    case AMDXDNA_FW_BUF_TRACE:
    // Fill handshake in BUF trace way...
    break;
    ....
    }
}

static int ve2_hwctx_config_dbg_buf(struct amdxdna_ctx *hwctx, u64 base_paddr,
				    struct uc_info_entry uc_info[], bool attach)
{
    ...
    for (u32 col = 0; col < hwctx->num_col; col++, uc_index++) {
        ...
        ve2_fill_handshake(&hs, base_paddr, &us_info[uc_index] /* maybe you needs other arguments? */);
        ...
    }
    ...
}
```

From my point of view, this logic will always be the same for these tree type of buffers and for future buffers extension. 
@rajkumar-xilinx, You can think about if there is other ways to avoid duplicate the same for-loop logic. I will not accept current duplicate coding blocks. It looks bad for long term maintenance.

mamin506 (changes requested):

src/driver/amdxdna/ve2_hwctx.c
rajkumar-xilinx: Hi @mamin506 , yes, filling the handshake block is looks good, but after filling it, we need to invoke AIE driver API with offset info like below. Here, 3 to 5 arguments different in each buffer type. 

`ret = aie_partition_write_privileged_mem(aie_dev, CERT_HANDSHAKE_OFF(col) +
							 offsetof(struct handshake,
								  trace.dtrace_addr_high),
							 sizeof(hs.trace), (void *)&hs.trace);`

rajkumar-xilinx: Hi @mamin506 @chvamshi-xilinx I updated patch with your suggestions, please review and let me know, thank you.

src/driver/amdxdna/ve2_hwctx.c
```diff
@@ -613,3 +613,147 @@ void ve2_hwctx_fini(struct amdxdna_ctx *hwctx)
...
+	struct device *aie_dev = hwctx->priv->aie_dev;
+	struct handshake hs = { 0 };
+	int ret = 0;
+
+	WARN_ON(!aie_dev);
+
+	switch (buf_type) {
+	case AMDXDNA_FW_BUF_DEBUG:
+		if (attach) {
+			hs.dbg_buf.dbg_buf_addr_high = upper_32_bits((u64)paddr);
```
houlz0507: paddr is already u64

src/driver/amdxdna/ve2_hwctx.c
```diff
@@ -613,3 +613,147 @@ void ve2_hwctx_fini(struct amdxdna_ctx *hwctx)
...
+	u32 prev_buf_sz;
+	u64 buf_paddr;
+	int ret = 0;
+	u32 buf_sz;
+
+	/* Update fw's handshake shared memory with debug/trace buffer details */
+	switch (type) {
+	case DRM_AMDXDNA_CTX_ASSIGN_DBG_BUF:
+		mdata_abo = amdxdna_gem_get_obj(client, mdata_hdl, AMDXDNA_BO_DEV);
+		if (!mdata_abo || !mdata_abo->mem.kva) {
```
houlz0507: why do you check kva. It looks kva never used here?

src/driver/amdxdna/ve2_hwctx.c
```diff
@@ -613,3 +613,147 @@ void ve2_hwctx_fini(struct amdxdna_ctx *hwctx)
...
+		}
+
+		mdata = (struct fw_buffer_metadata *)(mdata_abo->mem.kva);
+		if (!mdata) {
+			XDNA_ERR(xdna, "No metadata defined for bo %lld type %d", mdata_hdl, type);
+			amdxdna_gem_put_obj(mdata_abo);
+			return -EINVAL;
+		}
+
+		for (u32 col = 0; col < hwctx->num_col; col++) {
```
houlz0507, line 726: could you remove duplicated code for different types, such as: get mdata_abo, above for loop etc.

src/driver/amdxdna/ve2_hwctx.c
rajkumar-xilinx: Hi Lizhi, it used in below code

mdata = (struct fw_buffer_metadata *)(mdata_abo->mem.kva);

houlz0507 (approved):

PR #618, opened 2025-06-25
Updating pipeline to grab bins from artifactory + bump umq ver

hlaccabu:


hlaccabu:
Wait for elf flow to update to newer cert

PR #620, opened 2025-06-26
[Debug/Review] Implement polling on FW logging with delayed_work.

VinitAmd:


VinitAmd:
retest this please

maxzhen:
Overall, this implementation seems simpler than the timer one. But I'm not sure if the code is correct. The delayed work once triggered will keep triggering itself. Is this behavior OK if the delayed work is triggered by an interrupt? If every interrupt triggers a delayed work, soon, we'll end up many delayed works triggering themselves in the system?

vengutta18:
Actually, with the mod_delayed_work you delay the pending work item by the latest delay.
So essentially, there will be just one work item in the queue at any given point in time with the latest delay.


This is how I'm thinking about the flow->
If an interrupt is triggered, it will execute the work immediately to process the FW log.
During that, the work item with a delay of the polling interval gets added.
1. If you don't get an interrupt by that time, the work will be executed.
2. If you get an interrupt before the polling interval, then the work queued by the interrupt gets executed, and the previously queued work gets cancelled.
Essentially, we are doing the same work.

I want to understand the motive behind the interrupt+polling mechanism here. Should we only poll to read after the interrupt to read the remaining data or should we keep polling and reading the fw log buf irrespective of the interrupts.

Maybe if we want to read always, irrespective of the interrupts, then this might not work.

maxzhen:
> Actually, with the mod_delayed_work you delay the pending work item by the latest delay. So essentially, there will be just one work item in the queue at any given point in time with the latest delay.
> 
> This is how I'm thinking about the flow-> If an interrupt is triggered, it will execute the work immediately to process the FW log. During that, the work item with a delay of the polling interval gets added.
> 
> 1. If you don't get an interrupt by that time, the work will be executed.
> 2. If you get an interrupt before the polling interval, then the work queued by the interrupt gets executed, and the previously queued work gets cancelled.
>    Essentially, we are doing the same work.
> 
> I want to understand the motive behind the interrupt+polling mechanism here. Should we only poll to read after the interrupt to read the remaining data or should we keep polling and reading the fw log buf irrespective of the interrupts.
> 
> Maybe if we want to read always, irrespective of the interrupts, then this might not work.

The goal is "to read always, irrespective of the interrupts". FW only sends interrupt when the ring buffer is half full. If we only rely on the interrupt, driver may wait for too long to see the latest change in the log. Or the interrupt may never come if FW added log entry but not cross the half ring buffer boundary. Driver has to periodically check the ring buffer for new log entries.

VinitAmd:
PR #609 merged, closing the delayed_work impl.

PR #621, opened 2025-06-26
Create bin dir for npu3 vf xrt-smi

hlaccabu:


hlaccabu:
@NishadSaraf This dir should just be a symlink to 17f1_10, I don't think an addition to WHENCE is needed but let me know

hlaccabu:
Retest this please

PR #625, opened 2025-06-29
Added name() function to the amdxdna drv class

ManojTakasi:
Added name() function to the amdxdna drv class
Tested by building and running xrt-smi examine in hw_emu.


NishadSaraf:
retest this please

PR #635, opened 2025-07-08
shim: device: add pci device handle to device class

wendyliang25:
Add pci device handle which is a shared pointer to xrt_core::pci::dev, which is the underline PCIe device to make sure the pci device will not get released before the shim_xdna device is released, otherwise, the application can get crash as shim_xdna will access xrt_core::pci::dev when it is closing the device.

At the moment, there is an issue if I keep create a `xrt::device` and keep it in global space, when application is complete, the cleanup sequence can release the `xrt_core::pci::dev` before releasing `xrt::device` which needs to access `xrt_core::pci::dev` during closing. This will result in segfault or invalid access or double free errors. here is the example:
```
#include <iostream>
std::shared_ptr<xrt::device> device;
int main(void)
{
  unsigned int device_index = 0;
  device = std::make_shared<xrt::device>(device_index);
  if (!device) {
    std::cout << "ERROR: faield to create device." << std::endl;
    return -1;
  }
  return 0;
}
```
The same issue happens if we put the `xrt::bo` to global space as releasing `xrt::bo` also needs to access `xrt_core::pci:dev`.

### DIFF
--- a/src/shim/device.cpp
+++ b/src/shim/device.cpp
@@ -1444,6 +1444,7 @@ device::
 device(const pdev& pdev, handle_type shim_handle, id_type device_id)
   : noshim<xrt_core::device_pcie>{shim_handle, device_id, !pdev.m_is_mgmt}
   , m_pdev(pdev)
+  , m_pcidev_handle(xrt_core::pci::get_dev(device_id,is_userpf()))
 {
   m_pdev.open();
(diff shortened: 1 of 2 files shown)


EugeneHiew:

maxzhen:
ok to test

maxzhen (approved):

PR #638, opened 2025-07-09
Add runtime cfg support to event tracing.

VinitAmd:


src/driver/amdxdna/aie2_debugfs.c
```diff
@@ -469,7 +470,78 @@ static int aie2_log_runtime_cfg_show(struct seq_file *m, void *unused)
...
+	int ret;
+
+	if (len > EVENT_CATEGORY_BUF_SIZE) {
+		XDNA_ERR(ndev->xdna, "Input length %zu > buffer size %d",
+			 len, EVENT_CATEGORY_BUF_SIZE);
+		return -EINVAL;
+	}
+
+	kbuf = event_category_buf;
+	if (copy_from_user(kbuf, buf, len))
```
NishadSaraf, line 491: Before accessing pointers from the userspace directly, they must be validated. Use `access_ok()`

NishadSaraf (changes requested):
Thanks for the PR.

src/driver/amdxdna/aie2_debugfs.c
vengutta18: copy_from_user does it for you.

VinitAmd: **What copy_from_user() does:**
_1.It validates the user-space pointer (buf) to ensure it is accessible and safe to read from.
2. It attempts to copy len bytes from the user-space address buf to the kernel-space buffer kbuf.
3. If the user-space memory is not accessible (e.g., invalid pointer, not readable), it will not crash the kernel. Instead, it returns the number of bytes that could not be copied._

PR #648, opened 2025-07-16
moving to latest xrt for ve2 fixes

chvamshi-xilinx:
Moving to latest XRT for ve2 fixes. Need below change for ve2
https://github.com/Xilinx/XRT/pull/9066


chvamshi-xilinx:
ok to test

PR #651, opened 2025-07-18
Export key app health info back to XRT

NishadSaraf:
Export key app health info back to XRT.


NishadSaraf:
Marking this PR as `do-not-merge` until the latest XRT package is available

maxzhen (approved):

PR #661, opened 2025-07-22
Cuidx - cu_index in execbuf is not used anymore

xuhz:


EugeneHiew:

maxzhen (approved):

maxzhen:
ok to test

xdavidz (approved):
does this change work with older cert? I guess yes, because the cu_index should be 0 for most of the case.

This will work with both elf flow and xclbin flow?
This will also work with chained exec_buf cmd in the future?

PR #671, opened 2025-07-24
increase size of app health report

zmotiXilinx:
If the app health report structure is not big enough, firmware will reject the request.


EugeneHiew:

PR #673, opened 2025-07-25
print logging without using pr_debug

zmotiXilinx:


EugeneHiew:

maxzhen:
ok to test

src/driver/amdxdna/aie2_logging.c
```diff
@@ -182,7 +186,7 @@ static void aie2_print_log_buffer_data(struct amdxdna_dev_hdl *ndev)
 		msg_size = (header->argc) * sizeof(u32);
 		if (msg_size > 0 && (char *)(msg + msg_size) <= end) {
 			*(char *)((char *)msg + msg_size - 1) = 0;
-			pr_debug("[NPU FW]%s", msg);
+			XDNA_INFO(ndev->xdna, "[NPU FW %s]: %s", s_log_level_str[header->level], msg);
```
VinitAmd, line 189: 1. if in abnormal/mem-override case "header->level > 4" lead to out-of-bound exp

2. IMO, Injecting FW logs in kmd dmesg buffer in INFO format is not advised, since these logs are too heavy and they may not be needed always, enabling these logs may hide most of the other imp driver msg. So, I changed the earlier added XDNA_INFO format to pr_debug. (e.g. I ran one of the vairt+xrt_test_flow app for 6 mins with logging enable got 4 GB FW info level logs imagine adding them in dmesg info).

3. Adding log type tag is good, in coming xrt interface changes if the log buffer would be picked up to user space via xrt and parsed/displayed there then no use of adding here, I think this is temporary till the xrt interface is not implemented?

PR #678, opened 2025-07-30
ve2: updating buffer allocation logic to support cma allocation from multiple regions

bisingha-xilinx:
EDF/Vitis Common CED contains multiple DDR (or LPDDR) memory banks, this pr is to support DMA access to any of the memory banks. This pr has the initial changes to support this feature. 
Updated the logic to allocate buffers with dma_alloc_coherent() instead of directly using drm_gem_dma_create(). This update is required to support https://jira.xilinx.com/browse/VITIS-15145

### DIFF
--- a/src/driver/amdxdna/amdxdna_gem_of.c
+++ b/src/driver/amdxdna/amdxdna_gem_of.c
@@ -98,7 +98,7 @@ static struct drm_gem_dma_object *amdxdna_cma_create(struct drm_device *dev, siz
 
 	gem_obj->funcs = &amdxdna_gem_dma_funcs;
 
-	//manually init the drm gem obj
+	/* manually init the drm gem obj */
 	ret = drm_gem_object_init(dev, gem_obj, size);
 	if (ret)
 		goto error;


src/driver/amdxdna/amdxdna_gem_of.c
```diff
@@ -77,29 +84,82 @@ struct drm_gem_object *amdxdna_gem_create_object_cb(struct drm_device *dev, size
...
+		size = round_up(size, PAGE_SIZE);
+
+	cma_obj = amdxdna_cma_create(dev, size);
+	if (IS_ERR(cma_obj))
+		return ERR_PTR(-ENOMEM);
 
-	dma = drm_gem_dma_create(dev, size);
-	if (IS_ERR(dma))
-		return ERR_CAST(dma);
+	vaddr = dma_alloc_coherent(dev->dev, size, &dma_addr, GFP_KERNEL);
```
houlz0507, line 140: gem_dma bo was used which alloc non-coherent or wc buffer. Why do you need coherent buffer here? Non-coherent is preferred if possible.

src/driver/amdxdna/amdxdna_gem_of.c
```diff
@@ -77,29 +84,82 @@ struct drm_gem_object *amdxdna_gem_create_object_cb(struct drm_device *dev, size
...
+
+	gem_obj = kzalloc(sizeof(struct amdxdna_gem_obj), GFP_KERNEL);
+	if (!gem_obj)
+		return ERR_PTR(-ENOMEM);
+
+	cma_obj = container_of(gem_obj, struct drm_gem_dma_object, base);
+
+	gem_obj->funcs = &amdxdna_gem_dma_funcs;
+
+	//manually init the drm gem obj
```
houlz0507: please use /* */

src/driver/amdxdna/amdxdna_gem_of.c
```diff
@@ -77,29 +84,82 @@ struct drm_gem_object *amdxdna_gem_create_object_cb(struct drm_device *dev, size
...
 							 struct drm_file *filp)
 {
-	struct drm_gem_dma_object *dma;
+	struct drm_gem_dma_object *cma_obj;
 	struct amdxdna_gem_obj *abo;
 	size_t size = args->size;
+	dma_addr_t dma_addr;
+	void *vaddr;
 
 	/* Round up to more than 4K to ensure to allocate memory from CMA always */
```
houlz0507, line 130: Can you briefly explain why the CMA allocation can be determined by size?

NishadSaraf:
CI failed due to checkpatch warnings.

src/driver/amdxdna/amdxdna_gem_of.c
bisingha-xilinx: Hi @houlz0507
We observed one issue, for smaller buffer sizes, Linux allocates randomly CMA memory from some other region not from the dedicated CMA memory region. And rounding up the smaller buffer sizes to page size, guarantees to get CMA memory from the dedicated CMA memory region.

It creates an issue for some tests. In some usecases kernels do not have access/connectivity to all memory regions but to only CMA memory region. If Linux allocates CMA memory from some other region, then the test fails.

src/driver/amdxdna/amdxdna_gem_of.c
bisingha-xilinx: Hi @houlz0507 thank you for the review. Sure, I'll update it.

src/driver/amdxdna/amdxdna_gem_of.c
houlz0507: MM... how do you guarantee >=8K memory will be from CMA region in all circumstance?

bisingha-xilinx: Hi,
This is the lowest size I tested continuously multiple times having 1000/2000 iteration in each test. I didnt observe the issue. If I just round up to 4K (for smaller buffer sizes) then I saw the mentioned issue.

So, we aggreed to round up to 2*page_size for smaller sizes and round up with page_size for bigger sizes.

houlz0507: What if the CMA region is used up? I think alloc_coherent will try to allocate memory from other places instead of returning an error.  I did not see a way to make sure the memory is from CMA regions by using alloc_coherent.

bisingha-xilinx: Hi @houlz0507 
if the CMA region is used up, alloc_coherent is not allocating from other places. it throws error. In my test I've observed this behavior. Let me confirm again.

houlz0507: Please try small buffer e.g. 8k for your experiment.

bisingha-xilinx: Hi @houlz0507 thank you for the review.
Sure.

chvamshi-xilinx:
Hi @bisingha-xilinx , Lets send a separate PR once you implement and verify memory region logic. Closing this PR.

bisingha-xilinx:
> Hi @bisingha-xilinx , Lets send a separate PR once you implement and verify memory region logic. Closing this PR.

Sure Vamshi

PR #683, opened 2025-08-04
Updating XRT submodule to its latest commit

bisingha-xilinx:
Updated XRT submodule to its latest commit to get the fix for https://jira.xilinx.com/browse/CR-1245557.


bisingha-xilinx:
Closing it as the commit is not yet available in the TA

PR #696, opened 2025-08-14
Tcp server fix

xuhz:
we don't have a better solution in npu to set debugger q
1. they have difficulty configuring active hw ctx
2. need more thinking / discussion to disable hws during debug

before we have a better solution, we change to 
1. attach the debugger q before ctx runs
2. cert checks the q only when it detects a control code hang

this is still a good/useful feature to debug a control code hang though. better than nothing.


AMDGithubSCIMAdmin:

maxzhen:
ok to test

maxzhen (dismissed):

maxzhen:
retest this please

PR #700, opened 2025-08-19
Implement per context get app health IOCTL

NishadSaraf:
Implement per context get app health IOCTL.

### DIFF
--- a/src/driver/amdxdna/aie2_pci.c
+++ b/src/driver/amdxdna/aie2_pci.c
@@ -1217,42 +1217,16 @@ static int aie2_get_info(struct amdxdna_client *client, struct amdxdna_drm_get_i
 }
 
 static int aie2_query_ctx_status_array(struct amdxdna_client *client,
-				       struct amdxdna_drm_get_info_array *args)
+				       struct amdxdna_drm_hwctx_entry *tmp, pid_t pid, u32 ctx_id)
 {
-	struct amdxdna_drm_query_hwctx_array __user *buf;
-	struct amdxdna_drm_query_hwctx_array *tmp;
 	struct amdxdna_dev *xdna = client->xdna;
-	int idx, ctx_limit, ctx_cnt, min, i;
 	struct amdxdna_client *tmp_client;
 	struct aie2_mgmt_dma_hdl mgmt_hdl;
 	struct app_health_report *r;
 	struct amdxdna_ctx *ctx;
-	unsigned long ctx_id;
+	unsigned long id;
 	u32 hw_i = 0;
-	u32 buf_size;
-	int ret = 0;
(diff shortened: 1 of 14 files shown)


src/driver/amdxdna/aie2_pci.c
```diff
@@ -1217,42 +1217,17 @@ static int aie2_get_info(struct amdxdna_client *client, struct amdxdna_drm_get_i
 }
 
 static int aie2_query_ctx_status_array(struct amdxdna_client *client,
-				       struct amdxdna_drm_get_info_array *args)
+				       struct amdxdna_drm_hwctx_entry *tmp, long long pid,
```
maxzhen: should we use pid_t?

src/driver/amdxdna/aie2_pci.c
```diff
@@ -1327,44 +1308,95 @@ static int aie2_query_ctx_status_array(struct amdxdna_client *client,
...
 	ret = amdxdna_pm_resume_get(xdna);
 	if (ret)
 		return ret;
 
 	mutex_lock(&xdna->dev_lock);
-	mutex_lock(&xdna->dev_handle->aie2_lock);
+
+	ctx_limit = aie2_rq_context_limit(&xdna->dev_handle->ctx_rq);
+	WARN_ON(ctx_limit > AMDXDNA_MAX_NUM_ELEMENT);
+	ctx_cnt = aie2_rq_active_context(&xdna->dev_handle->ctx_rq);
```
maxzhen: Move to aie2_query_ctx_status_array?

src/driver/amdxdna/aie2_pci.c
```diff
@@ -1327,44 +1308,95 @@ static int aie2_query_ctx_status_array(struct amdxdna_client *client,
...
+				 ctx_cnt, args->num_element);
+			args->num_element = ctx_cnt;
+			return -ENOSPC;
+		}
+
+		if (aie2_query_ctx_status_array(client, tmp, 0, 0))
+			goto exit;
+
+		break;
+	case DRM_AMDXDNA_HW_CONTEXT_HEALTH:
```
maxzhen: You're not just getting ctx health, so why not call it DRM_AMDXDNA_HW_CONTEXT?

maxzhen (changes requested):

src/driver/amdxdna/aie2_pci.c
NishadSaraf: Moved the above lines to the `DRM_AMDXDNA_HW_CONTEXT_ARRAY` case. As args->num_element needs to be greater than or equal to ctx_cnt. If not, should fail early.

PR #708, opened 2025-08-23
remove workaround for supporting unlimited instruction BO

maxzhen:


maxzhen:
retest this please - disk full

PR #716, opened 2025-08-28
Firmware_log json addition and df_bw test-case update

aktondak:
This PR adds the firmware_log json configuration and makes query changes to xdna driver build.
This PR also updates the elf used by df-bw test-case as per the following CR : https://jira.xilinx.com/browse/AIESW-5911
Also commenting out the firmware debug buffer initialization until driver ioctls are ready. (For coverity fix)

### DIFF
--- a/tools/bins/configs/firmware_log.json
+++ b/tools/bins/configs/firmware_log.json
@@ -23,6 +23,12 @@
     "ipu_log_message_header": {
       "fields": [
         {
+          "name": "timestamp",
+          "type": "uint64_t",
+          "width": 64,
+          "format": "0x{:016X}"
+        },
+        {
           "name": "format",
           "enumeration": "ipu_log_format",
           "type": "uint32_t",


maxzhen (dismissed):

tools/bins/configs/firmware_log.json
```diff
@@ -0,0 +1,76 @@
...
+        "wrn": 2,
+        "inf": 3,
+        "dbg": 4
+      }
+    }
+  },
+  "structures": {
+    "ipu_log_message_header": {
+      "fields": [
+        {
```
NishadSaraf, line 25: Hi @aktondak, this doesn't seem latest to me. Can you please double check?

aktondak: Thanks for pointing out. Yes there's a timestamp field addition. I've made the required changes.

maxzhen (approved):

PR #719, opened 2025-09-01
[EDGEML-11139] When QOS priority is low, dpm level should be low.

VinitAmd:


maxzhen:
retest this please

maxzhen:
> Are we sure this is the correct behavior and same as Windows? When priority is low, only priority matters, no need to look at other QoS attributes?

Confirmed that this is the same behavior on Windows.

gyang1099 (approved):

PR #720, opened 2025-09-01
Add shim changes to dump log buffer

rbramand-xilinx:
- Added required shim changes for log buffer dumping
- Updated XRT submodule with required commit id that has required userspace changes for log buffer


chvamshi-xilinx:
Adding Max as this PR contains changes in xdna shim aswell.

PR #721, opened 2025-09-02
Update xrt-smi validate binaries using VTD

rchane:
This PR adds VTD as a submodule to amdxdna and updates xrt-smi validate binaries in /tools/bins/ at driver build-time with their counterparts in VTD. Due to some file renames and possible future changes to structure of VTD, removal of files, and addition of files, this first iteration explicitly copies the necessary files one-to-one. This mirrors what was done on MCDM side in order to use VTD files. Also cleaned up some unnecessary files in /tools/bins.

Testing was done on STX machine to verify comparable numbers for xrt-smi validate tests.


rchane:
tools/bins will be removed and I will make the necessary changes to CMake/shim code, thank you.

rchane:
Closing PR, as relevant changes will be overwritten by migration to VTD archive approach that is currently in development.

PR #730, opened 2025-09-08
shim changes to dump log buffer

AdvaitNaik:
- required shim changes for log buffer dumping
- reference - https://github.com/amd/xdna-driver/pull/720

### DIFF
--- a/src/shim/buffer.cpp
+++ b/src/shim/buffer.cpp
@@ -790,6 +790,8 @@ dbg_buffer::
   if (m_ctx_id == AMDXDNA_INVALID_CTX_HANDLE)
     return;
 
+  // TODO FIXME 
+#if 0
   try {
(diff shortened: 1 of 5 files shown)


src/shim/buffer.cpp
```diff
@@ -790,13 +790,16 @@ dbg_buffer::
   if (m_ctx_id == AMDXDNA_INVALID_CTX_HANDLE)
     return;
 
+  // TODO FIXME 
+#if 0
```
maxzhen, line 794: Please fix :-).

AdvaitNaik: discussed with @rbramand-xilinx and created ticket - https://jira.xilinx.com/browse/CR-1249680
rahul will make required changes in xrt

PR #732, opened 2025-09-10
Add xrt_test option for vf concurrency test

hlaccabu:


xdavidz (changes requested):
see https://github.com/amd/xdna-driver/blob/main/test/shim_test/multi_threads.h#L39
which can catch test failures and automatically return 0 for all pass and non-zero for 1 or more test failed.

I think the current thread join doesn't check any test faliures?

xdavidz (approved):

PR #733, opened 2025-09-10
Unify DPT (Debug/Profile/Trace) firmware debug across generations

NishadSaraf:
- Introduce a single DPT-based infrastructure for firmware debug/profile/trace across current and future devices. Remove legacy event-trace/DRAM logging to cut redundancy.
- Improve ring handling by extending 32-bit FW pointers to 64-bit in-driver, making wrap and tail tracking robust and transparent
- Enable firmware logging by default at ERROR level. Simplify usage via a boolean debugfs node (dump_fw_log) to toggle printing to dmesg.
- Provide module params for advanced control (fw_log_level, fw_log_size, poll_fw_log).
- Establish common management-DMA helpers for buffer allocation/handling. More consolidation to follow.


maxzhen (changes requested):
Please queue the same amdxdna_dpt_worker in both timer and irq handler. There is no reason to do different things in them.

xdavidz (approved):
nice to have those as ops callback. @hlaccabu can we test this after this merged? We need this working on aie4 before next weekend.

maxzhen:
> nice to have those as ops callback. @hlaccabu can we test this after this merged? We need this working on aie4 before next weekend.

There is an upcoming PR for aie4 part of callback change soon. We can start debugging on aie4 once both PRs are merged.

VinitAmd:
is event trace feature removed from FW debugging ? @NishadSaraf @maxzhen

NishadSaraf:
> is event trace feature removed from FW debugging ? @NishadSaraf @maxzhen

Hi @VinitAmd, I'll have another PR to enable it with this new infra.

src/driver/amdxdna/amdxdna_mgmt.c
```diff
@@ -0,0 +1,114 @@
...
+int amdxdna_mgmt_buff_clflush(struct amdxdna_mgmt_dma_hdl *dma_hdl, u32 offset, size_t size)
+{
+	if (!dma_hdl)
+		return -EINVAL;
+
+	if (offset + size > dma_hdl->size)
+		return -EINVAL;
+
+	/*
+	 * After flushing the buffer and handing it over to the device,
```
VinitAmd, line 71: Typo "handing"

src/driver/amdxdna/amdxdna_dpt.h
```diff
@@ -0,0 +1,60 @@
...
+
+struct amdxdna_dpt {
+	bool				enabled;
+	u8				minor;
+	u8				major;
+	u32				payload_version;
+	char				name[20];
+	struct amdxdna_dev		*xdna;
+	struct amdxdna_mgmt_dma_hdl	*dma_hdl;
+	struct wait_queue_head		wait;
```
VinitAmd, line 38: Would be better to follow linux code standard of Reverse X-mas Tree format.

src/driver/amdxdna/amdxdna_mgmt.c
```diff
@@ -0,0 +1,114 @@
...
+	 * the user must wait for the device to complete its operations and return
+	 * control before attempting to write to the buffer again.
+	 */
+	drm_clflush_virt_range(dma_hdl->vaddr + offset, size ? size : dma_hdl->size);
```
VinitAmd, line 75: offset should be renamed to footer/metadata/metaInfo for better understanding of dma buffer partition.

src/driver/amdxdna/amdxdna_dpt.c
```diff
@@ -0,0 +1,431 @@
...
+
+	/* Retain the state of dump_to_dmesg across suspend/resume */
+	fw_log_dump_to_dmesg = xdna->fw_log->dump_to_dmesg;
+
+	ret = xdna->dev_info->ops->fw_log_fini(xdna);
+	if (ret)
+		XDNA_ERR(xdna, "Failed to disable fw logging: %d", ret);
+
+	amdxdna_dpt_irq_fini(log_hdl);
+	amdxdna_dpt_enable_polling(log_hdl, false);
```
VinitAmd, line 424: is this call required ?, already being set inside below call _amdxdna_dpt_dump_to_dmesg_

src/driver/amdxdna/amdxdna_dpt.c
```diff
@@ -0,0 +1,431 @@
...
+
+	if (dump) {
+		dpt->local_buffer = kzalloc(fw_log_size, GFP_KERNEL);
+		if (!dpt->local_buffer) {
+			XDNA_ERR(dpt->xdna, "Failed to allocate fw fetch buffer");
+			return -ENOMEM;
+		}
+		amdxdna_dpt_enable_polling(dpt, true);
+	} else {
+		if (!poll_fw_log)
```
VinitAmd, line 325: Here !poll_fw_log condition is used to disable polling but line no: 323 doesn't use this condition, this is not symmetrical use case.

src/driver/amdxdna/amdxdna_dpt.c
```diff
@@ -0,0 +1,431 @@
...
+		amdxdna_mgmt_buff_clflush(dma_hdl, 0, end);
+		memcpy(buffer + (log_size - start),
+		       amdxdna_mgmt_buff_get_cpu_addr(dma_hdl, 0), end);
+	} else {
+		amdxdna_mgmt_buff_clflush(dma_hdl, start, end - start);
+		memcpy(buffer, amdxdna_mgmt_buff_get_cpu_addr(dma_hdl, start), end - start);
+	}
+
+	*size = req_size;
+	dpt->head = tail;
```
VinitAmd, line 167: should be at line no: 139

src/driver/amdxdna/amdxdna_dpt.c
```diff
@@ -0,0 +1,431 @@
...
+		if (ret != -EOPNOTSUPP)
+			XDNA_ERR(xdna, "Failed to configure fw logging: %d", ret);
+		else
+			ret = 0;
+		goto mfree;
+	}
+
+	ret = amdxdna_dpt_irq_init(log_hdl);
+	if (ret)
+		XDNA_ERR(xdna, "Failed to init fw logging IRQ: %d", ret);
```
VinitAmd, line 387: if irq_int failed, simply init fw logging should fail ? is it safe to continue FW logging?

src/driver/amdxdna/amdxdna_mgmt.c
NishadSaraf: I don't see the typo. Can you double check?

src/driver/amdxdna/amdxdna_dpt.h
NishadSaraf: There is no such rule to follow for the format of structure definitions. I prefer to have them defined by what makes more logical sense and group similar ones together.

src/driver/amdxdna/amdxdna_mgmt.c
NishadSaraf: These are generic helper APIs for managing management buffers. Renaming the offset according to your suggestion would be incorrect.

src/driver/amdxdna/amdxdna_dpt.c
NishadSaraf: Here we disable polling, if it was enabled by setting the modparam. `amdxdna_dpt_dump_to_dmesg()` disables polling if it was enabled to facilitate dumping to dmesg.

src/driver/amdxdna/amdxdna_dpt.c
NishadSaraf: We should unconditionally enable polling when streaming the FW log to dmesg, because IRQs are only generated when the buffer is half or completely full and may miss entries. When stopping, disable polling only if we enabled it for this purpose.

src/driver/amdxdna/amdxdna_dpt.c
NishadSaraf: HEAD must be updated only after a local copy of the fw log buffer has been taken.

src/driver/amdxdna/amdxdna_dpt.c
NishadSaraf: Some platforms do not yet have IRQ support, but we still would like to support FW logging.

src/driver/amdxdna/amdxdna_dpt.c
VinitAmd: check return at  line no 150, if control returns from here how to update tail to avoid overflow in next read ?

src/driver/amdxdna/amdxdna_dpt.h
VinitAmd: You can find here https://docs.kernel.org/process/maintainer-kvm-x86.html 5.4.2. Coding Style, it's preferred not mandatory, since code would be upstream later.

src/driver/amdxdna/amdxdna_mgmt.c
VinitAmd: would be incorrect when change the api declaration, but makes more sense when offset is renamed according to memory info from calling location.

NishadSaraf: The meaning of offset is context dependent. It can refer to an offset within the app health report buffer, the telemetry buffer, the firmware log buffer, etc.

src/driver/amdxdna/amdxdna_dpt.c
NishadSaraf: TAIL is updated by `amdxdna_update_tail()` (on line 268) irrespective of the HEAD value. HEAD updates only when the data is read from the FW log buffer by the driver. If the reader is slow, there is no way to avoid the FW from overflowing the data. As for the case where the required buffer size is more than the available size (line 150), in today's implementation (even in the worst case), such a scenario is not possible, as the driver buffer size is the same FW log buffer (ref line 318). However, once xrt-smi support is added, I'll be sure to add support to handle the overflow case.

src/driver/amdxdna/amdxdna_dpt.h
NishadSaraf: The document mentions variable declaration and not definition.

VinitAmd (approved):

PR #735, opened 2025-09-10
clean up async error test case

maxzhen:
@wendyliang25, when you have the code change to fetch async error from driver, please populate the verify_result() portion of the change here.


wendyliang25:
> @wendyliang25, when you have the code change to fetch async error from driver, please populate the verify_result() portion of the change here.

Thanks @maxzhen , sure, i will

PR #737, opened 2025-09-11
XRT, VTD submodule update and shim changes

aktondak:
This PR updates the XRT  submodule to latest. There are following changes in this PR 

1. Adds shim code to query archive path from installed driver path (XRT : https://github.com/Xilinx/XRT/pull/9213)
2. Edits shim code to consume watch mode flag passed down from xrt-smi for event tracing and firmware logging. (XRT : https://github.com/Xilinx/XRT/pull/9218)
3. Updates context health report to get ctx id and PID from driver. (XRT : https://github.com/Xilinx/XRT/pull/9221)


CMake/pkg.cmake
```diff
@@ -73,6 +73,12 @@ install(DIRECTORY ${AMDXDNA_BINS_DIR}/download_raw/xbutil_validate/bins/
   PATTERN "*.yaml"
   )
 
+# Install VTD runner archive file
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/VTD/runner/xrt_smi_strx.a
```
maxzhen: this is only for stx, not phx? how about phx?

aktondak: I wanted to do it in phases and add phoenix later. But now I think of it, this PR might fail pipeline for xrt-smi validate failure on phoenix. 
Let me update this PR to add phoenix archive as well once this PR fails pipeline with expected behavior.

maxzhen: > I wanted to do it in phases and add phoenix later. But now I think of it, this PR might fail pipeline for xrt-smi validate failure on phoenix. Let me update this PR to add phoenix archive as well once this PR fails pipeline with expected behavior.

Yeah, it may fail. Good if you can also have phx .a file

PR #748, opened 2025-09-17
create cma bo with carvedout fashion for ve2

bisingha-xilinx:
This pr removes the temporary amdxdna_gem_of.c and amdxdna_gem_of.h. This is to create CMA BOs in carvedout fashion for VE2 usecases and uses the existing framework provided by the amdxdna_gem.c/.h

1. It uses existing import, creation and obj free framework
2. It has its own dma_buf_ops in a newly added file amdxdna_cma.c. It they support map, unmap, vmap, vunmap, release and mmap as well. All are functional.
4. Currently it creates only coherent memory later we'll support for non-coherent also for cacheable usecases for VE2
5. Tested all the hw unit tests in Telluride. All are passing


src/driver/amdxdna/amdxdna_cma.c
```diff
@@ -0,0 +1,169 @@
...
+{
+    struct amdxdna_cmabuf_priv *cmabuf = dbuf->priv;
+    size_t size = vma->vm_end - vma->vm_start;
+    int ret;
+
+    if (size > cmabuf->size)
+        return -EINVAL;
+
+    /* clear the vm_pgoff to avoid dma_buf_ops.mmap failure */
+    vma->vm_pgoff = 0;
```
houlz0507: Please save vm_pgoff before setting it to zero and restore it after dma_mmap_coherent

src/driver/amdxdna/amdxdna_cma.c
```diff
@@ -0,0 +1,169 @@
...
+
+    ret = dma_mmap_coherent(cmabuf->dev->dev, vma,
+                             cmabuf->cpu_addr,
+                             cmabuf->dma_addr,
+                             cmabuf->size);
+    if (ret)
+        return ret;
+
+    vma->vm_private_data = cmabuf;
+    vma->vm_ops = NULL;
```
houlz0507: Why vm_private_data is set when vm_ops = NULL? Probably just remove above two lines?

src/driver/amdxdna/amdxdna_cma.c
```diff
@@ -0,0 +1,169 @@
...
+
+	iosys_map_set_vaddr(map, cmabuf->cpu_addr);
+
+	return 0;
+}
+
+static void amdxdna_cmabuf_vunmap(struct dma_buf *dbuf, struct iosys_map *map)
+{
+    /* This is a no-op. No operation needed for CMA memory */
+}
```
houlz0507: Remove this empty function and line 121?

bisingha-xilinx:
Hi @houlz0507,
Thank you for the review. I've updated the PR accordingly.

src/driver/amdxdna/amdxdna_cma.c
```diff
@@ -0,0 +1,168 @@
...
+	vm_flags_set(vma, VM_IO | VM_DONTEXPAND | VM_DONTDUMP);
+
+	ret = dma_mmap_coherent(cmabuf->dev->dev, vma,
+			cmabuf->cpu_addr,
+			cmabuf->dma_addr,
+			cmabuf->size);
+	if (ret)
+		return ret;
+
+	vma->vm_pgoff = vm_pgoff;
```
houlz0507, line 99: Maybe better to replace line 98-103 as below?
vma->vm_pgoff = vm_pgoff;
return ret;

src/driver/amdxdna/amdxdna_gem.h
```diff
@@ -33,6 +33,7 @@ struct amdxdna_mem {
 	u32				nr_pages;
 	struct list_head		umap_list;
 	bool				map_invalid;
+	bool				cma;
```
houlz0507: Why do you need this flag? I think you are using global xdna->cma? This seems redundant.

bisingha-xilinx: Hi Lizhi, removed it. Marking this conversation resolved.

src/driver/amdxdna/amdxdna_cma.c
bisingha-xilinx: Hi, Lizhi, yes, we can do that. Now Im restoring the vm_pgoff first. Marking this conversation as resolved.

houlz0507 (dismissed):

bisingha-xilinx:
> Should we also clean up the driver interface and remove AMDXDNA_BO_DMA type from amdxdna_accel.h?

Hi @maxzhen, yes. since we will be no longer using amdxdna_gem_of.c/.h. I've removed it. Thank you for reviewing.

I've updated the PR, now the build is also passing. Could you please review.

Thank you.

bisingha-xilinx:
Hi @houlz0507 I've updated the pr by adding the amdxdna_cma.o to Kbuild and removed the AMDXDNA_BO_DMA type from amdxdna_accell.h. Other than that there are no functional changes. Now build is also passing. Could you please review again. 
Thank you.

maxzhen (approved):

PR #750, opened 2025-09-17
Disable TXN app health test

NishadSaraf:
Disable TXN app health test.


maxzhen:
retest this please - birman 15 not available

wendyliang25 (approved):
how about add some comments to describe the reason to disable this test is due to this test sometimes failing, it the root cause is under investigation? thanks

PR #768, opened 2025-09-24
[XRT-SMI]Removing redundant files

aktondak:
This PR cleans up all the files that are redundant after archive migration of xrt-smi tests. For any new file to be added for xrt-smi. Kindly add it to the xrt-smi archive present in [VTD repository](https://github.com/Xilinx/VTD) and follow the archive route


maxzhen (commented):
Wondering why these mobilenet binaries are not removed?

maxzhen (approved):

PR #772, opened 2025-09-25
test: shim: add read async error multi times in multi threads

wendyliang25:
Add test case to read async errors multiple times in multiple threads.

To test async error read when there is no errors, we will need to run the test after xdna module is probed and before any runs launched on the hardware.


AMDGithubSCIMAdmin:

maxzhen:
ok to test

maxzhen (approved):

PR #773, opened 2025-09-25
test: shim: add instruction code invalid address access test

wendyliang25:
Add test case to have instruction code access invalid address we should expect timeout and then, if we start a good run, the good run should finish properly.

### DIFF
--- a/test/shim_test/io_test.cpp
+++ b/test/shim_test/io_test.cpp
@@ -538,3 +538,14 @@ void TEST_async_error_multi(device::id_type id, std::shared_ptr<device>& sdev, a
   threads.run_test(id, sdev, arg);
 }
 
+void
+TEST_instr_invalid_addr_io(device::id_type id, std::shared_ptr<device>& sdev, arg_type& arg)
+{
+  elf_io_timeout_test_bo_set invalid_addr_txn_set{sdev.get(), "timeout.xclbin",
+                                                  "instr_invalid_addr.elf", 0xFFFFFFFF};
+  // verification is inside run()
+  invalid_addr_txn_set.run();
+
+  std::vector<uint64_t> params = {IO_TEST_NORMAL_RUN, 1};
+  elf_io(id, sdev, params, "design.xclbin");
+}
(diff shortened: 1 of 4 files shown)


AMDGithubSCIMAdmin:

maxzhen (changes requested):
Please just expand elf_io_timeout_test_bo_set class to reduce duplicated code. No need to have a dedicated test case class for invalid addr test.

test/shim_test/io_test.cpp
```diff
@@ -39,10 +39,12 @@ alloc_and_init_bo_set(device* dev, const char *xclbin)
     base = std::make_unique<io_test_bo_set>(dev);
     break;
   case KERNEL_TYPE_TXN:
+    std::cout << __func__ << ", KERNEL_TYPE_TXN\n";
     base = std::make_unique<elf_io_test_bo_set>(dev, std::string(xclbin));
     break;
   case KERNEL_TYPE_TXN_PREEMPT:
   case KERNEL_TYPE_TXN_FULL_ELF_PREEMPT:
+    std::cout << __func__ << ", KERNEL_TYPE_TXN, preempt\n";
```
maxzhen: remove?

test/shim_test/io_test.cpp
```diff
@@ -39,10 +39,12 @@ alloc_and_init_bo_set(device* dev, const char *xclbin)
     base = std::make_unique<io_test_bo_set>(dev);
     break;
   case KERNEL_TYPE_TXN:
+    std::cout << __func__ << ", KERNEL_TYPE_TXN\n";
```
maxzhen: remove?

maxzhen (changes requested):

wendyliang25:
> Please just expand elf_io_timeout_test_bo_set class to reduce duplicated code. No need to have a dedicated test case class for invalid addr test.

Hi @maxzhen , i have changed the timeout test set to enable to pass more arguments to the constructor so that both test cases can use the same timeout class. could you check again? thanks

maxzhen:
ok to test

maxzhen (approved):

PR #775, opened 2025-09-26
Fix uninitialized app health report pointer

NishadSaraf:
Fix uninitialized app health report pointer.


NishadSaraf:
Locally verified the PR. Active context shows accurate TXN OP ID and context PC, while the idle one shows reset values. 
app_health1

PR #777, opened 2025-09-28
update testcase timeout value and check

AdvaitNaik:
- provided cmd line option to change timeout value

```
root@qemuvm:~# ./advanaik/bins/bin/xrt_test -h

Usage: xrt_test.elf [options] [test case ID separated by space]
Options:
        ...
        -w: timeout in seconds (default 600 sec, some simnow server are slow)
        ...

        Example Usage: ./xrt_test <# for stress test> -s 20 -d vadd -x vadd
                       Run stress test with Vadd kernel and xclbin for 20 rounds
```
- compare and dump ofm to file after timeout

### DIFF
--- a/test/xrt_test/xrt_test.cpp
+++ b/test/xrt_test/xrt_test.cpp
@@ -241,6 +241,32 @@ void check_umq_resnet50_result(int *ofm, const std::string& filename)
 
 template <typename TEST_BO>
 void
+dump_ofm_to_file(TEST_BO& ofm_bo)
+{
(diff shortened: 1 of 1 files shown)


xdavidz (approved):

xdavidz (changes requested):
can you also add all layer resnet50 timeout handling too?
can you dump the ofm into a file when times out?

AdvaitNaik:
> can you also add all layer resnet50 timeout handling too? can you dump the ofm into a file when times out?

done

PR #783, opened 2025-09-29
Update get async error ioctl to return only 0 for success case

wendyliang25:
This patch has two changes:
 * returns 0 only in when get async error success
 * restructure aie2 get array ioctl implementation to split get async
    error information implementation and get other hardware context
    information implementation.
* remove aie2_error get async error implementation as it is just
   a very thin wrapper which is not necessary.
* specify only one elment to get async error information in shim layer

### DIFF
--- a/src/driver/amdxdna/aie2_pci.c
+++ b/src/driver/amdxdna/aie2_pci.c
@@ -1329,19 +1329,45 @@ exit:
 	return ret;
 }
 
-static int aie2_get_array(struct amdxdna_client *client, struct amdxdna_drm_get_array *args)
+static int aie2_get_array_async_error(struct amdxdna_dev *xdna, struct amdxdna_drm_get_array *args)
+{
+	struct amdxdna_async_error tmp;
+	void __user *buf;
+	int ret;
+
+	buf = u64_to_user_ptr(args->buffer);
+	if (!access_ok(buf, sizeof(tmp))) {
+		XDNA_ERR(xdna, "Failed to access assync error buffer, 0x%lx",
+			 sizeof(tmp));
+		return -EFAULT;
+	}
+
+	ret = amdxdna_error_get_last_async(xdna, &xdna->dev_handle->async_errs_cache, 1, &tmp);
+	if (ret < 0)
+		goto exit;
+	/* only return last async error */
+	args->num_element = ret;
+	if (ret > 0) {
+		args->element_size = sizeof(tmp);
+		if (copy_to_user(buf, &tmp, sizeof(tmp))) {
+			ret = -EFAULT;
+			goto exit;
+		}
+	}
+
+	ret = 0;
+
+exit:
+	return ret;
+}
+
+static int aie2_get_array_hwctx(struct amdxdna_client *client, struct amdxdna_drm_get_array *args)
 {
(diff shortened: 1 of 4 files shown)


AMDGithubSCIMAdmin:

maxzhen:
ok to test

maxzhen:
Verified in a virtio environment.

maxzhen (dismissed):

wendyliang25:
ok to test

maxzhen (dismissed):

src/driver/amdxdna/aie2_pci.c
```diff
@@ -1329,20 +1329,48 @@ static int aie2_query_ctx_status_array(struct amdxdna_client *client,
...
+	int ret;
+
+	buf = u64_to_user_ptr(args->buffer);
+	if (!access_ok(buf, sizeof(tmp))) {
+		XDNA_ERR(xdna, "Failed to access assync error buffer, 0x%lx",
+			 sizeof(tmp));
+		return -EFAULT;
+	}
+
+	mutex_lock(&xdna->dev_lock);
```
houlz0507: why need to take global lock here?  There is a cache lock been taken in amdxdna_error_get_last_async

src/driver/amdxdna/aie2_pci.c
```diff
@@ -1329,20 +1329,48 @@ static int aie2_query_ctx_status_array(struct amdxdna_client *client,
...
+	ret = amdxdna_error_get_last_async(xdna, &xdna->dev_handle->async_errs_cache, 1, &tmp);
+	if (ret < 0)
+		goto exit;
+	// only return last async error;
```
houlz0507: Please use /**/

src/driver/amdxdna/aie2_pci.c
wendyliang25: Hi @houlz0507 , i had thought the dev_lock is to lock the device itself for all ioctls, that is to avoid the driver doing something such as to destroy or release device resources

wendyliang25: Hi @houlz0507 , updated to remove the dev_lock

src/driver/amdxdna/aie2_pci.c
wendyliang25: updated to use `/**/`

src/driver/amdxdna/aie2_pci.c
```diff
@@ -1329,20 +1329,46 @@ static int aie2_query_ctx_status_array(struct amdxdna_client *client,
...
+	args->num_element = ret;
+	if (ret > 0) {
+		args->element_size = sizeof(tmp);
+		if (copy_to_user(buf, &tmp, sizeof(tmp))) {
+			ret = -EFAULT;
+			goto exit;
+		}
+	}
+
+	ret = 0;
```
NishadSaraf, line 1358: This seems redundant, but okay. :)

wendyliang25: Hi @NishadSaraf , the return can be >0 to indicate how many cached async errors from the function even though it is always from `1` but this function is the return for the ioctl, since most caller considers ret != 0 as error, and thus set it to 0 for success.

PR #787, opened 2025-10-03
Fix SMU power off issue

amd-kirkirov:


AMDGithubSCIMAdmin:

src/driver/amdxdna/npu4_family.h
```diff
@@ -35,6 +36,7 @@ extern const struct amdxdna_dev_priv npu4_dev_priv;
 #define MPNPU_PUB_SCRATCH13            0x30100A0
 #define MPNPU_PUB_SCRATCH14            0x30100A4
 #define MPNPU_PUB_SCRATCH15            0x30100A8
+#define MPNPU_SUSP_RDY                 0x3010AC0
```
maxzhen: Is this the MPIPU_C2PMSG_112? Why do you call it MPNPU_SUSP_RDY?

amd-kirkirov: Yes it was MPIPU_C2PMSG_112. It is removed since not needed.

maxzhen (dismissed):

maxzhen:
ok to test

src/driver/amdxdna/aie2_psp.c
```diff
@@ -83,6 +83,21 @@ static int psp_exec(struct psp_device *psp, u32 *reg_vals)
 	return 0;
 }
 
+int aie2_psp_waitmode_poll(struct psp_device *psp)
+{
+	int mode_reg = -1, ret;
+
+	ret = readx_poll_timeout(readl, PSP_REG(psp, PSP_PWAITMODE_REG), mode_reg,
```
houlz0507, line 90: Is this register npu4 only? What about other hardware?

maxzhen: It should work the same way on PHX, just confirmed with FW team. @amd-kirkirov , please add the register to PHX platform as well.

amd-kirkirov: PHX register is added

PR #792, opened 2025-10-09
amdxdna: Update async error handling to be seperate from XRT

wendyliang25:
The amdxdna kernel driver last async error caching implementation should be independent to XRT. The error code encoding in the kernel driver is the same as XRT error code today, but it should be independent to future XRT changes. This patch did the following changes:
* rename XRT/xrt to AMDXDNA/amdxdna in macros, function names, and variable names
* reorganize the code to keep the implementation in aie2_error.c and aie2_pci.c, and remove amdxdna_error.c
* record the error tile location in extra error code field. Today, XRT doesn't expose the the extra error code information to user, however, the XRT error structure `xclErrorLast` passed to the SHIM layer does contain this extra XRT error code field, and thus, keeps the error location in the extra error code.


AMDGithubSCIMAdmin:

maxzhen:
ok to test

PR #795, opened 2025-10-09
use exact naming from baremetal test example

xdavidz:


xdavidz:
> Line 550 needs change to ofm_gold_path

fixed, thanks.

PR #802, opened 2025-10-13
Minor tweaks to build xdna-driver UMD for upstreaming with XRT

stsoe:
Refactor CMake such that upstreaming builds only the XDNA UMD plugin library while using XRT as provided by upstream project.

When building for upstreaming, it is assume that a top-level project provides XRT headers and targets as needed when building the xdna driver plugin library.

This changes in this PR simply disables the XRT part of xdna-driver when XRT_UPSTREAM is defined.

No changes for native xdna-driver builds.


AMDGithubSCIMAdmin:

maxzhen:
ok to test

PR #804, opened 2025-10-14
Ve2 runner to archive migration

ManojTakasi:
Archive files with the .a extension are not supported as part of /usr/... because the dwarfsrc files package attempts to extract debug symbols even though it's not an ELF binary.
Since this archive is part of XRT, we **cannot skip** the entire package from any checks. Therefore, the archive is downloaded as an ar extension package to avoid being scanned by QA checks.

This has been tested on the Telluride board using the generated Image/rootfs with these changes.

```
amd-edf:/home/amd-edf# xrt-smi validate -d 0
open : DEV name  /dev/accel/accel0
Validate Device           : [0000:00:00.0]
    Platform              : Telluride
    Power Mode            : Default
-------------------------------------------------------------------------------
Test 1 [0000:00:00.0]     : latency 
    Details               : Average latency: 31.0 us                            
    Test Status           : [PASSED]
-------------------------------------------------------------------------------
```

Currently, only the latency test is added. We will include more test cases once the corresponding ELF binaries and XCLBINs are ready.

### DIFF
--- a/src/shim_ve2/CMakeLists.txt
+++ b/src/shim_ve2/CMakeLists.txt
@@ -36,6 +36,13 @@ install (TARGETS ${XDNA_VE2_TARGET}
   LIBRARY DESTINATION ${XRT_INSTALL_LIB_DIR}
   )
 
-install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/Runner/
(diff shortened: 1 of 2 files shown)


ManojTakasi:
Thanks @aktondak, this is really helpful.

PR #809, opened 2025-10-16
amdxdna: ve2: update firmware version retrieval design

rajkumar-xilinx:
-Add hwctx state in ve2 driver
-Use amdxdna core API amdxdna_gem_dev_addr() for accessing device address


rajkumar-xilinx:
Verified the change on HW. All tests are passing.

chvamshi-xilinx (approved):

PR #812, opened 2025-10-17
CMA buf clean up

maxzhen:
This is a follow up PR for PR#801 to further clean up CMA buf implementation. Make it more aligned with carveout buf implementation. It also try to automatically detect if CMA is usable instead of forcibly set by driver.

### DIFF
--- a/src/driver/amdxdna/ve2_hwctx.c
+++ b/src/driver/amdxdna/ve2_hwctx.c
@@ -575,7 +575,7 @@ static int ve2_submit_cmd_chain(struct amdxdna_ctx *hwctx, struct amdxdna_sched_
 		void *cmd_data;
 		u32 cmd_data_len;
 
-		abo = amdxdna_gem_get_obj(hwctx->client, boh, AMDXDNA_BO_CMD);
+		abo = amdxdna_gem_get_obj(hwctx->client, boh, AMDXDNA_BO_SHARE);
 		if (!abo) {
 			XDNA_ERR(xdna, "Failed to find cmd BO %d", boh);
 			return -ENOENT;


src/driver/amdxdna/ve2_hwctx.c
```diff
@@ -934,8 +934,7 @@ int ve2_hwctx_config(struct amdxdna_ctx *hwctx, u32 type, u64 mdata_hdl, void *b
 			XDNA_ERR(xdna, "Get metadata bo %lld failed for type %d", mdata_hdl, type);
 			return -EINVAL;
 		}
-		cmabuf = mdata_abo->dma_buf->priv;
```
rajkumar-xilinx, line 937: please remove cmabuf declaration above

src/driver/amdxdna/amdxdna_cma_buf.c
```diff
@@ -164,3 +171,14 @@ struct dma_buf *amdxdna_get_cma_buf(struct drm_device *dev,
 	kfree(cmabuf);
 	return ERR_PTR(ret);
 }
+
+bool amdxdna_use_cma(void)
+{
+#ifdef CONFIG_CMA
+	static bool cma = !!dev_get_cma_area(NULL);
```
rajkumar-xilinx: Compile time error in using this API. /scratch/telluride/7oct_xdna/Telluride_public/XDNA/src/driver/amdxdna/build/driver/amdxdna/amdxdna_cma_buf.c: In function 'amdxdna_use_cma':
| /scratch/telluride/7oct_xdna/Telluride_public/XDNA/src/driver/amdxdna/build/driver/amdxdna/amdxdna_cma_buf.c:182:29: error: implicit declaration of function 'dev_get_cma_area' [-Werror=implicit-function-declaration]
|   182 |         static bool cma = !!dev_get_cma_area(NULL);
|       |                             ^~~~~~~~~~~~~~~~
| /scratch/telluride/7oct_xdna/Telluride_public/XDNA/src/driver/amdxdna/build/driver/amdxdna/amdxdna_cma_buf.c:182:27: error: initializer element is not constant
|   182 |         static bool cma = !!dev_get_cma_area(NULL);
|       |                           ^
| cc1: all warnings being treated as errors

rajkumar-xilinx: If we only need to check CMA availability, There is no direct API anymore. 
#ifdef CONFIG_DMA_CMA
              return true;  
#else
               return false;   
#endif

src/driver/amdxdna/ve2_hwctx.c
maxzhen: fixed.

src/driver/amdxdna/amdxdna_cma_buf.c
maxzhen: fixed.

src/driver/amdxdna/amdxdna_gem.c
```diff
@@ -835,10 +830,12 @@ amdxdna_drm_create_share_bo(struct drm_device *dev,
 
 	if (args->vaddr)
 		abo = amdxdna_gem_create_user_object(dev, args);
-	else if (xdna->use_cma)
+#ifdef AMDXDNA_DEVEL
```
xdavidz, line 833: my only concern is that we now have to set iommu_off to use either cma or carvdout.

I think in the past even with iommu on, we still can leverage the carvdout.

In some bringup situation, we might want to test physical address regardless if iommu is on/off.

maxzhen: When you have iommu on, you can use iova or pasid, why do you need physical address in that case?

xdavidz: > When you have iommu on, you can use iova or pasid, why do you need physical address in that case?

I agree, this is much cleaner design. So we have 3 combinations to support below:
1. pasid, allocate bo dynamatically (iommu_mode=0, plus correct grub, aka. no iommu enforce)
2. iova, allocate bo dynamatically (iommu_mode=1, plus correct grub, aka. intel_iommu=on or amd_iommu=force_isolation)
3. physical address from cma or carvedout (iommu=off, cma or memmap configured in grub)

I will wait for this PR merged then retest my npu3 bringup PRs.

PR #815, opened 2025-10-20
fix crash when hmm_invalidate is NULL

houlz0507:


maxzhen:
> it seems we keep the hmm_invalidate ops for subdriver to implement still? Or, we should remove this callback?

Once this PR is merged, you can update core and remove the aie4 callback.

PR #829, opened 2025-10-28
Add support for xrt-smi to change firmware log configs

NishadSaraf:
Add support for xrt-smi to change firmware log configs.

### DIFF
--- a/src/driver/amdxdna/amdxdna_dpt.c
+++ b/src/driver/amdxdna/amdxdna_dpt.c
@@ -287,31 +287,6 @@ static void amdxdna_dpt_enable_polling(struct amdxdna_dpt *dpt, bool enable)
 	dpt->polling = enable;
 }
 
-int amdxdna_dpt_dump_to_dmesg(struct amdxdna_dpt *dpt, bool dump)
-{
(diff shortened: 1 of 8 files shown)


src/driver/amdxdna/amdxdna_dpt.c
```diff
@@ -588,9 +563,29 @@ static int amdxdna_fw_trace_fini(struct amdxdna_dev *xdna)
...
+
+	if (dump) {
+		dpt->local_buffer = kzalloc(dpt->size, GFP_KERNEL);
+		if (!dpt->local_buffer) {
+			XDNA_ERR(dpt->xdna, "Failed to allocate FW fetch buffer");
+			return -ENOMEM;
+		}
+		amdxdna_dpt_enable_polling(dpt, true);
+		/* Drain any data already logged in the buffer before dump_to_dmesg was enabled */
+		amdxdna_dpt_drain_pending_data(dpt);
```
maxzhen, line 579: if dump is true, do you also want to drain right before resume?

NishadSaraf: Yes, that's already been implemented,

PR #830, opened 2025-10-28
ve2: Read cert's cmd completion status and set it

rajkumar-xilinx:
cert updates cmd completion status in host queue memory after command completes. ve2 driver should read this status and update the same in the driver instead of updating CMD_COMPLETED.


rajkumar-xilinx:
All hw tests passed. Tried stress_runs test multiple times, no issues observed.

PR #822, opened 2025-10-24
Ve2 aie_coredump changes

HimanshuChoudhary-Xilinx:
PR to support aie_coredump for ve2 platform
spec - https://amd.atlassian.net/wiki/spaces/AIE/pages/1098471914/AIE+Coredump

1. this api return needed size if buffer size is lesser then needed size
2. we split ve2_partition_read into multiple call for mem and core tile as 0x0 point to memory address

PR #832, opened 2025-10-28
Ve2 aie_coredump changes from PR #822

saifuddin-xilinx:
PR to support aie_coredump for ve2 platform
spec - https://amd.atlassian.net/wiki/spaces/AIE/pages/1098471914/AIE+Coredump

1. this api return needed size if buffer size is lesser then needed size
2. we split ve2_partition_read into multiple call for mem and core tile as 0x0 point to memory address

### DIFF
--- a/src/driver/amdxdna/ve2_hwctx.c
+++ b/src/driver/amdxdna/ve2_hwctx.c
@@ -755,10 +755,6 @@ int ve2_cmd_wait(struct amdxdna_ctx *hwctx, u64 seq, u32 timeout)
 		}
 
 		/*
-		 * amdxdna_cmd_set_state(job->cmd_bo,
-		 *                       priv_ctx->hwctx_hsa_queue.hq_complete.hqc_mem[seq]);
-		 */
-		/*
 		 * below check need to be removed once we have a clean solution
 		 * to use completion signal
 		 */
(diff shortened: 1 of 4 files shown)


src/include/uapi/drm_local/amdxdna_accel.h
```diff
@@ -516,6 +516,20 @@ struct amdxdna_drm_aie_reg {
...
+ * @context_id: The hw context id
+ * @size:  The size of bytes to read/write
+ * @buf_p: The buffer to store write data
+ *
+ * This is used for DRM_AMDXDNA_READ_AIE_COREDUMP parameters.
+ */
+struct amdxdna_drm_aie_coredump {
+	__u32 context_id;
+	__u32 size;
+	__u64 buf_p;
```
maxzhen, line 530: @NishadSaraf reminds me that the nested buffer schema does not work on virtio environment. Can we change it to a flat buffer with a header (in our case, the header contains just context id).

src/include/uapi/drm_local/amdxdna_accel.h
```diff
@@ -628,6 +642,7 @@ struct amdxdna_drm_get_info {
 #define	DRM_AMDXDNA_QUERY_RESOURCE_INFO			12
 #define	DRM_AMDXDNA_GET_FRAME_BOUNDARY_PREEMPT_STATE	13
 #define	DRM_AMDXDNA_QUERY_VE2_FIRMWARE_VERSION		14
+#define	DRM_AMDXDNA_READ_AIE_COREDUMP			15
```
maxzhen, line 645: We stopped extending get_info and switched to get_array. So, let's move both 14 and 15 to get_array. And you can remove 've2' from the name of the enum to make it generic.

saifuddin-xilinx:
Closing this as created a new pr #838 to track

PR #840, opened 2025-10-30
Fixed resolver. Added final set of changes from of_driver to here

saifuddin-xilinx:
This PR has the final set of changes from of_driver to here


rajkumar-xilinx:
When I added this change in aie2ps of driver (https://gitenterprise.xilinx.com/XRT/Telluride/pull/566 ), I verified the following test scenario to confirm if the sharing logic is working or not.
- Distribute temporal sharing requests more evenly across the [0–35] column range.
- This avoids repeated use of the same partition and improves resource utilization.
So, please try this test scenario, if it works then patch looks good to me.

chvamshi-xilinx (dismissed):

saifuddin-xilinx:
Closing this one. Same is created another PR #842

PR #845, opened 2025-10-31
Counting internal/external BO usage

maxzhen:
This is to support the XRT change: https://github.com/Xilinx/XRT/pull/9364

### DIFF
--- a/src/driver/amdxdna/amdxdna_gem.c
+++ b/src/driver/amdxdna/amdxdna_gem.c
@@ -419,16 +419,14 @@ amdxdna_gem_shmem_del_bo_usage(struct amdxdna_gem_obj *abo)
 {
 	struct amdxdna_client *client = abo->client;
 
-	// Imported BO should not be counted.
-	if (!client)
+	if (!abo->acct_total)
 		return;
 
 	mutex_lock(&client->mm_lock);
 
-	if (abo->acct_total) {
-		client->total_bo_usage -= abo->mem.size;
-		abo->acct_total = false;
-	}
+	client->total_bo_usage -= abo->mem.size;
+	abo->acct_total = false;
+
 	if (abo->acct_int) {
 		client->total_int_bo_usage -= abo->mem.size;
 		abo->acct_int = false;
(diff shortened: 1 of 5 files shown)


src/shim/buffer.cpp
```diff
@@ -176,6 +207,25 @@ bo_addr_align(int type)
...
+  auto flags = xcl_bo_flags{bo_flags};
+  auto boflags = (static_cast<uint32_t>(flags.boflags) << 24);
+  auto bouse = flags.use;
+
+  if (boflags != XCL_BO_FLAGS_HOST_ONLY &&
+      boflags != XCL_BO_FLAGS_CACHEABLE &&
+      boflags != XCL_BO_FLAGS_EXECBUF)
+    return AMDXDNA_BO_INVALID;
+
+  if (has_dev_mem && boflags == XCL_BO_FLAGS_CACHEABLE)
```
wendyliang25, line 222: just curious, the reason XCL_BO_FLAGS_CACHEABLE can still be set while `has_dev_mem` is not set, is it because setting flags to `XCL_BO_FLAGS_CACHEABLE` is common between AIE2 and AIE4? otherwise, just wonder if `!has_dev_mem && boflags == XCL_BO_FLAGS_CACHEABLE` is still valid? thanks

src/driver/amdxdna/amdxdna_gem.c
```diff
@@ -396,6 +396,47 @@ static void amdxdna_imported_obj_free(struct amdxdna_gem_obj *abo)
...
 
+static void
+amdxdna_gem_shmem_add_bo_usage(struct amdxdna_gem_obj *abo, bool internal)
+{
+	struct amdxdna_client *client = abo->client;
+
+	mutex_lock(&client->mm_lock);
+
+	client->total_bo_usage += abo->mem.size;
+	abo->acct_total = true;
```
wendyliang25, line 407: looks like `acct_total` is only used in `amdxdna_gem_shmem_del_bo_usage()` just wonder for bos which will go to `amdxdna_gem_shmem_del_bo_usage()`, does it always call `amdxdna_gem_shmem_add_bo_usage` in creation path? thanks

src/shim/buffer.cpp
maxzhen: Yes, code above SHIM layer can still use this flag without knowing if underlying device is of aie2 or aie4 platform.

src/driver/amdxdna/amdxdna_gem.c
maxzhen: The bo creation may fail before it reaches amdxdna_gem_shmem_add_bo_usage(), in which case, when it is eventually freed, we should not call amdxdna_gem_shmem_del_bo_usage().

maxzhen:
retest this please - sudo infra issue

wendyliang25 (approved):

wendyliang25 (approved):

PR #852, opened 2025-11-06
Fixing the gem_prime_import. Avoiding double free to avoid crash

bisingha-xilinx:
This pr fixes the crash due to double free of user ptr BOs if amdxdna_bo_dma_map() fails means if the user ptr BOs are not contiguous.

Earlier amdxdna_gem_prime_import() used to double free if amdxdna_bo_dma_map()'s contiguous check fails. drm_gem_object_put() will trigger the amdxdna_gem_shmem_obj_free() which already triggers the dma_buf_unmap_attachment_unlocked()/dma_buf_detach()/dma_buf_put(). 

In this fix just after putting the gem obj, im just returning the ERR_PTR.

### DIFF
--- a/src/driver/amdxdna/amdxdna_gem.c
+++ b/src/driver/amdxdna/amdxdna_gem.c
@@ -863,7 +863,7 @@ amdxdna_gem_prime_import(struct drm_device *dev, struct dma_buf *dma_buf)
 		ret = amdxdna_bo_dma_map(abo);
 		if (ret) {
 			drm_gem_object_put(gobj);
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx:
Merging as this is needed only for VAI_5_1. No corresponding XRT commit in TA

chvamshi-xilinx:
@maxzhen / @houlz0507 , Sorry, i have merged this change by mistake. Please reivew and suggest if this need to be reverted. 
I was supposed to merge some other change

PR #858, opened 2025-11-11
Update the module name and git hash

ManojTakasi:
Update the module name and git hash

### DIFF
--- a/src/driver/amdxdna/Makefile
+++ b/src/driver/amdxdna/Makefile
@@ -37,6 +37,12 @@ ifeq ($(XDNA_BUS_TYPE), of)
 DEFINES += -DAMDXDNA_OF
 endif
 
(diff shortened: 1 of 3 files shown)


src/driver/amdxdna/Makefile
```diff
@@ -37,6 +37,12 @@ ifeq ($(XDNA_BUS_TYPE), of)
 DEFINES += -DAMDXDNA_OF
 endif
 
+GIT_HASH := $(shell git rev-parse HEAD)
+ifndef MODULE_VER_STR
+MODULE_VER_STR := $(GIT_HASH)
+endif
+DEFINES += -DMODULE_VER_STR='\"$(MODULE_VER_STR)\"'
```
chvamshi-xilinx, line 44: As this is used only in of driver, you can make this change inside 36th column if condition

PR #859, opened 2025-11-12
Added memory usage reporting support per hw context

bisingha-xilinx:
This PR has changes to support memory usage reporting per hw context. Note as per https://jira.xilinx.com/browse/AIESW-14008 and as per offline discussion, memory usage of a context is the total bo usage of its client. So, if a client has n contexts, each context will have same memory usage. 

Along with memory usage support it prints corresponding PID, ctx state, ctx id, priority. Other parameters are kept as N/A only for now. Based on the requirement will enable others (eg. process name, etc)

1. Tested by having multi hw context in a single host application (single process scenario) -> PASSED. getting expected results with "xrt-smi examine -r aie-partitions -d 0"
2. Tested by having multi hw context in multiple host application (multi process scenario) -> PASSED. getting expected results with "xrt-smi examine -r aie-partitions -d 0"
3. currently by default from userspace passing arg for 32 contexts. if in the driver there are more than 32 contexts as a fallback userspace will try again with driver returned context count ->PASSED. getting expected results with "xrt-smi examine -r aie-partitions -d 0"

Here is a snippet of xrt-smi output (multi process scenario):
xrt-smi-memory-reporting

### DIFF
--- a/src/driver/amdxdna/ve2_debug.c
+++ b/src/driver/amdxdna/ve2_debug.c
@@ -11,73 +11,9 @@
 #include "ve2_mgmt.h"
 #include "ve2_res_solver.h"
 
-static int ve2_get_hwctx_status(struct amdxdna_client *client, struct amdxdna_drm_get_info *args)
-{
-	size_t hwctx_data_sz = sizeof(struct amdxdna_drm_query_hwctx);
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx:
retest this please

PR #863, opened 2025-11-14
Update handshake packet to align with firmware

khalshai-xilinx:
**Description**
This PR is to update the handshake packet structure. The changes are mainly focusing on command chain, preemption, tracing, synchronization.

restore_page :- This info is used in preemption along with command chain failure flag.
trace_save:- This info is also used in preemption for tracing feature.
DDR Access tracking:- The four registers give info of last executed ddr communication is error scenarios.
chain_flag :- This member is used to tell firmware if its last command in command chain or not.
completion_status :- This also used to give update on command execution in command chain.


AMDGithubSCIMAdmin:

chvamshi-xilinx:
ok to test

PR #867, opened 2025-11-17
driver: ve2: Set error_index properly for command chains

rajkumar-xilinx:
Add error_index tracking in ve2_hwctx.c to identify which subcommand failed in a command chain.
In timeout case Set error_index to 0, and in Error case traverse chain slots to find failing command and set its index.
It enables XRT core to properly handle subcommand failures in command chains.


rajkumar-xilinx:
WIP

PR #877, opened 2025-11-20
[XRT-SMI] Code cleanup and xrt update

aktondak:
This PR updates the XRT code to latest to include changes for npu3 event tracing 
This PR also removes all the artifact files which are not required anymore post archive migration

### DIFF
--- a/tools/bins/configs/trace_events.json
+++ b/tools/bins/configs/trace_events.json
@@ -1,0 +1,1164 @@
+{
+  "version": {
+    "product": 0,
+    "schema": 0,
+    "major": 1,
+    "minor": 0
+  },
+  "data_format": {
+    "event_bits": 16,
+    "payload_bits": 48
+  },
+  "lookups": {
+    "ipu_messages": {
+      "1": "IPU_MSG_REGISTER_PDI",
+      "2": "IPU_MSG_CREATE_CONTEXT",
+      "3": "IPU_MSG_DELETE_CONTEXT",
+      "4": "IPU_MSG_GET_TELEMETRY",
+      "5": "IPU_MSG_RESET_PARTITION",
+      "6": "IPU_MSG_EXECUTE_BUFFER",
+      "7": "IPU_MSG_SYNC_BO",
+      "8": "IPU_MSG_DPU_SELF_TEST",
+      "10": "IPU_MSG_UNREGISTER_PDI",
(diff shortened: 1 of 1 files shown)


maxzhen (dismissed):

NishadSaraf:
retest this please

chvamshi-xilinx (dismissed):

aktondak:
retest this please

aktondak:
retest this please

maxzhen:
Adding back a file to keep a directory alive??? Why do we need that directory if all files under it is no longer needed?

aktondak:
> Adding back a file to keep a directory alive??? Why do we need that directory if all files under it is no longer needed?

My best guess is that the amd-aie repo's build script has a call in its build.sh which assumes this folder is present and tries to copy data from it. I added it back just to check if the pipeline passes. If it does, the assertion is correct and this PR will become a multi-repo checkin where we'll have to remove the call and also "download_xbutil_artifacts()" from amd-aie as well

maxzhen:
> > Adding back a file to keep a directory alive??? Why do we need that directory if all files under it is no longer needed?
> 
> My best guess is that the amd-aie repo's build script has a call in its build.sh which assumes this folder is present and tries to copy data from it. I added it back just to check if the pipeline passes. If it does, the assertion is correct and this PR will become a multi-repo checkin where we'll have to remove the call and also "download_xbutil_artifacts()" from amd-aie as well

OK, this is for debugging purpose. That is fine.

aktondak:
> > > Adding back a file to keep a directory alive??? Why do we need that directory if all files under it is no longer needed?
> > 
> > 
> > My best guess is that the amd-aie repo's build script has a call in its build.sh which assumes this folder is present and tries to copy data from it. I added it back just to check if the pipeline passes. If it does, the assertion is correct and this PR will become a multi-repo checkin where we'll have to remove the call and also "download_xbutil_artifacts()" from amd-aie as well
> 
> OK, this is for debugging purpose. That is fine.

Looks like the assertion is right as it is passing now. We can do either of following: 
1. Merge this and then open amd-aie PR to fix build script and update xdna submodule. 
2. Wait for amd-aie fix to go in first and then merge this PR fully with all files removed. 
I'd prefer 1 as I need amd-aie to point to latest xdna as priority for event-trace feature, but let me know what's the best order.

maxzhen:
I'm fine with 1.

PR #891, opened 2025-11-29
Build fixes.

lubosz:
I propose 2 patches that involve the build system.

## CMake/pkg: Don't fail on unsupported Linux distros

Instead of emitting a `FATAL_ERROR`, emit a `WARNING`.
Besides this `FATAL_ERROR` the build succeeds.

This makes the build not to refuse to work on Arch Linux.

## shim/cmake: Use CMake generator expression to determine build type

Use the modern way of determining a debug build.
In addition, only set `XDNA_SHIM_DEBUG` on the `XDNA_TARGET`.

I tested this patch with all occurrences of `XDNA_SHIM_DEBUG` in the code
and it behaves as expected.

This removes the old string comparison which failed when
`CMAKE_BUILD_TYPE` was not set.

Following CMake error was triggered in that case:
```
CMake Error at src/shim/CMakeLists.txt:39 (if):
  if given arguments:

    "STREQUAL" "Debug"

  Unknown arguments specified
```

Instead of just wrapping the string comparison with "", which would
resolve the issue, I have decided to instead implement the modern way of
detecting debug mode.

See:
https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#conditional-expressions

### DIFF
--- a/src/shim/CMakeLists.txt
+++ b/src/shim/CMakeLists.txt
@@ -29,11 +29,8 @@ target_compile_definitions(${XDNA_TARGET} PRIVATE
   XRT_ENABLE_AIE
   XRT_AIE_BUILD
   XRT_BUILD
+  $<$<CONFIG:Debug>:XDNA_SHIM_DEBUG>
   )
-
-if(${CMAKE_BUILD_TYPE} STREQUAL "Debug")
-  add_definitions(-DXDNA_SHIM_DEBUG)
-endif()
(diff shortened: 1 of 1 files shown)


maxzhen:
ok to test

amd-gevadlam:
retest this please

AMDGithubSCIMAdmin:

PR #895, opened 2025-12-03
Adding further aie4 shim test support

hlaccabu:


src/shim/buffer.cpp
```diff
@@ -23,6 +23,8 @@ use_to_fw_debug_type(uint8_t use)
   switch (use) {
   case XRT_BO_USE_DEBUG:
     return AMDXDNA_FW_BUF_DEBUG;
+  case XRT_BO_USE_UC_DEBUG:
+    return AMDXDNA_FW_BUF_DEBUG;
```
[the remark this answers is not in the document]
xdavidz: This is reasonable. @hlaccabu can you add some comments here?

hlaccabu:
> With this, can we get a pass when run "shim_test" without any index number?

Currently yes, but there are a few more tests that I think I can enable with minimal changes before we merge.

hlaccabu:
> With this, can we get a pass when run "shim_test" without any index number?

Yes to this question on silicon, 38 tests passing.
{9E76890B-021C-4451-BAEA-550369D53810}

hlaccabu:
retest this please

xdavidz:
> > With this, can we get a pass when run "shim_test" without any index number?
> 
> Currently yes, but there are a few more tests that I think I can enable with minimal changes before we merge.

We should enable all shim_test and just run shim_test without any args.

Are there any shim_test that we decide to run for aie2 only? If not many, we should enable all possible shim_test for aie4.

hlaccabu:
> We should enable all shim_test and just run shim_test without any args.
> 
> Are there any shim_test that we decide to run for aie2 only? If not many, we should enable all possible shim_test for aie4.

Yes with this PR we can pass 38 shim tests on aie4 by running "./shim_test.elf" without any arguments, the rest are skipped. There are a few tests that will stay as aie2 only, more will be enabled for aie4 after this PR but I did not want to do every change all in one PR that would be too large.

src/shim/umq/hwq.h
```diff
@@ -56,6 +56,10 @@ class hwq_umq : public hwq
 
   uint64_t
   issue_single_exec_buf(const cmd_buffer *cmd_bo, bool last_of_chain);
+
+  // Cache for command buffer headers (firmware corrupts them)
+  // Key: BO handle, Value: cached header word
+  std::map<uint32_t, uint32_t> m_cmd_header_cache;
```
maxzhen: This may not work well. The exec buf can be reused by different applications...

test/shim_test/io.h
```diff
@@ -90,6 +99,7 @@ class io_test_bo_set_base
   const int m_FLAG_OPT =      1 << 1;
   const int m_FLAG_NO_FILL =  1 << 2;
   const int m_FLAG_DEV_BUF =  1 << 3;
+  static std::map<uint32_t, uint32_t> s_cmd_header_cache;
```
maxzhen: Shouldn't need a std::map here. Just a int32_t would do.

test/shim_test/io.h
```diff
@@ -90,6 +99,7 @@ class io_test_bo_set_base
   const int m_FLAG_OPT =      1 << 1;
   const int m_FLAG_NO_FILL =  1 << 2;
   const int m_FLAG_DEV_BUF =  1 << 3;
+  static uint32_t s_cached_subcmd_header;
```
maxzhen: Why use "s_" not "m_" as the rest of the code does?

hlaccabu: Since it's a static member variable vs. regular member variable, I can change it to m_ if preferred. I guess it doesn't need to be a static member, each instance of the class can have its own version.

xdavidz (approved):

maxzhen:
> Copilot encountered an error and was unable to review this pull request. You can try again by re-requesting a review.

We successfully confused AI :-)

PR #902, opened 2025-12-07
fix: Let the driver build on ArchLinux too

X-Ryl669:
The current main branch doesn't build on Arch linux (and all other derived distributions). 
Please notice that Arch linux is very barebone, so this PR is also useful for any *linux from scratch* distributions.

The required changes are small, but painful to apply as a patch since it implies ever evolving submodule hashes for xrt.
It would a be a lot easier if these changes were merged in.

# Description of the changes:
1. Set the CPACK package type to TGZ if it detects an Arch based distribution. This create a simple archive with all build files. The distribution's package will bootstrap from this archive instead of hacky unpacking the debian package (which doesn't contains all the required files)
2. Optionally allow to build the driver with Clang, since it's what being used in some Arch distributions.
3. Fix the issue in the code when building with Clang (missing C23 extension that was used to declare a variable in a `case` branch)

### DIFF
--- a/src/driver/CMakeLists.txt
+++ b/src/driver/CMakeLists.txt
@@ -138,15 +138,15 @@ install(FILES ${CMAKE_BINARY_DIR}/driver/${XDNA_TAR_GZ}
   COMPONENT ${XDNA_COMPONENT}
   )
 
-set(MAKE_DRV "cd driver/${XDNA_DRV_DIR}; make ${XDNA_DRV_BLD_FLAGS_DKMS} KERNEL_SRC=\${kernel_source_dir}; cd ../..")
-set(CLEAN_DRV "cd driver/${XDNA_DRV_DIR}; make clean KERNEL_SRC=\${kernel_source_dir}; cd ../..")
+set(MAKE_DRV "cd driver/${XDNA_DRV_DIR}; make ${XDNA_DRV_BLD_FLAGS_DKMS} KERNEL_SRC=\${kernel_source_dir}")
+set(CLEAN_DRV "cd driver/${XDNA_DRV_DIR}; make clean KERNEL_SRC=\${kernel_source_dir}")
 set(MODULE_DRV "BUILT_MODULE_NAME[0]=${XDNA_DRV}
 BUILT_MODULE_LOCATION[0]=\"driver/${XDNA_DRV_DIR}/build/driver/amdxdna\"
 DEST_MODULE_LOCATION[0]=\"/kernel/extras\"")
 
 if(XDNA_DRV_INT_SRC_DIR)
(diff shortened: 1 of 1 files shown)


AMDGithubSCIMAdmin:

X-Ryl669:
I've just noticed PR #866, and a part of this PR will be useless if #866 is merged. However, the part that builds with Clang isn't in #866 so it's still required. I'll update this PR once #866 is merged, please don't merge meanwhile to avoid double work.

kashif:
@X-Ryl669 apologies about the merge conflicts, want me to fix?

X-Ryl669:
Thank you for your offer. I've tried to build the latest version, but I've add many issues, so it was better for me to work out how to solve them practically. 

I think it's OK now. There are many things to fix in XRT repository, but it's not xdna-driver business so it's for another day. I've managed to avoid changing xrt submodule, so it can be merged in this repository.

src/driver/amdxdna/aie4_pci.c
```diff
@@ -723,7 +723,7 @@ static int aie4_pcidev_init(struct amdxdna_dev_hdl *ndev)
 	struct amdxdna_dev *xdna = ndev->xdna;
 	struct pci_dev *pdev = to_pci_dev(xdna->ddev.dev);
 	void __iomem *tbl[PCI_NUM_RESOURCES] = {0};
-	const struct firmware *npufw, *certfw;
+	const struct firmware *npufw = NULL, *certfw = NULL;
```
X-Ryl669, line 726: Without this, npufw can be used uninitialized later on in the function in the error handling code and clang errors out and refuse to build.

src/driver/CMakeLists.txt
```diff
@@ -138,15 +138,15 @@ install(FILES ${CMAKE_BINARY_DIR}/driver/${XDNA_TAR_GZ}
   COMPONENT ${XDNA_COMPONENT}
   )
 
-set(MAKE_DRV "cd driver/${XDNA_DRV_DIR}; make ${XDNA_DRV_BLD_FLAGS_DKMS} KERNEL_SRC=\${kernel_source_dir}; cd ../..")
-set(CLEAN_DRV "cd driver/${XDNA_DRV_DIR}; make clean KERNEL_SRC=\${kernel_source_dir}; cd ../..")
```
X-Ryl669, line 142: There's a constant error in Xilinx's code with dealing with PWD in DKMS build configuration. The dkms.conf MAKE and CLEAN directives aren't shell command, and they are run in their own subshell, so the PWD doesn't leak to the top subshell. 

In addition, on Clang built kernel, DKMS appends USE_LLVM=1 to the given MAKE command so it **must** be a make command not a shell script.

X-Ryl669: The same is true for all XRT dkms.conf.in files that all present the same issue. Hopefully, they aren't used in this repository for this driver so it's hidden under the rug.

maxzhen:
ok to test

PR #907, opened 2025-12-09
Resolver logic fixed. 

saifuddin-xilinx:
We have actually pick the solution from aiarm driver to here.


maxzhen:
retest this please

PR #922, opened 2025-12-13
Use unified firmware version for protocol compatibility

NishadSaraf:
Replace separate protocol_major/protocol_minor fields with a unified version number (major * 100 + minor) for simpler version comparisons. This design allows easy support for multiple firmware major versions by adding entries to the fw_ver_reqs array, and enables simple >= comparisons for feature/opcode availability checks.

### DIFF
--- a/src/driver/amdxdna/aie2_pci.h
+++ b/src/driver/amdxdna/aie2_pci.h
@@ -32,6 +32,11 @@
 #define AIE2_INTERVAL	20000	/* us */
 #define AIE2_TIMEOUT	1000000	/* us */
 
+/* Firmware version encoding: major in high 32 bits, minor in low 32 bits */
+#define AIE2_FW_VERSION(major, minor)	(((u64)(major) << 32) | (minor))
+#define AIE2_FW_MAJOR(version)		upper_32_bits(version)
+#define AIE2_FW_MINOR(version)		lower_32_bits(version)
+
 /* Firmware determines device memory base address and size */
 #define AIE2_DEVM_BASE	0x4000000
 #define AIE2_DEVM_SIZE	SZ_64M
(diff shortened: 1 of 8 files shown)


src/driver/amdxdna/aie2_pci.h
```diff
@@ -32,6 +32,11 @@
 #define AIE2_INTERVAL	20000	/* us */
 #define AIE2_TIMEOUT	1000000	/* us */
 
+/* Firmware version encoding: major * 100 + minor */
+#define AIE2_FW_VERSION(major, minor)	((major) * 100 + (minor))
```
maxzhen: it may be unlikely for fw minor to go beyond 100, but why do we have to have this assumption? can AIE2_FW_VERSION() simply return a u64 with high 32 bit for major and low 32 bit for minor?

src/driver/amdxdna/aie2_pci.h
```diff
@@ -398,6 +402,11 @@ struct aie2_hw_ops {
 	int (*set_dpm)(struct amdxdna_dev_hdl *ndev, u32 dpm_level);
 };
 
+struct aie2_supported_fw_ver {
+	u32 major;
```
maxzhen: Redundant, please remove.

src/driver/amdxdna/aie2_pci.h
```diff
@@ -406,16 +415,16 @@ enum aie2_fw_feature {
...
 };
 
 #define AIE2_FEATURE_ON(ndev, feature)	test_bit(feature, &(ndev)->feature_mask)
 
 struct amdxdna_dev_priv {
 	const char			*fw_path;
-	u64				protocol_major;
-	u64				protocol_minor;
+	const struct aie2_supported_fw_ver *supported_fw_vers;
+	u32				num_supported_fw_vers;
```
maxzhen: A single combined fw version is not enough?

src/driver/amdxdna/aie2_pci.h
NishadSaraf: will fix this

src/driver/amdxdna/aie2_pci.h
NishadSaraf: Should be, unless we need to deprecate support for a particular firmware version. As discussed offline, I'll maintain only the minimum version.

maxzhen (approved):

PR #935, opened 2025-12-18
Makefile changes to pick config_kerenl.h in yocto builds

ManojTakasi:
Makefile changes to pick config_kerenl.h in yocto builds

### DIFF
--- a/src/driver/tools/configure_kernel.sh
+++ b/src/driver/tools/configure_kernel.sh
@@ -22,7 +22,7 @@ fi
 
 # ---- Output header path ------------------------------------------------
 
-OUT="driver/amdxdna/config_kernel.h"
+OUT="${OUT:-driver/amdxdna/config_kernel.h}"
 
 # ---- Helper: try to compile a small snippet ----------------------------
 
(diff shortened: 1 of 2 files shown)


src/driver/tools/configure_kernel.sh
```diff
@@ -21,7 +21,7 @@ fi
 
 # ---- Output header path ------------------------------------------------
 
-OUT="driver/amdxdna/config_kernel.h"
+OUT="${OUT:-driver/amdxdna/config_kernel.h}"
```
maxzhen, line 25: Why making this change? The OUT needs to be set to this value. Otherwise, it won't work.

ManojTakasi: Hi Max,

In Yocto builds, we are not building the driver in the same location (src/driver/amdxdna). Instead, the driver is built in a nested directory:src/driver/amdxdna/build/driver/amdxdna/.

Because of this, the default OUT directory from configure_kernel.sh will not work. I wanted to invoke configure_kernel.sh from the Makefile in the directory where it is running. In the command, I’m passing the OUT variable so that if it’s not set, Bash uses the default value (driver/amdxdna/config_kernel.h).
I’m setting this in the Makefile as:
@KERNEL_SRC=$(KERNEL_SRC) **OUT="$(SRC_DIR)/config_kernel.h"** bash $(SRC_DIR)/../../../../tools/configure_kernel.sh

I have verified the build and tested it on hardware boards in Yocto builds, and it works as expected.

maxzhen: OK, sounds good, then.

chvamshi-xilinx (approved):

chvamshi-xilinx:
@maxzhen , please review and merge if change looks ok.

PR #939, opened 2025-12-19
Optimize DPT suspend/resume to preserve buffers across power transitions

NishadSaraf:
Previously, DPT suspend called full fini which freed DMA buffers and deallocated handles, while resume called full init which reallocated everything and cleared buffers. This was inefficient and caused loss of any unread log/trace data.

Optimize suspend/resume by introducing dedicated lightweight functions: Suspend path:
- Drain any pending log/trace data to dmesg before suspending
- Free IRQ and stop timers
- Preserve DMA buffer and handle across suspend
- Skip sending fw_log/trace_fini to FW
- Preserve dump_to_dmesg state in the handle itself

Resume path:
- Resubmit the same buffer to FW without clearing it
- Reinitialize IRQ
- Restart polling timers if they were active
- Restore dump_to_dmesg timer if it was enabled before suspend

With this optimization, the driver's u64 DPT tail pointer/offset ever increments across suspend/resume cycles as the firmware now resumes writing from where it left off.

Also fix timer cleanup in fini functions:
- Reset timer_refs refcount to 0 before shutdown
- Use timer_shutdown_sync() instead of timer_delete_sync()

### DIFF
--- a/src/driver/amdxdna/amdxdna_dpt.c
+++ b/src/driver/amdxdna/amdxdna_dpt.c
@@ -21,7 +21,7 @@
 #define AMDXDNA_DPT_FW_LOG_MAX_DEFAULT		1
 #define AMDXDNA_DPT_FW_TRACE_MAX_DEFAULT	(~0)
 
-static u8 fw_log_level = 1;
+static u8 fw_log_level;
 module_param(fw_log_level, byte, 0444);
 MODULE_PARM_DESC(fw_log_level,
 		 " Firmware log verbosity: 0: DISABLE 1: ERROR (Default) 2: WARN 3: INFO 4: DEBUG");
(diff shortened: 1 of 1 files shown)


src/driver/amdxdna/amdxdna_dpt.c
```diff
@@ -21,7 +21,7 @@
 #define AMDXDNA_DPT_FW_LOG_MAX_DEFAULT		1
 #define AMDXDNA_DPT_FW_TRACE_MAX_DEFAULT	(~0)
 
-static u8 fw_log_level = 1;
+static u8 fw_log_level;
 module_param(fw_log_level, byte, 0444);
 MODULE_PARM_DESC(fw_log_level,
 		 " Firmware log verbosity: 0: DISABLE 1: ERROR (Default) 2: WARN 3: INFO 4: DEBUG");
```
maxzhen: fix the comment? default now is disabled, right?

PR #950, opened 2025-12-23
command chain fix in case of timeout

HimanshuChoudhary-Xilinx:
command chain fix in case of timeout
TODO: get right index from where hang started in error_index


chvamshi-xilinx:
retest this please

PR #952, opened 2025-12-23
vxdna: build: not warn for missing DRM_IOCTL_SET_CLIENT_NAME

wendyliang25:
DRM_IOCTL_SET_CLIENT_NAME is introduced since 6.13, change cmake warning to information message if it is missing.


wendyliang25:
ok to test

PR #956, opened 2025-12-26
fix BO offset and size check

maxzhen:


chvamshi-xilinx (approved):

src/driver/amdxdna/amdxdna_gem.c
```diff
@@ -1211,7 +1211,9 @@ int amdxdna_drm_sync_bo_ioctl(struct drm_device *dev,
 	}
 	abo = to_xdna_obj(gobj);
 
-	if (gobj->size < args->offset + args->size) {
+	if (gobj->size < args->offset ||
+	    gobj->size < args->size   ||
+	    args->offset > gobj->size - args->size) {
```
wendyliang25, line 1216: minor, just to keep this line consistent to the above two lines, maybe change it to `gobj->size < args->offset + args->size`

wendyliang25 (approved):

src/driver/amdxdna/amdxdna_gem.c
maxzhen: this is done to avoid overflow from args->offset + args->size. gobj->size - args->size is safe since line 1215 has done the check.

PR #962, opened 2026-01-05
sync up with aie4_msg_priv.h for 5.13 version

xdavidz:
@hlaccabu this is the 1st step to make mailbox message definition aligned with latest fw.
Then, I will clean up several unused mailbox messages.
Finally, I will bring the async message working for aie4. This is a must for debugging.

Please hold off merging until I fully test this.

### DIFF
--- a/src/driver/amdxdna/aie4_devel.c
+++ b/src/driver/amdxdna/aie4_devel.c
@@ -39,7 +39,7 @@
  */
 int aie4_hello_test(struct amdxdna_dev_hdl *ndev, struct amdxdna_ctx *ctx)
 {
-	DECLARE_AIE4_MSG(aie4_create_hw_context, AIE4_MSG_OP_CREATE_HW_CONTEXT);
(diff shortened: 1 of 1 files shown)


src/driver/amdxdna/aie4_pci.c
```diff
@@ -797,7 +797,7 @@ static int aie4_msg_destroy_context(struct amdxdna_dev_hdl *ndev, u32 hw_context
 
 int aie4_create_context(struct amdxdna_dev_hdl *ndev, struct amdxdna_ctx *ctx)
 {
-	DECLARE_AIE4_MSG(aie4_create_hw_context, AIE4_MSG_OP_CREATE_HW_CONTEXT);
```
hlaccabu, line 800: Looks like this change is needed in aie4_devel.c as well for aie4_create_hw_context to aie4_msg_create_hw_context

xdavidz: great catch. I will fix this.

PR #972, opened 2026-01-06
Adding command retry logic in driver level

saifuddin-xilinx:
Pull request overview
This PR implements a retry mechanism for handling full hardware queues in the AMD XDNA driver. When command submission fails due to a full queue, the driver now waits (using IRQ-driven events) for slots to become available rather than immediately failing. The implementation includes a 5-second timeout and handles both single commands and command chains with partial submission support.

Key changes:
Introduces IRQ-driven wait mechanism with 5-second timeout for queue slot availability
Splits queue slot reservation and commit into separate phases to enable parallel packet preparation
Adds partial submission support for command chains to resume from interruption point


saifuddin-xilinx:
Created a new PR after resolving the conflicts
https://github.com/amd/xdna-driver/pull/976

PR #973, opened 2026-01-06
Adding npu3 fw to info.json and configuring build.sh for download

hlaccabu:
Needed to change the build.sh. Since npu3 needs 2 firmwares, checking if the directory exists for the device ID would remove the directory with 1 of the updated fws before script could add the second fw. Better to check if exact file exists.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -94,7 +94,7 @@ download_npufws()
       fi
 
       echo "Download $device NPUFW version $version:"
-      if [ -d "${firmware_dir}/${pci_dev_id}_${pci_rev_id}" ]; then
+      if [ -f "${firmware_dir}/${pci_dev_id}_${pci_rev_id}/$fw_name" ]; then
         rm -r ${firmware_dir}/${pci_dev_id}_${pci_rev_id}
       fi
       mkdir -p ${firmware_dir}/${pci_dev_id}_${pci_rev_id}
--- a/tools/info.json
+++ b/tools/info.json
@@ -22,6 +22,22 @@
 			"fw_name": "npu.dev.sbin"
 		},
 		{
+			"device": "npu3",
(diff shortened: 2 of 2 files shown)


build/build.sh
```diff
@@ -94,7 +94,7 @@ download_npufws()
       fi
 
       echo "Download $device NPUFW version $version:"
-      if [ -d "${firmware_dir}/${pci_dev_id}_${pci_rev_id}" ]; then
+      if [ -f "${firmware_dir}/${pci_dev_id}_${pci_rev_id}/$fw_name" ]; then
```
xdavidz, line 97: I don't quite understand this operation. This is in a loop, say we have 2 fw_name "npu.dev.sbin and cert.dev.sbin".

Then the 1st loop will remove the dir of /lib/firmware/amdnpu/17f1 because there is a npu_dev.sbin.
The 2nd loop will then check cert.dev.sbin, the script will remove the /lib/firmware/amdnpu/17f1 again because cert.dev.sbin is not there?

I think for clear flow, we either remove the file npu.dev.sbin or cert.dev.sbin.
or,
we move this dir check above the loop?

hlaccabu: No, first loop will check for 17f1_10/npu.dev.sbin. If that exists, it will remove 17f1_10 dir (both npu and cert files) then create 17f1_10/npu.dev.sbin. Next loop, it will check for 17f1_10/cert.dev.sbin, which has already been removed, so it skips removing 17f1_10. This will end with both fws being in 17f1_10, updated to the latest versions.

xdavidz: > No, first loop will check for 17f1_10/npu.dev.sbin. If that exists, it will remove 17f1_10 dir (both npu and cert files) then create 17f1_10/npu.dev.sbin. Next loop, it will check for 17f1_10/cert.dev.sbin, which has already been removed, so it skips removing 17f1_10. This will end with both fws being in 17f1_10, updated to the latest versions.

ok to me.

PR #975, opened 2026-01-07
skip per-page fault loop for CMA buffers

bisingha-xilinx:
### Problem and Fix for https://jira.xilinx.com/browse/AIESW-21501:
Buffer Object (BO) import time was scaling linearly O(n) with buffer size,
causing significant latency in VART applications. Import time increased
from ~250μs for 1.5MB buffers to ~2900μs for 24MB buffers.

### Root Cause:
In amdxdna_gem_shmem_insert_pages(), after calling dma_buf_mmap() for
imported buffers, a per-page handle_mm_fault() loop was executed to fault
in each page. For CMA buffers, this is redundant because dma_buf_mmap() 
already establishes the complete mapping efficiently.

### Solution:
Add a check to detect CMA buffers and skip the
per-page fault loop for them. Non-CMA buffer types retain the existing
behavior for backward compatibility with other platforms.

### Results (24MB buffer import):
- Before: 2908 μs (O(n) - ~0.48 μs/page)
- After:    24 μs (O(1))
- Speedup: 121x

### Tested:
1. all the basic XRT hw tests on ve2
2. bo import latency test. giving expected constant latency O(1) by increasing the bo size. earlier it was O(n).
3. Also tested the given VART application from https://jira.xilinx.com/browse/AIESW-21501 (I ran the ./x_plus_ml_vart_old). With this fix, there is an overall latency improvement from "Inference thread 0 average time : 7.35578 ms" to "Inference thread  0 average time : 5.86632 ms".

### DIFF
--- a/src/driver/amdxdna/ve2_debug.c
+++ b/src/driver/amdxdna/ve2_debug.c
@@ -5,6 +5,9 @@
 #include <linux/device.h>
 #include <linux/version.h>
 #include <linux/vmalloc.h>
+#include <linux/completion.h>
+#include <linux/jiffies.h>
+#include <linux/sched.h>
 
 #include "ve2_fw.h"
 #include "ve2_of.h"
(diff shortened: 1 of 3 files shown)


src/driver/amdxdna/amdxdna_gem.c
```diff
@@ -518,19 +518,28 @@ static int amdxdna_gem_shmem_insert_pages(struct amdxdna_gem_obj *abo,
...
-			vma->vm_ops->close(vma);
-			XDNA_ERR(xdna, "Fault in page failed");
-			return -EFAULT;
-		}
+	/*
+	 * For CMA buffers (nents == 1), the exporter's mmap handler called
+	 * via dma_buf_mmap() establishes the complete mapping.
+	 * For non-contiguous buffers, we must fault in each page individually.
+	 */
+	if (abo->base.sgt && abo->base.sgt->nents == 1) {
```
houlz0507: please use !xdna->notifier_wq. Inserting the pages immediately is mainly for SVM.

bisingha-xilinx: Hi @houlz0507, thank you for quickly reviewing it. I've updated the check accordingly.
Thank you.

houlz0507 (approved):

chvamshi-xilinx (approved):

PR #989, opened 2026-01-10
vxdna update: throw std::invalid_argument if virtio gpu device is not NPU and coverity, clang tidy defects fixes

wendyliang25:
shim: virtio: throw std::invalid_argument when opening non NPU dev

    When the NPU application starts, it scans potential devices to identify which
    one is an NPU device. XRT continues scanning only when it encounters a
    std::invalid_argument exception, but not other exceptions.
    In the case of a virtio VM, this patch modifies the shim platform's virtio
    implementation to throw std::invalid_argument instead of xrt_core::system_error,
    allowing XRT to keep scanning

vxdna: hwctx: add defensive cleanup path for std::thread ctor failure

    The hwctx constructor launches a polling thread to retire fences.
    If std::thread throws (e.g., due to resource exhaustion), we should
    tear down the already-created kernel objects to avoid leaks.

    Add a try/catch around thread creation:
    - Destroy syncobj via DRM_IOCTL_SYNCOBJ_DESTROY (m_syncobj_handle)
    - Destroy hwctx via DRM_IOCTL_AMDXDNA_DESTROY_HWCTX (m_hwctx_handle)
    - Rethrow to propagate the constructor failure

    This keeps hwctx construction exception-safe and prevents dangling
    DRM handles.

* coverity defects:
      * CID 899062: Performance inefficiencies (COPY_INSTEAD_OF_MOVE)
      * errno Unreliable After ifstream::open() (Coverity: ERRNO_USAGE)
    * some clang-tidy warnings:
      * misc-include-cleaner
      * modernize-deprecated-headers
      * C-Style Cast

    Integer overflow prevention:
    - Add overflow check before iovec buffer allocation in vxdna_bo constructor
    - Add overflow check for driver-returned array sizes in get_info()

    Exception safety and noexcept:
    - Mark ~vxdna_bo() and ~vxdna_hwctx() destructors as noexcept
    - Add [[nodiscard]] to export_resource_fd() to prevent ignored return values

    Resource safety:
    - Fix ERRNO_CLOBBER: save errno before close() in os_dupfd_cloexec()

    Type safety and consistency:
    - Use AMDXDNA_INVALID_BO_HANDLE constant instead of magic number 0
    - Add explicit static_cast for buffer.size() to int32_t in read_sysfs()

    Modern C++ style:
    - Replace C headers with C++ equivalents (<cstdint>, <cstring>, etc.)
    - Replace (void)param cast pattern with [[maybe_unused]] attribute
    - Apply [[maybe_unused]] to unused device parameters in ccmd handlers
    - Apply [[maybe_unused]] to ctx_flags and name in create_ctx()

    Behavior improvement:
    - Log info message instead of throwing when DRM_IOCTL_SET_CLIENT_NAME
      is unavailable (graceful degradation)

### DIFF
--- a/src/vxdna/src/vaccel_amdxdna.cpp
+++ b/src/vxdna/src/vaccel_amdxdna.cpp
@@ -11,14 +11,13 @@
 #include <cerrno>
 #include <cstddef>
 #include <cstdint>
+#include <cstdlib>
 #include <cstring>
 #include <fcntl.h>
 #include <filesystem>
 #include <fstream>
 #include <climits>
 #include <sstream>
-#include <cstdlib>
-#include <cstring>
 #include <sys/stat.h>
 #include <thread>
 #include <vector>
@@ -105,6 +104,8 @@ vxdna_bo(const std::shared_ptr<vaccel_resource> &res, int ctx_fd_in,
     auto num_iovs = res->get_iovecs(&iovecs);
 
     // Use vector to avoid VLA and potential stack overflow
+    if (num_iovs > ((UINT32_MAX - sizeof(amdxdna_drm_va_tbl)) / sizeof(amdxdna_drm_va_entry)))
+        VACCEL_THROW_MSG(-EINVAL, "Too many iovecs: %u", num_iovs);
     size_t buf_size = sizeof(amdxdna_drm_va_tbl) + sizeof(amdxdna_drm_va_entry) * num_iovs;
     std::vector<uint8_t> buf_vec(buf_size);
     auto tbl = reinterpret_cast<amdxdna_drm_va_tbl*>(buf_vec.data());
@@ -216,7 +217,7 @@ vxdna_bo(const std::shared_ptr<vaccel_resource> &res, int ctx_fd_in,
 }
 
 vxdna_bo::
-~vxdna_bo()
+~vxdna_bo() noexcept
 {
     vxdna_dbg("vxdna Destroying bo: ctx_fd=%d, handle=%u, vaddr=%lx, map_size=%lu",
                m_ctx_fd, m_bo_handle, m_vaddr, m_map_size);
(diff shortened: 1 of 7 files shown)


src/vxdna/src/vaccel_amdxdna.cpp
```diff
@@ -290,31 +292,44 @@ vxdna_hwctx(const vxdna_context &ctx,
...
+    } catch (...) {
+        vxdna_err("hwctx(): failed to start polling thread.");
+        // Clean up hwctx and syncobj
+        struct drm_syncobj_destroy sync_arg = {};
+        sync_arg.handle = m_syncobj_handle;
+        ioctl(m_ctx_fd, DRM_IOCTL_SYNCOBJ_DESTROY, &sync_arg);
+
+        struct amdxdna_drm_destroy_hwctx hwctx_arg = {};
+        hwctx_arg.handle = m_hwctx_handle;
+        ioctl(m_ctx_fd, DRM_IOCTL_AMDXDNA_DESTROY_HWCTX, &hwctx_arg);
```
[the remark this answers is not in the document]
wendyliang25: ioctl() is a system call, there is no wrapper for ioctl in vxdna

src/shim/virtio/platform_virtio.cpp
```diff
@@ -373,7 +373,7 @@ drv_open(const std::string& sysfs_name) const
 
   auto fd = dev_fd();
   if (get_capset(fd) != VIRTGPU_DRM_CONTEXT_AMDXDNA)
-    shim_err(EINVAL, "%s is not NPU device", sysfs_name.c_str());
+    throw std::invalid_argument(sysfs_name + " is not NPU device");
```
maxzhen, line 381: Need a comment to remind us why shim_err is not appropriate.

wendyliang25: Thanks @maxzhen , i have update the patch with comments

PR #992, opened 2026-01-13
Fixed ve2 build

saifuddin-xilinx:


chvamshi-xilinx:
Hi @houlz0507 ,
Merging as this is needed for ve2 . Please send a separate PR if this needs to be fixed differrently

PR #1000, opened 2026-01-15
Integer overflow in AMD XDNA driver

saifuddin-xilinx:


maxzhen (changes requested):
This change won't work well. The args->size is not used in user pointer bo creation, checking it right after ioctl will reject valid user bo creation request. I'd suggest that you can create a wrapper for PAGE_ALIGN(args->size) where you can do the check. Then, replace where PAGE_ALIGN() is used against args->size. This should work.

src/driver/amdxdna/amdxdna_gem.c
```diff
@@ -1015,6 +1042,15 @@ int amdxdna_drm_create_bo_ioctl(struct drm_device *dev, void *data, struct drm_f
...
 		 args->type, args->vaddr, args->size, args->flags);
+
+	/* Validate size to prevent PAGE_ALIGN integer overflow */
+	if (args->size == 0 || args->size > SIZE_MAX - (PAGE_SIZE - 1)) {
+		if (args->size == 0)
+			XDNA_ERR(xdna, "Invalid zero size for BO creation");
+		else
+			XDNA_ERR(xdna, "Size 0x%llx would overflow PAGE_ALIGN", args->size);
+		return -EINVAL;
+	}
```
maxzhen: shouldn't you remove these lines?

maxzhen (changes requested):

src/driver/amdxdna/amdxdna_gem.c
saifuddin-xilinx: Oops!! Missed this. Sorry for that.
Updated. Thanks for reviewing.

maxzhen (approved):

PR #1013, opened 2026-01-20
Add git_hash field for npu3 cert firmware in info.json

NishadSaraf:
Add git_hash field for npu3 cert firmware in info.json


NishadSaraf:
retest this please

PR #1014, opened 2026-01-20
Updated the placement of initialization of capacity variable.

khalshai-xilinx:
**Description**
This PR relocates the initialization of the capacity field in the HSA queue header to occur earlier in the function, before the loop that sets queue slots to invalid. The change moves the capacity assignment and its associated power-of-2 validation check from after the slot initialization loop to before it.

Changes:

Moved capacity field initialization and validation earlier in the function execution flow


AMDGithubSCIMAdmin:

chvamshi-xilinx:
ok to test

PR #1025, opened 2026-01-23
Even more aie4 shim test enablement

hlaccabu:
42 passing shim tests with these changes

### DIFF
--- a/test/shim_test/shim_test.cpp
+++ b/test/shim_test/shim_test.cpp
@@ -837,10 +837,9 @@ std::vector<test_case> test_list {
   test_case{ "measure real kernel latency", {},
     TEST_POSITIVE, dev_filter_is_aie2, TEST_io_latency, { IO_TEST_NORMAL_RUN, IO_TEST_IOCTL_WAIT, 32000, 0 }
   },
-  //Skip this test for now until b2b vadd is fixed
-  //test_case{ "measure real kernel latency (aie4)", {},
-    //TEST_POSITIVE, dev_filter_is_aie4, TEST_io_latency, { IO_TEST_NORMAL_RUN, IO_TEST_IOCTL_WAIT, 32000, 1 }
-  //},
+  test_case{ "measure real kernel latency (aie4)", {},
+    TEST_POSITIVE, dev_filter_is_aie4, TEST_io_latency, { IO_TEST_NORMAL_RUN, IO_TEST_IOCTL_WAIT, 32000, 1 }
+  },
   test_case{ "create and free debug bo", {-1, -1},
     TEST_POSITIVE, dev_filter_xdna, TEST_create_free_debug_bo, { 0x1000 }
   },
@@ -946,16 +945,15 @@ std::vector<test_case> test_list {
   test_case{ "Multi context IO test 4 (npu4)", {},
     TEST_POSITIVE, skip_dev_filter, TEST_multi_context_io_test, { 20, 0 }
   },
-  // Skip this test for now until b2b vadd is fixed
-  //test_case{ "Multi context IO test 1 (npu3)", {},
-   // TEST_POSITIVE, dev_filter_is_npu3, TEST_multi_context_io_test, { 4, 1 }
-  //},
-  //test_case{ "Multi context IO test 2 (npu3)", {},
-    //TEST_POSITIVE, dev_filter_is_npu3, TEST_multi_context_io_test, { 8, 1 }
-  //},
-  //test_case{ "Multi context IO test 3 (npu3)", {},
-    //TEST_POSITIVE, skip_dev_filter, TEST_multi_context_io_test, { 32, 1 }
-  //},
+  test_case{ "Multi context IO test 1 (npu3)", {},
+    TEST_POSITIVE, dev_filter_is_npu3, TEST_multi_context_io_test, { 4, 1 }
+  },
+  test_case{ "Multi context IO test 2 (npu3)", {},
+    TEST_POSITIVE, dev_filter_is_npu3, TEST_multi_context_io_test, { 16, 1 }
+  },
+  test_case{ "Multi context IO test 3 (npu3)", {},
+    TEST_POSITIVE, dev_filter_is_npu3, TEST_multi_context_io_test, { 20, 1 }
+  },
   test_case{ "Create and destroy devices", {},
     TEST_POSITIVE, dev_filter_xdna, TEST_create_destroy_device, {}
   },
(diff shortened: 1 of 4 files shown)


hlaccabu:
> @hlaccabu have you verified this on board 10+ times? make sure it is stable enough please.

yes just went 10/10 on silicon, all 42 tests passed each time

test/shim_test/shim_test.cpp
```diff
@@ -379,9 +388,11 @@ TEST_create_destroy_virtual_context(device::id_type id, std::shared_ptr<device>&
   int is_negative = static_cast<unsigned int>(arg[0]);
   int num_virt_ctx;
 
-  // XDNA driver by default supports 6 virtual context on npu1 and 32 virtual context on npu4
+  // XDNA driver by default supports 6 virtual context on npu1, 128 on npu3, and 32 virtual context on npu4
   if (device_id == npu1_device_id)
     num_virt_ctx = 6;
+  else if (device_id == npu3_device_id)
```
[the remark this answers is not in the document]
hlaccabu: @xdavidz Can VF do same amount of ctxs?

src/driver/amdxdna/aie4_message.c
```diff
@@ -304,9 +304,11 @@ int aie4_query_aie_telemetry(struct amdxdna_dev_hdl *ndev, u32 type, dma_addr_t
 		return -EINVAL;
 	}
 
+	req.type = (type == 0) ? AIE4_TELEMETRY_TYPE_PERF_COUNTER : type;
 	req.buf_addr = addr;
+	req.pasid.raw = 0;
 	req.buf_size = size;
-	req.type = type;
+	req.hw_context_id = 0;
```
xdavidz, line 299: why set context_id to 0? I think 0 is a valid context_id, isn't it?

src/driver/amdxdna/aie4_pci.c
```diff
@@ -1612,14 +1612,12 @@ static int aie4_query_telemetry(struct amdxdna_client *client,
 
 	memset(buff, 0, aligned_sz);
 	drm_clflush_virt_range(buff, aligned_sz); /* device can access */
-	ret = aie4_query_aie_telemetry(ndev, type, dma_addr + sizeof(u64), aligned_sz);
```
xdavidz, line 1726: what is the reason behind this change that is different than the aie2? I am going to delete dup code between aie2 and aie4 even

src/driver/amdxdna/aie4_pci.c
```diff
@@ -1612,14 +1612,12 @@ static int aie4_query_telemetry(struct amdxdna_client *client,
...
+	ret = aie4_query_aie_telemetry(ndev, type, dma_addr, aligned_sz);
 	if (ret) {
 		XDNA_ERR(xdna, "Get telemetry failed ret %d", ret);
 		goto free_buf;
 	}
 
-	print_hex_dump_debug("telemetry: ", DUMP_PREFIX_OFFSET, 16, 4, buff,
```
xdavidz, line 1732: removing this will make dmesg disabled for debugging. can we have this line unchanged or it is flooding?

src/shim/device.cpp
```diff
@@ -963,8 +963,53 @@ struct telemetry
   static constexpr uint32_t NPU_MAX_SLEEP_COUNT = 9;
   static constexpr uint32_t NPU_MAX_OPCODE_COUNT = 30;
   static constexpr uint32_t NPU_MAX_DTLB_COUNT = 12;
+  static constexpr uint16_t NPU3_DEVICE_ID = 0x17f1;
+  static constexpr uint16_t NPU3A_DEVICE_ID = 0x17f3;
```
xdavidz: The telemetry mailbox message is for classic and PF driver right?
Then we need 17f2.

Or it is vf only? Each vf has its own telemetry? PF should not send this out?

src/shim/device.cpp
```diff
@@ -992,11 +1037,49 @@ struct telemetry
...
+    bool is_npu4 = (device_id == NPU4_DEVICE_ID);
+
     switch (key) {
     case key_type::aie_telemetry:
     {
       query::aie_telemetry::result_type output;
 
+      auto& pci_dev_impl = get_pcidev_impl(device);
+
+      if (is_npu3) {
```
xdavidz: The device.cpp is production code not test code right?

I do not like the current way that we handle the npu3 in if else statement. This will make the OOO cpp looks like a C and we keep adding if (is_npu3) if (is_npu3a) etc. Any better design in the CPP?

Define a class?

test/shim_test/shim_test.cpp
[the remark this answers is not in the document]
xdavidz: can you test if you can get some number? I think VF can.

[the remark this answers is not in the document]
hlaccabu: Sure, I can test that!

src/driver/amdxdna/aie4_message.c
hlaccabu: Because the request is for global telemetry (not ctx specific) so fill the field with 0. This is consistent with windows implementation.

src/driver/amdxdna/aie4_pci.c
hlaccabu: It is flooding, since telemetry info is large it completely fills the dmesg. We can figure something else out in between, maybe just print first couple of lines?

src/shim/device.cpp
hlaccabu: Ah yes you're correct, let me fix that

src/shim/device.cpp
hlaccabu: Yes good idea, let me clean this up.

src/driver/amdxdna/aie4_pci.c
hlaccabu: This has been removed from aie2 as well.
in line 994 of aie2_pci.c:
ret = aie2_query_aie_telemetry(xdna->dev_handle, dma_hdl, header.type, size, &ver);

We can still probably eliminate dup code here

test/shim_test/shim_test.cpp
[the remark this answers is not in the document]
hlaccabu: Added implementation based on Max's suggestion, works well

hlaccabu:
@xdavidz addressed several of your comments and made some enhancements, let me know what you think. All tests passing on classic, 1 preemption test fails on VF due to issues we spoke about.

maxzhen:
retest this please

hlaccabu:
Closing to create several smaller PRs with these changes

PR #1030, opened 2026-01-26
Deprecate NPU 2 (Strix A0) device support

NishadSaraf:
Remove NPU 2 (Strix A0) device support as it is obsolete and no longer needed.


NishadSaraf:
retest this please

PR #1033, opened 2026-01-27
Fix fw unload/reload issue

hlaccabu:


src/driver/amdxdna/aie4_psp.c
```diff
@@ -130,6 +130,17 @@ void aie4_psp_stop(struct psp_device *psp)
...
 
+void aie4_psp_destroy(struct device *dev, void *psp_hdl)
+{
+	struct psp_device *psp = psp_hdl;
+
+	if (!psp)
+		return;
+
+	psp->fw_buffer = NULL;
+	psp->certfw_buffer = NULL;
```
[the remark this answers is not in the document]
hlaccabu: buffers allocated using devm_kzalloc, automatically freed

src/driver/amdxdna/aie4_pci.c
```diff
@@ -1181,12 +1184,24 @@ static void aie4_pcidev_fini(struct amdxdna_dev_hdl *ndev)
...
 	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
 
 	aie4_iommu_fini(ndev);
 
 	aie4_pcidev_fini(ndev);
 
+	if (ndev->psp_hdl) {
+		aie4_psp_destroy(&pdev->dev, ndev->psp_hdl);
+		ndev->psp_hdl = NULL;
+	}
```
NishadSaraf: Same as below.

src/driver/amdxdna/aie4_pci.c
```diff
@@ -1181,12 +1184,24 @@ static void aie4_pcidev_fini(struct amdxdna_dev_hdl *ndev)
...
+
+	if (ndev->npufw || ndev->certfw) {
+		aie4_release_firmware(ndev, ndev->npufw, ndev->certfw);
+		ndev->npufw = NULL;
+		ndev->certfw = NULL;
+	}
```
NishadSaraf: Safer to do the args check, and NULL init inside `aie4_release_firmware()`

NishadSaraf (changes requested):
Thanks for the quick debug. Minor comments. Rest looks good to me.

src/driver/amdxdna/aie4_psp.c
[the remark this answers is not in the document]
maxzhen: Devm_xxx managed resources are released when driver is unloaded or detached, in which case, there is no need to set pointers to NULL. If you still need to remember to set pointers to NULL, you might as well just release the resources yourself. What is the point of asking devm_ to manage it for you?

src/driver/amdxdna/aie4_pci.c
```diff
@@ -530,6 +530,9 @@ static void aie4_hw_stop(struct amdxdna_dev *xdna)
 
 	aie4_fw_unload(ndev);
 
+	/* set mailbox alive to 0 */
+	aie4_fw_clear_alive(xdna);
```
maxzhen, line 534: Why do we need to perform this clear?

xdavidz:
Hi @hlaccabu , I set a do-no-merge label for now, because I am working on 3 questions that I need a clear thought
1) the request_firmware is cached, that means if the fw is updated but with same name, the linux doesn't refresh the fw memory. I am working on understanding how this happens and solutions;
2) the release_firmware is missing all around the code, which is stongly recommend by linux kernel document. I need to fix this along with suspend/resume flow.
3) I need to check how winodws driver handle this, especially if there is missing steps in smu or psp that linux driver is missing.

All 3 questions are my major concern. The code above seems "fix" the issue, but setting devm memory to NULL and each time use a new request firmware buffer do not explain the 3 above questions. I don't understand full, so please give me sometime to digest. I will maintain this PR with my additional tests and thoughts.

src/driver/amdxdna/aie4_pci.c
hlaccabu: I added this to match windows flow when uninstalling driver, I do not think it's necessary but wanted to match their flow... Ok with me if @xdavidz wants to remove this.

maxzhen: Can we ask why they need to do it in Win?

src/driver/amdxdna/aie4_psp.c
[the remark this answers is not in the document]
hlaccabu: That's true, I set to null to make sure those pointers are clear but it shouldn't be necessary. The main fix is being able to call release_firmware()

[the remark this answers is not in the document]
maxzhen: > That's true, I set to null to make sure those pointers are clear but it shouldn't be necessary. The main fix is being able to call release_firmware()

makes sense. psp does require releasing the fw memory before you can download a new one.

hlaccabu:
> Hi @hlaccabu , I set a do-no-merge label for now, because I am working on 3 questions that I need a clear thought
> 
> 2. the release_firmware is missing all around the code, which is stongly recommend by linux kernel document. I need to fix this along with suspend/resume flow.

For number 2, why do we need it elsewhere? We only want to call it in the 2 places it's in correct? 
1. When something goes wrong in aie4_request_firmware()
2. When uninstalling driver

src/driver/amdxdna/aie4_pci.c
hlaccabu: Sure! I will ask

hlaccabu:
@xdavidz I removed some of the unnecessary code that Max and Nishad pointed out and tested, still works as intended. You can maintain this PR from here as desired.

src/driver/amdxdna/aie4_pci.c
hlaccabu: Ok we can remove this, I got a tad confused looking through windows flow I guess. Only called when doing FLR like our current implementation.

hlaccabu: @xdavidz we can remove this

xdavidz:
> > Hi @hlaccabu , I set a do-no-merge label for now, because I am working on 3 questions that I need a clear thought
> > 
> > 2. the release_firmware is missing all around the code, which is stongly recommend by linux kernel document. I need to fix this along with suspend/resume flow.
> 
> For number 2, why do we need it elsewhere? We only want to call it in the 2 places it's in correct?
> 
> 1. When something goes wrong in aie4_request_firmware()
> 2. When uninstalling driver

I think sometimes AI doesn't give the right way.
Here is the linux document, we should always release the firmware right after using it.
https://docs.kernel.org/driver-api/firmware/request_firmware.html

This is also on par with upstream version.

xdavidz:
> Hi @hlaccabu , I set a do-no-merge label for now, because I am working on 3 questions that I need a clear thought
> 
> 1. the request_firmware is cached, that means if the fw is updated but with same name, the linux doesn't refresh the fw memory. I am working on understanding how this happens and solutions;
> 2. the release_firmware is missing all around the code, which is stongly recommend by linux kernel document. I need to fix this along with suspend/resume flow.
> 3. I need to check how winodws driver handle this, especially if there is missing steps in smu or psp that linux driver is missing.
> 
> All 3 questions are my major concern. The code above seems "fix" the issue, but setting devm memory to NULL and each time use a new request firmware buffer do not explain the 3 above questions. I don't understand full, so please give me sometime to digest. I will maintain this PR with my additional tests and thoughts.

Ok, I figured out, the No.1 is not true. As long as release is called, refcount goes to 0, we should be fine, This fixes most of the case.
We only need to do this request_firmware use it and release_firmware right a way.

Windows driver doesn't do anything special than this.

I don't know how to apply my changes on top of yours, so I am going to close this PR and re-create one. Will add you guys as reviewer.

hlaccabu:
> I think sometimes AI doesn't give the right way. Here is the linux document, we should always release the firmware right after using it. https://docs.kernel.org/driver-api/firmware/request_firmware.html
> 
> This is also on par with upstream version.

I see, I understand aie2 does this but we we're thinking it may be bad to have to re-request firmware every time a suspend/resume is called.

PR #1032, opened 2026-01-27
ve2: Set error_index properly for command chain type

rajkumar-xilinx:
Add error_index tracking in ve2_hwctx.c to identify which subcommand failed in a command chain.
In timeout case Set error_index to 0, and in Error case traverse chain slots to find failing command and set its index.
It enables XRT core to properly handle subcommand failures in command chains.

PR #1036, opened 2026-01-28
ve2: Set error_index properly for command chain type (#1032)

rajkumar-xilinx:
Add error_index tracking in ve2_hwctx.c to identify which subcommand failed in a command chain.
In timeout case Set error_index to 0, and in Error case traverse chain slots to find failing command and set its index.
It enables XRT core to properly handle subcommand failures in command chains.

(cherry picked from commit feb8b8778d6d657f56ef26f0723388e377ea8e7d)


rajkumar-xilinx:
All hw tests passed.

PR #1037, opened 2026-01-28
Implement MPNPU work buffer attach/detach routine

NishadSaraf:
Implement MPNPU work buffer attach/detach routine.

### DIFF
--- a/src/driver/amdxdna/aie4_message.c
+++ b/src/driver/amdxdna/aie4_message.c
@@ -149,12 +149,6 @@ int aie4_check_firmware_version(struct amdxdna_dev_hdl *ndev)
 	xdna->fw_ver.sub = resp.fw_patch;
 	xdna->fw_ver.build = resp.fw_build;
 
-	//if (resp.protocol_major != ndev->version.major) {
-		//XDNA_ERR(xdna, "Incompatible firmware protocol version major %d minor %d",
-			 //resp.protocol_major, resp.protocol_minor);
-		//return -EINVAL;
-	//}
-
 	XDNA_DBG(xdna, "FW version: %d.%d.%d.%d", xdna->fw_ver.major,
 		 xdna->fw_ver.minor, xdna->fw_ver.sub, xdna->fw_ver.build);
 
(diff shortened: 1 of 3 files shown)


maxzhen (dismissed):

src/driver/amdxdna/aie4_message.c
```diff
@@ -149,12 +149,6 @@ int aie4_check_firmware_version(struct amdxdna_dev_hdl *ndev)
 	xdna->fw_ver.sub = resp.fw_patch;
 	xdna->fw_ver.build = resp.fw_build;
 
-	//if (resp.protocol_major != ndev->version.major) {
```
xdavidz, line 152: I will add some checking later for major minor compatible operation when approach to production. delete this code is fine.

src/driver/amdxdna/aie4_msg_priv.h
```diff
@@ -198,11 +198,11 @@ enum aie4_msg_opcode {
...
 	AIE4_MSG_OP_CALIBRATE_CLOCK                  = 0x10009,
-	AIE4_MSG_OP_START_EVENT_TRACE                = 0x1000a,
-	AIE4_MSG_OP_STOP_EVENT_TRACE                 = 0x1000b,
-	AIE4_MSG_OP_SET_EVENT_TRACE_CATEGORIES       = 0x1000c,
-	AIE4_MSG_OP_DRAM_WORK_BUFFER                 = 0x1000d,
-	AIE4_MSG_OP_RELEASE_DRAM_WORK_BUFFER         = 0x1000e,
+	AIE4_MSG_OP_START_EVENT_TRACE                = 0x1000A,
+	AIE4_MSG_OP_STOP_EVENT_TRACE                 = 0x1000B,
+	AIE4_MSG_OP_SET_EVENT_TRACE_CATEGORIES       = 0x1000C,
+	AIE4_MSG_OP_ATTACH_WORK_BUFFER               = 0x1000D,
```
xdavidz: I do not recommend change the name of this, because we use a script to sync with the published header file with fw and windows.
changing name will increase the difficulty to sync for long run.

Is there a specific reason you want to change the naming?

src/driver/amdxdna/aie4_msg_priv.h
```diff
@@ -1754,55 +1754,53 @@ struct aie4_msg_get_runtime_cfg_resp {
...
+ * AIE4_MSG_OP_DETACH_WORK_BUFFER
  * Release DRAM work buffer response
  *
- * @status: enum aie4_msg_status
+ * @status: enum npu_msg_status
  */
-struct aie4_msg_release_dram_work_buffer_resp {
-	enum aie4_msg_status status;
+struct aie4_detach_work_buffer_resp {
+	u32 status;
```
xdavidz: we should use enum for this case, I think linux upstream code would prefer enum than u32?

src/driver/amdxdna/aie4_pci.c
```diff
@@ -682,10 +705,32 @@ static int aie4_pcidev_init(struct amdxdna_dev_hdl *ndev)
 	if (ret)
 		goto release_fw;
 
+	if (!is_npu3_vf_dev(pdev) && !skip_work_buffer) {
```
xdavidz, line 740: can this block of code become a single function, just like the others in this function?
like: aie4_work_buffer_init/fini?

NishadSaraf: Sure.

src/driver/amdxdna/aie4_msg_priv.h
NishadSaraf: Good catch. I'll fix this

src/driver/amdxdna/aie4_msg_priv.h
NishadSaraf: From the driver's point of view, attach/detach made more sense than the one defined in the published header file.

For the AIE2 generation, we have cherry-picked only the relevant Linux driver-supported interface definitions. This avoids bringing in unused opcode here.

Can you please share the script you are using to automate this?

src/driver/amdxdna/aie4_pci.c
```diff
@@ -682,10 +737,16 @@ static int aie4_pcidev_init(struct amdxdna_dev_hdl *ndev)
...
```
xdavidz, line 740: I'd think this line if (!is_npu3_vf_dev(pdev) && !skip_work_buffer) can be moved to the aie4_alloc_work_buffer and return 0 to continue. But it is not a must, we can sort the code later when upstreaming.

PR #1047, opened 2026-02-03
[ve2] Auto-select CMA region based on DT topology

bisingha-xilinx:
Implement automatic CMA memory region selection to eliminate manual mem_index specification. Users now only need to provide start_col and based on that while creating hw context, it can have a dedicated cma region associated with the hw context if DT has the topology. So for that particular hw context all the internal external buffers can be allocated to that correct dedicated cma region

### Tested:
1. Tested the provided special usecase in which the 36 aie cols are divided into 3 parts. 0-23 connected to only cma region0, 24-31 connected to only cma region1 and 32-35 connected to only cma region2. Tested by having multiple combinations.
2. Tested existing XRT hw unit tests for ve2 with the existing platform. All the tests are passing.

### DIFF
--- a/src/driver/amdxdna/amdxdna_ctx.c
+++ b/src/driver/amdxdna/amdxdna_ctx.c
@@ -138,6 +138,12 @@ int amdxdna_drm_create_hwctx_ioctl(struct drm_device *dev, void *data, struct dr
 		goto free_name;
 	}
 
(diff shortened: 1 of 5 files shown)


bisingha-xilinx:
Closing this PR as the fixes will go in main also. Once thats done I'll port the fixes in vai_6.1

PR #1056, opened 2026-02-06
Fix missing ctx_srcu protection in aie2_get_coredump()

NishadSaraf:
Hold the ctx_srcu read lock for the entire duration of hwctx usage.


NishadSaraf:
retest this please

PR #1061, opened 2026-02-09
Enable telemetry for aie4

hlaccabu:


src/shim/device.cpp
```diff
@@ -1111,16 +1147,193 @@ struct telemetry
...
+      return telemetry->query_misc_telemetry(pci_dev_impl);
+
+    case key_type::opcode_telemetry:
+      return telemetry->query_opcode_telemetry(pci_dev_impl);
+
+    case key_type::rtos_telemetry: {
+      bool is_aie4 = (device_id == NPU3_DEVICE_ID || device_id == NPU3_PF_DEVICE_ID ||
+                      device_id == NPU3_DEVICE_ID1 || device_id == NPU3_PF_DEVICE_ID1);
+      if (!is_aie4 && device_id != NPU4_DEVICE_ID) {
+        return query::rtos_telemetry::result_type{};
```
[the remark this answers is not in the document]
hlaccabu: This is incorrect

src/driver/amdxdna/aie4_message.c
```diff
@@ -292,9 +292,11 @@ int aie4_query_aie_telemetry(struct amdxdna_dev_hdl *ndev, u32 type, dma_addr_t
 		return -EINVAL;
 	}
 
+	req.type = (type == 0) ? AIE4_TELEMETRY_TYPE_PERF_COUNTER : type;
 	req.buf_addr = addr;
+	req.pasid.raw = 0;
```
NishadSaraf: Need to set the PASID here. Pass the value captured in `struct amdxdna_client`.

src/driver/amdxdna/aie4_pci.c
```diff
@@ -1697,9 +1698,10 @@ static int aie4_query_telemetry(struct amdxdna_client *client,
 				struct amdxdna_drm_get_info *args)
 {
 	struct amdxdna_dev *xdna = client->xdna;
+	struct amdxdna_mgmt_dma_hdl *dma_hdl;
 	struct amdxdna_dev_hdl *ndev;
 	dma_addr_t dma_addr;
-	size_t aligned_sz;
+	size_t aligned_sz, size;
```
NishadSaraf: nit. reverse x-mas

src/driver/amdxdna/aie4_pci.c
```diff
@@ -1714,28 +1716,41 @@ static int aie4_query_telemetry(struct amdxdna_client *client,
...
 	}
 
+	/* AIE4 firmware requires minimum 128KB buffer for telemetry */
 	ndev = xdna->dev_handle;
-	aligned_sz = PAGE_ALIGN(args->buffer_size);
-	buff = dma_alloc_noncoherent(xdna->ddev.dev, aligned_sz, &dma_addr,
-				     DMA_FROM_DEVICE, GFP_KERNEL);
-	if (!buff)
-		return -ENOMEM;
+	size = max_t(size_t, args->buffer_size, SZ_128K);
```
NishadSaraf: nit. Better to maintain a macro for the minimum buffer size required by the firmware in `msg_priv.h`

src/driver/amdxdna/aie4_pci.c
```diff
@@ -1714,28 +1716,41 @@ static int aie4_query_telemetry(struct amdxdna_client *client,
...
+
+	if (args->buffer_size < SZ_128K)
```
NishadSaraf: IMO, it is better to fail here than to return possibly truncated data to userspace. Also, in the failure case, the driver should return the expected buffer size in `args->buffer_size` to the userspace.

src/driver/amdxdna/aie4_pci.c
```diff
@@ -1714,28 +1716,41 @@ static int aie4_query_telemetry(struct amdxdna_client *client,
...
+
+	buff = amdxdna_mgmt_buff_get_cpu_addr(dma_hdl, 0);
+	if (IS_ERR(buff)) {
+		XDNA_ERR(xdna, "Failed to get CPU address for telemetry buffer");
+		ret = PTR_ERR(buff);
+		goto free_buf;
+	}
+
+	dma_addr = amdxdna_mgmt_buff_get_dma_addr(dma_hdl);
+	aligned_sz = dma_hdl->aligned_size;
```
NishadSaraf: The aligned size calculation is implemented to work around a known firmware issue. This is internal to the mgmt buffer allocation API and shouldn't be accessed here. Here `size` value is what needs to be passed to the firmware.

src/driver/amdxdna/aie4_pci.c
```diff
@@ -1714,28 +1716,41 @@ static int aie4_query_telemetry(struct amdxdna_client *client,
...
 
 	memset(buff, 0, aligned_sz);
-	drm_clflush_virt_range(buff, aligned_sz); /* device can access */
-	ret = aie4_query_aie_telemetry(ndev, type, dma_addr + sizeof(u64), aligned_sz);
+	amdxdna_mgmt_buff_clflush(dma_hdl, 0, 0);
+	ret = aie4_query_aie_telemetry(ndev, type, dma_addr, aligned_sz);
 	if (ret) {
 		XDNA_ERR(xdna, "Get telemetry failed ret %d", ret);
 		goto free_buf;
 	}
```
NishadSaraf, line 1748: Maybe be a good idea to add `print_hex_dump_debug()` here.

src/shim/device.cpp
```diff
@@ -963,8 +963,60 @@ struct telemetry
...
+  static constexpr uint16_t NPU3_DEVICE_ID = 0x17f1;
+  static constexpr uint16_t NPU3_PF_DEVICE_ID = 0x17f2;
+  static constexpr uint16_t NPU3_DEVICE_ID1 = 0x1b0a;
+  static constexpr uint16_t NPU3_PF_DEVICE_ID1 = 0x1b0b;
   static constexpr uint16_t NPU4_DEVICE_ID = 0x17f0;
 
+  static bool
+  is_aie4(uint16_t device_id)
+  {
+    return (device_id == NPU3_DEVICE_ID || device_id == NPU3_PF_DEVICE_ID ||
```
NishadSaraf: Better to have an array to maintain a list of device IDs of all NPU3 variants. What do you think?

src/shim/device.cpp
```diff
@@ -1111,16 +1154,187 @@ struct telemetry
...
+    
+    return std::make_unique<telemetry_handler>();
+  }
+
+  using result_type = std::any;
+
+  static result_type
+  get(const xrt_core::device* device, key_type key)
+  {
+    auto device_id = sysfs_fcn<uint16_t>::get(get_pcidev(device), "", "device");
```
NishadSaraf: For consistency, please use `xrt_core::device_query<xrt_core::query::pcie_id>(device)`

src/shim/device.cpp
```diff
@@ -1111,16 +1154,187 @@ struct telemetry
...
+    case key_type::opcode_telemetry:
+      return telemetry->query_opcode_telemetry(pci_dev_impl);
+
+    case key_type::rtos_telemetry: {
+      if (!is_aie4(device_id) && device_id != NPU4_DEVICE_ID) {
+        return query::rtos_telemetry::result_type{};
+      }
+      return telemetry->query_rtos_telemetry(pci_dev_impl);
+    }
+
```
NishadSaraf: nit. trailing whitespaces. fix here and a few other locations in this file.

src/driver/amdxdna/aie4_message.c
```diff
@@ -292,9 +292,11 @@ int aie4_query_aie_telemetry(struct amdxdna_dev_hdl *ndev, u32 type, dma_addr_t
 		return -EINVAL;
 	}
 
+	req.type = (type == 0) ? AIE4_TELEMETRY_TYPE_PERF_COUNTER : type;
```
NishadSaraf: This mapping looks hacky. The type of telemetry data must be set in the IOCTL request.

src/shim/device.cpp
```diff
@@ -1111,16 +1154,187 @@ struct telemetry
...
+      return telemetry->query_aie_telemetry(pci_dev_impl);
+
+    case key_type::misc_telemetry:
+      return telemetry->query_misc_telemetry(pci_dev_impl);
+
+    case key_type::opcode_telemetry:
+      return telemetry->query_opcode_telemetry(pci_dev_impl);
+
+    case key_type::rtos_telemetry: {
+      if (!is_aie4(device_id) && device_id != NPU4_DEVICE_ID) {
```
NishadSaraf: This check seems to beat the purpose of having separate device handlers. Maybe better to handle this minor difference in the base class implementation?

NishadSaraf (changes requested):
Hi @hlaccabu, Thanks for the change. Added request for minor tweaks and improvements.

src/driver/amdxdna/aie4_pci.c
hlaccabu: But I am taking the max between user requested size and SZ_128K, so I am enforcing proper size if they request too little. Should I swap this out to enforce user space to request correct size and return when they don't request enough?

src/driver/amdxdna/aie4_pci.c
hlaccabu: Removed this statement as there is so much telemetry data from this msg response it floods the dmesg. Is there a better way to do this?

src/shim/device.cpp
hlaccabu: I can create a handler for non aie4 and npu4 devices to return this empty value on rtos case and the rest will have the base aie2 implementation

src/shim/device.cpp
hlaccabu: Agreed.

src/driver/amdxdna/aie4_pci.c
NishadSaraf: If the driver receives a buffer smaller than the size a future firmware version might return, this implementation will truncate the data. Instead, the driver can return `-ENOSPC`, so the next time the user space can return with the expected size in `args->buffer_size`. For this, you'll need to catch `-ENOSPC` in your shim code.

src/driver/amdxdna/aie4_pci.c
NishadSaraf: 1. You can always choose to print only the required amount of data, which may not be the whole 128K.
2. Debug prints aren't enabled by default, and when enabled, can be selectively enabled/disabled using dynamic debug control on the fly.

NishadSaraf (approved):

PR #1064, opened 2026-02-09
Add AIE tile read/write support for AIE4

NishadSaraf:
Implement AIE register and memory access for AIE4.

### DIFF
--- a/src/driver/amdxdna/aie2_pci.c
+++ b/src/driver/amdxdna/aie2_pci.c
@@ -1567,13 +1567,10 @@ static int aie2_aie_tile_read(struct amdxdna_client *client, struct amdxdna_drm_
 			goto unlock_srcu;
 		}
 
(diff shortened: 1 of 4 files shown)


maxzhen:
In general, there are quite some duplicated code b/w aie2 and aie4. Let's refactor the code later and try make them common. It will significantly reduce the amount of work for upstreaming later.

xdavidz (approved):

xdavidz:
> In general, there are quite some duplicated code b/w aie2 and aie4. Let's refactor the code later and try make them common. It will significantly reduce the amount of work for upstreaming later.

We should discuss how we proceed the upstream work for this asap. I have a PF driver patchset ready already.
To start with the VF/classic driver, those dup code is the most significant effort for upstreaming.

I now get smu and psp combined in upstream patch and assigned the backport work to Soham.
For others, we need to layout a plan.

PR #1072, opened 2026-02-12
re-org suspend resume based on s3/s4 definition

xdavidz:
Try to make each PR smaller. The buffers needs more sophisticated design and coding. Basically, all buffers should be allocated /freed once and live with the s3 ram suspend/resume and s4 hibernate disk involved suspend/resume. More PR to do the fine tuning.


donwalkarsoham:
Tested this change on device and driver seg fault issue due to use after free (duplicate free attempts) scenario related to [AIESW-24681](https://jira.xilinx.com/browse/AIESW-24681) have been fixed.

hlaccabu (approved):

PR #1078, opened 2026-02-16
ve2: change indirect pkt array to [uc][slot] and fix driver with this format

rajkumar-xilinx:
-The hsa_queue.h layout for hq_indirect_pkt is changed from [slot][uc] to [uc][slot]. -This preserves the overall BO layout (header, packets, completion) but improves per UC locality and cache locality in CERT/firmware access.

Driver updates:
* ve2_create_host_queue(): init hq_indirect_pkt as [uc][slot]
* submit_command_indirect(): use [uc][slot] for CPU/DMA, write exec payloads
* packet_dump(): dump indirect packets with new order


rajkumar-xilinx:
Verification is in progress, one test is failing
SOME TESTS FAILED:
 - multi_thread_single_hwctx/

PR #1091, opened 2026-02-19
Add o_cmds, r_cmds, and c_rounds for every xrt_test

hlaccabu:


test/xrt_test/xrt_test.cpp
```diff
@@ -28,9 +29,9 @@
 namespace {
 
 using arg_type = const std::vector<uint64_t>;
-unsigned c_rounds = 1;
+uint64_t c_rounds = 1;
 unsigned o_cmds = 1;
-unsigned r_cmds = 24;
+unsigned r_cmds = 1;
```
maxzhen, line 36: Maybe we can set default to 0. In this case, each test case can decide their own r_cmds. Some can't do runlist so will use 1. The nop one can use 24 to cover runlist test.

hlaccabu: I still included a runlist test (#7) that starts by saving original r_cmds value, changing r_cmds to 24, and switching r_cmds back to original value once test is complete. This will allow for testing max runlist commands with test 7 while keeping each other test intact

test/xrt_test/xrt_test.cpp
```diff
@@ -312,306 +310,177 @@ sync_bo_from_dev(xrt_bo& bo)
...
+      submitted++;
+    }
+  }
+  
+  auto end = std::chrono::high_resolution_clock::now();
+  auto duration_us = std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
+  std::cout << "Executed total " << c_rounds << " runlists in " << duration_us
+	    << "us with max " << o_cmds << " outstanding runlist (" << r_cmds
+	    << " commands per runlist), average latency: "
+	    << duration_us * 1.0 / (r_cmds * c_rounds) << "us\n";
```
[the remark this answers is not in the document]
hlaccabu: That's correct, statement comes from runlist test

test/xrt_test/xrt_test.cpp
maxzhen: not sure i like this idea where you do not honor the r_cmds if specified by user.

hlaccabu: Every other tests honors it, but the reason for runlist test in this context is to test the max runlist commands. I would remove the runlist test altogether, but it is best if we have a test case number to execute this in stress test in other repo. I can rename test to max runlist

test/xrt_test/xrt_test.cpp
```diff
@@ -633,232 +502,88 @@ TEST_xrt_umq_single_col_resnet50_all_layer(int device_index, arg_type& arg)
...
-    elfpath.empty() ? local_path(path + "nop/nop.elf") : elfpath);
+  auto device = xrt::device{device_index};
+  auto elf = xrt::elf(elfpath.empty() ? local_path(path + "nop/nop.elf") : elfpath);
   xrt::hw_context hwctx{device, elf};
   xrt::kernel kernel = xrt::ext::kernel{hwctx, "DPU:dpu"};
-
-  // Create all runs
-  auto num_runs_per_batch = r_cmds * o_cmds;
-  for (int i = 0; i < num_runs_per_batch; i++)
+  unsigned n = o_cmds * r_cmds;
```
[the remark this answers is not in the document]
hlaccabu: It would be logically dead code

test/xrt_test/xrt_test.cpp
hlaccabu: Renamed the test + had execute batches take in a parameter for r_cmds instead of using global variable, that way r_cmds remains unchanged if running runlist test in parallel with other test

maxzhen: I'm OK if you keep a runlist test case which does not look at global r_cmds and uses 24 internally.

maxzhen:
Can I get a sample output from a default run?

maxzhen (approved):

test/xrt_test/xrt_test.cpp
```diff
@@ -1015,6 +755,10 @@ main(int argc, char **argv)
         }
         case 'r': {
           val = std::stoi(optarg);
+          if (val > 24) {
```
xdavidz: can this magic 24 be a valuable? like max_runlist or max_r_rounds?

xdavidz (approved):

test/xrt_test/xrt_test.cpp
hlaccabu: Yes, will do

PR #1092, opened 2026-02-19
Add new npu devices, rev IDs, and create npu3_family

hlaccabu:
Rev ID 11 uses different SMU registers


src/driver/amdxdna/npu3_regs.c
```diff
@@ -64,6 +67,33 @@ const struct amdxdna_dev_priv npu3_dev_priv = {
...
+	.psp_regs_off   = {
+		DEFINE_BAR_OFFSET(PSP_CMD_REG,    NPU3_PSP, MPASP_C2PMSG_123_ALT_1),
+		DEFINE_BAR_OFFSET(PSP_ARG0_REG,   NPU3_PSP, MPASP_C2PMSG_156_ALT_1),
+		DEFINE_BAR_OFFSET(PSP_ARG1_REG,   NPU3_PSP, MPASP_C2PMSG_157_ALT_1),
+		DEFINE_BAR_OFFSET(PSP_ARG2_REG,   NPU3_PSP, MPASP_C2PMSG_123_ALT_1),
+		DEFINE_BAR_OFFSET(PSP_INTR_REG,   NPU3_PSP, MPASP_C2PMSG_73_ALT_1),
+		DEFINE_BAR_OFFSET(PSP_STATUS_REG, NPU3_PSP, MPASP_C2PMSG_123_ALT_1),
+		DEFINE_BAR_OFFSET(PSP_RESP_REG,   NPU3_PSP, MPASP_C2PMSG_156_ALT_1),
+	},
+	.smu_regs_off   = {
```
xdavidz: Please check with @maxzhen if we need to bump npu3 to npu6 or npu7 because this is a successor of different hardware that use different set of SMU.

Seems PSP still the same.

xdavidz (changes requested):

src/driver/amdxdna/npu10_regs.c
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ * Copyright (C) 2026, Advanced Micro Devices, Inc.
+ */
+
+#include "npu3_family.h"
+
+const struct amdxdna_dev_info dev_npu10_info = {
+	.default_vbnv		= "RyzenAI-npu10",
```
xdavidz, line 9: Looks like only 7 has special smu definition and 8,9,10 are the same as 3. Can we use the npu3 name for 8,9,10 for now to save some npu numbers?

Thus, all others are sharing npu3 as vbnv name.
7 has it's own SMU. 3 and 7 are sharing npu3_family.h?

maxzhen: Either way works for me. We can add just 7 for now as @xdavidz suggested. I'm asking xrt-smi team to maintain and show platform names by themselves instead of retrieving from vbnv from driver. The name has to be the same across Win and Lin anyway. So, our driver's vbnv may not be relevant in the future and can be optionally removed.

PR #1101, opened 2026-02-20
amdxdna: add FLR support for aie2 devices

FIM43-Redeye:
## Summary

- Implement `reset_prepare` and `reset_done` callbacks for aie2 (Phoenix/Hawk Point NPU)
- These were previously only implemented for aie4 (Strix Point), leaving aie2 unable to handle PCIe Function Level Reset
- The device advertises `FLReset+` in PCIe DevCap and exposes `flr` as its reset method, but the driver left the aie2_ops callbacks NULL

Without these callbacks, sysfs reset (`echo 1 > /sys/bus/pci/devices/.../reset`), the PCI error recovery handler, and TDR escalation (follow-up patch) cannot perform a hardware-level reset.

`reset_prepare` tears down the firmware stack (TDR, contexts, PM, firmware, mailbox, PSP, SMU). `reset_done` rebuilds it in reverse order. Locking follows the established `dev_lock -> aie2_lock` hierarchy.

## Test plan

- Verified device advertises `FLReset+` via `lspci -vvv` (DevCap register)
- Built and loaded on Phoenix (NPU1, device 0x1502) with kernel 6.19.0
- checkpatch: 0 errors, 0 warnings

### DIFF
--- a/src/driver/amdxdna/aie2_pci.c
+++ b/src/driver/amdxdna/aie2_pci.c
@@ -489,6 +489,149 @@ static int aie2_hw_resume(struct amdxdna_dev *xdna)
 	return 0;
 }
 
+static void aie2_reset_prepare(struct amdxdna_dev *xdna)
+{
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+
+	XDNA_INFO(xdna, "FLR reset prepare start");
+
+	/*
+	 * Stop TDR before taking dev_lock. cancel_work_sync() waits
+	 * for any in-progress TDR work, which also takes dev_lock.
+	 */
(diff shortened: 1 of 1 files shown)


AMDGithubSCIMAdmin:

maxzhen (commented):
We can't support FLR on aie2 devices due to hardware issues. It just does not work reliably and will make things even worse.

FIM43-Redeye:
Coming back to this to say thanks for explaining! I'm working on a cycle-accurate emulator for XDNA, so I've been riding my AIE2 really hard and it's been wedging in weird ways at times. It helps a lot to understand constraints like these, thanks for taking the time.

Issue #1083, opened 2026-02-18
How does a Kernel determine, if it made progress?


ilot95:
I want to run Kernel which moves a lot of data, aka. runs long.
I was getting timeouts on wait() even tough no timeout is set. 
In dmesg I saw: "amdxdna 0000:0c:00.1: amdxdna_tdr_work: Device isn't making progress... Count 243"
By increasing the timeout in the Kernel module with
sudo rmmod amdxdna  
sudo modprobe amdxdna timeout_in_sec=20
I was able to run my Kernel. 
Data is moved with multiple DMA transfers in MLIR code.
How does a Kernel determine if it made progress?
I found: aie2_sched_notify which resets the timer, but I am not sure this method is called.

maxzhen:
Driver just use a timer and default timeout is 2 seconds.

PR #1102, opened 2026-02-20
amdxdna: add progressive TDR recovery escalation

FIM43-Redeye:
## Summary

- Add progressive recovery escalation to TDR: soft recovery -> firmware reset -> PCIe FLR -> device declared unrecoverable
- Factor firmware teardown/rebuild into shared helpers (`aie2_fw_teardown`/`aie2_fw_rebuild`) used by both the PCI error handler and TDR paths
- When the AIE array is stuck, soft recovery (context stop/restart) has no effect -- the hardware itself is jammed. TDR loops indefinitely (see #1083). This patch gives it teeth.

Depends on #1101 (FLR support for aie2).

### Escalation model

| Counter | Action | On success |
|---------|--------|------------|
| 1 | Soft recovery (stop/restart contexts) | Reset to 0 |
| 2 | Firmware reset (suspend/resume) | Reset to 0 |
| 3 | PCIe FLR (full hardware reset) | Reset to 0 |
| 4+ | Device declared unrecoverable, TDR stopped | Reboot required |

The counter also resets to 0 when the device resumes making progress on its own.

### Locking

Uses manual `mutex_lock`/`mutex_unlock` instead of `guard(mutex)` because the FLR path needs to release `dev_lock` around the PCI reset. The lock hierarchy (`dev_lock -> aie2_lock`) is preserved throughout.

## Test plan

- Built and loaded on Phoenix (NPU1, device 0x1502) with kernel 6.19.0
- TDR fires at count 1, soft recovery succeeds, counter resets -- verified via dmesg
- checkpatch: 0 errors, 0 warnings
- Escalation to firmware reset and FLR not yet triggered in testing (soft recovery has been sufficient), but the code paths exist as a safety net for truly stuck hardware

### DIFF
--- a/src/driver/amdxdna/aie2_pci.c
+++ b/src/driver/amdxdna/aie2_pci.c
@@ -489,6 +489,121 @@ static int aie2_hw_resume(struct amdxdna_dev *xdna)
 	return 0;
 }
 
+/*
+ * Tear down the firmware stack: PM, firmware, mailbox, PSP, SMU.
+ * Caller must hold aie2_lock. Sets dev_status to AIE2_DEV_INIT.
+ */
+static void aie2_fw_teardown(struct amdxdna_dev_hdl *ndev)
+{
+	ndev->dev_status = AIE2_DEV_INIT;
+	aie2_pm_fini(ndev);
+	aie2_mgmt_fw_fini(ndev);
+	xdna_mailbox_stop_channel(ndev->mgmt_chann);
+	xdna_mailbox_destroy_channel(ndev->mgmt_chann);
+	ndev->mgmt_chann = NULL;
+	if (ndev->mbox) {
+		xdna_mailbox_destroy(ndev->mbox);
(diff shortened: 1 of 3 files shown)


AMDGithubSCIMAdmin:

maxzhen:
In case a model run takes a very long time, you'll need to break it down to multiple inferencing command and make sure each one does not exceed 2 seconds. In the future, we could allow application to pass down a timeout value for better timeout handling.

Today, when driver detected timeout, the only thing driver can reliably recover is to destroy the context and re-create it. If that does not work, power cycle is required. Other mechanism does not work well.

FIM43-Redeye:
Coming back to this to say thanks for explaining! I'm working on a cycle-accurate emulator for XDNA, so I've been riding my AIE2 really hard and getting an unusual amount of TDRs and wedges. It helps a lot to understand constraints like these, thanks for taking the time.

PR #1104, opened 2026-02-21
Add support for generating "standalone" DKMS artifacts

superm1:
As distros package XRT natively, it can still be useful to easily download the "out-of-tree" kernel driver to validate something.

Add support for creating a DKMS source package target to the build system to support this.
Create artifacts of both the source and binary packages at every commit.
* Source packages are useful for building into binary packages on a build infrastructure
* Binary packages can be directly installable

On tagged releases upload them to [this PPA](https://launchpad.net/~amd-team/+archive/ubuntu/xrt).
In order to actually upload to the PPA, `PPA_GPG_PRIVATE_KEY` must be declared as a repository secret in order to sign before uploading.
We can talk offline about what GPG key should be used for this purpose.

### DIFF
--- a/.github/workflows/build-dkms-source.yml
+++ b/.github/workflows/build-dkms-source.yml
@@ -143,10 +143,10 @@ jobs:
       run: |
         # Get current version from changelog
         CURRENT_VERSION=$(dpkg-parsechangelog -S Version)
-        # Remove any existing release suffix (e.g., ~noble1, ~questing1)
-        BASE_VERSION=$(echo "$CURRENT_VERSION" | sed 's/~[a-z]*[0-9]*$//')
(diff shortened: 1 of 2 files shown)


AMDGithubSCIMAdmin:

superm1:
CC @jgmelber

superm1:
The reason for the initial CI build failure is that containers need to be created in the repo for the first time.  This can be triggered one of two ways:
1. Manual workflow dispatch in GH actions
2. Nightly

If you want to see CI pass now, you'll need the container merged first.  But here is a run showing it working in my fork:
https://github.com/superm1/xdna-driver/actions/runs/22258300169/job/64392479275

PR #1116, opened 2026-02-25
col_event -> cert_comp

maxzhen:
Change col_event to cert_comp since it is really just for cert completion notification.
Instead of using a linked list, use xarray to save cert_comp which is looked up by msix index.
Real timeout recovery will be in next PR.

### DIFF
--- a/src/driver/amdxdna/aie4_pci.c
+++ b/src/driver/amdxdna/aie4_pci.c
@@ -161,7 +161,7 @@ static void cert_timer(struct timer_list *t)
 #endif
 	struct amdxdna_dev *xdna = ndev->xdna;
 	struct cert_comp *cert_comp;
-	long unsigned int msix_idx;
+	unsigned long msix_idx;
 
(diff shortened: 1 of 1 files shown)


src/driver/amdxdna/aie4_error.c
```diff
@@ -264,8 +264,7 @@ static void aie4_ctx_disconnect(struct amdxdna_dev_hdl *ndev, u32 hw_ctx_id)
 		xa_for_each(&client->ctx_xa, ctx_id, ctx) {
 			if (ctx->priv && ctx->priv->hw_ctx_id == hw_ctx_id) {
 				ctx->priv->status = CTX_STATE_DISCONNECTED;
-				ctx->priv->col_entry->needs_reset = true;
```
xdavidz, line 267: I don't quite follow the logic why we remove the needs_reset, but seems fine with current login.

Can you please test both polling/intr mode if all works, including running FLR. Then I am good with this change. Seems clean other parts.

PR #1122, opened 2026-02-26
Re-schedule partition at context destroy

amd-borpanov:
When context yield is requested, but it gets deleted before the yield, it will not yield. So, re-schedule the corresponding partition before delete.


AMDGithubSCIMAdmin:

maxzhen:
ok to test

PR #1128, opened 2026-02-26
Support elf flow for BO tests

hlaccabu:


test/shim_test/io.cpp
```diff
@@ -224,6 +224,17 @@ elf_init_no_arg_cmd(xrt::elf& elf, cuidx_type idx, bool dump, bo& cmd, bo& inst)
 
 } // namespace
 
+std::unique_ptr<io_test_bo_set_base>
+create_bo_set_for_device(device* dev, bool use_ubuf, const char* tag)
+{
+  auto device_id = device_query<query::pcie_device>(dev);
+  if (device_id == npu3_device_id || device_id == npu3_device_id1) {
+    const char* aie4_tag = (tag && tag[0]) ? tag : "good";
+    return std::make_unique<elf_full_io_test_bo_set>(dev, aie4_tag);
```
maxzhen, line 237: you want to return error when use_ubuf is true since you don't actually support it and fail loudly.

hlaccabu: Ah good point, I realize test 59 creates the user bos inside shim test, but 60 relies on ddr_range so I disabled that for aie4 for now.

PR #1132, opened 2026-02-27
Re-schedule partition at context destroy (#1122)

amd-borpanov:
When context yield is requested, but it gets deleted before the yield, it will not yield. So, re-schedule the corresponding partition before delete.

(cherry picked from commit 970fa3f6988d758e5e1bca6d1c49f5a54d7e02ff)


AMDGithubSCIMAdmin:

PR #1135, opened 2026-02-28
Updated the VE2 driver to enable getting and setting AIE frequencies using the xrt-smi advanced --aie-clock -s/-g commands

ManojTakasi:
Updated the VE2 driver to enable getting and setting AIE frequencies using the xrt-smi advanced --aie-clock -s/-g commands.

The current command xrt-smi examine --advanced --report clocks -d 0 reports the AIE clock value but does not support modifying it. We require an API that allows setting the AIE frequency based on user-provided arguments (e.g., 600M, 1000M, 100000K). Since XRT already supports frequency control through xrt-smi advanced --aie-clock -s/-g, we have added the necessary structures for the corresponding queries and updated the driver implementation to handle frequency changes via the DRM_AMDXDNA_SET_CLOCK_FREQ IOCTL command.

### DIFF
--- a/src/driver/amdxdna/ve2_debug.c
+++ b/src/driver/amdxdna/ve2_debug.c
@@ -830,6 +830,111 @@ int ve2_get_array(struct amdxdna_client *client, struct amdxdna_drm_get_array *a
 	return ret;
 }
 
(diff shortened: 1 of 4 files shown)


chvamshi-xilinx (approved):

andrej:
Thank you for enabling this!

PR #1229, opened 2026-04-01
accel/amdxdna: DMA buffer helpers and AIE4 work buffer support

NishadSaraf:
This series introduces a common DMA buffer abstraction for the amdxdna
driver and uses it to add DRAM work buffer support for AIE4 devices.
The NPU firmware has strict constraints on host-allocated DMA buffers:
sizes must be power-of-2, addresses must be naturally aligned, and
buffers must not cross a 64 MB boundary. Currently, the AIE2 code
opens its own alloc/free helpers without enforcing these rules.

Patches 1-2 introduce struct aie_dma_hdl with alloc/free/clflush
helpers that enforce the firmware constraints, then migrate the
existing AIE2 call sites.

Patch 3 builds on this to allocate and attach a DRAM work buffer
required by AIE4 firmware for runtime operation. Only the PF driver
performs the allocation; VF devices skip it.

Patch 4 fixes a minor SR-IOV teardown issue where the driver
unconditionally sent a destroy-VFs command even when no VFs existed.

### DIFF
--- a/drivers/accel/amdxdna/aie.c
+++ b/drivers/accel/amdxdna/aie.c
@@ -4,6 +4,7 @@
  */
 
 #include <linux/errno.h>
+#include <linux/sizes.h>
 
 #include "aie.h"
 #include "amdxdna_mailbox_helper.h"
@@ -118,4 +119,38 @@ void amdxdna_vbnv_init(struct amdxdna_dev *xdna)
 	xdna->vbnv = amdxdna_lookup_vbnv(info->rev_vbnv_tbl, rev);
 	if (!xdna->vbnv)
 		xdna->vbnv = info->default_vbnv;
+}
+
+void *amdxdna_alloc_msg_buffer(struct amdxdna_dev *xdna, u32 *size,
+			       dma_addr_t *dma_addr)
+{
+	void *vaddr;
+	int order;
+
+	*size = max_t(u32, *size, SZ_8K);
+	order = get_order(*size);
+	if (order > MAX_PAGE_ORDER)
+		return ERR_PTR(-EINVAL);
+	*size = PAGE_SIZE << order;
+
+	if (amdxdna_iova_on(xdna))
+		return amdxdna_iommu_alloc(xdna, *size, dma_addr);
+
+	vaddr = dma_alloc_noncoherent(xdna->ddev.dev, *size, dma_addr,
+				      DMA_FROM_DEVICE, GFP_KERNEL);
+	if (!vaddr)
+		return ERR_PTR(-ENOMEM);
+
+	return vaddr;
+}
+
+void amdxdna_free_msg_buffer(struct amdxdna_dev *xdna, size_t size,
+			     void *cpu_addr, dma_addr_t dma_addr)
+{
+	if (amdxdna_iova_on(xdna)) {
+		amdxdna_iommu_free(xdna, size, cpu_addr, dma_addr);
+		return;
+	}
+
+	dma_free_noncoherent(xdna->ddev.dev, size, cpu_addr, dma_addr, DMA_FROM_DEVICE);
 }
--- a/drivers/accel/amdxdna/aie.h
+++ b/drivers/accel/amdxdna/aie.h
@@ -95,6 +95,11 @@ int aie_send_mgmt_msg_wait(struct aie_device *aie, struct xdna_mailbox_msg *msg)
 int aie_check_protocol(struct aie_device *aie, u32 fw_major, u32 fw_minor);
 void amdxdna_vbnv_init(struct amdxdna_dev *xdna);
 
+void *amdxdna_alloc_msg_buffer(struct amdxdna_dev *xdna, u32 *size,
+			       dma_addr_t *dma_addr);
+void amdxdna_free_msg_buffer(struct amdxdna_dev *xdna, size_t size,
+			     void *cpu_addr, dma_addr_t dma_addr);
+
 /* aie_psp.c */
 struct psp_device *aiem_psp_create(struct drm_device *ddev, struct psp_config *conf);
 int aie_psp_start(struct psp_device *psp);
--- a/drivers/accel/amdxdna/aie4_message.c
+++ b/drivers/accel/amdxdna/aie4_message.c
@@ -25,3 +25,23 @@ int aie4_suspend_fw(struct amdxdna_dev_hdl *ndev)
 
 	return ret;
 }
+
+int aie4_attach_work_buffer(struct amdxdna_dev_hdl *ndev, u32 pasid,
+			    dma_addr_t addr, u32 size)
+{
+	DECLARE_AIE_MSG(aie4_msg_attach_work_buffer, AIE4_MSG_OP_ATTACH_WORK_BUFFER);
+	struct amdxdna_dev *xdna = ndev->aie.xdna;
+	int ret;
+
+	req.buff_addr = addr;
+	req.buff_size = size;
+	req.pasid = FIELD_PREP(AIE4_MSG_PASID_MASK, pasid);
+	if (pasid)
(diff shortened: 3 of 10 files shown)


drivers/accel/amdxdna/aie.c
```diff
@@ -119,3 +126,124 @@ void amdxdna_vbnv_init(struct amdxdna_dev *xdna)
...
 }
+
+/*
+ * Validate DMA buffer constraints required by firmware:
+ * 1. Size must be in [8KB, 64MB]
+ * 2. Size must be a power of 2
+ * 3. Address must be aligned to buffer size
+ * 4. Buffer must not cross a 64MB boundary
+ */
+static int aie_dma_buf_validate(struct amdxdna_dev *xdna, struct aie_dma_hdl *hdl)
```
houlz0507: Why do not use aie2_alloc_msg_buffer()? Can you just move it here without re-implement?

NishadSaraf: Sure. Do we want to drop the OOT driver workaround along with it? The staging driver doesn't exactly implement the firmware workaround, capping the maximum allocated size to 4M.

drivers/accel/amdxdna/aie.h
```diff
@@ -88,13 +90,40 @@ struct amdxdna_rev_vbnv {
...
+/*
+ * struct aie_dma_hdl - DMA buffer handle for firmware communication
+ * @xdna: back-pointer to the device
+ * @dir: DMA data direction
+ * @vaddr: kernel virtual address
+ * @dma_addr: device DMA address
+ * @size: requested buffer size
+ * @aligned_size: actual allocation size (power-of-2 aligned for FW)
+ */
+struct aie_dma_hdl {
```
houlz0507: Why is aie_dma_hdl required?  Do we really need *xdna, dir, size here? Is it possible to just keep the original interface? And you may merge this patch with next one (aie2)

drivers/accel/amdxdna/aie4_msg_priv.h
```diff
@@ -46,4 +46,25 @@ struct aie4_msg_destroy_vfs_resp {
 	enum aie4_msg_status status;
 } __packed;
 
+union aie4_msg_pasid {
+	__u32 raw;
+	struct {
+		__u32 pasid     : 20;
+		__u32 rsvd      : 11;
+		__u32 pasid_vld : 1;
```
houlz0507: Please do not use bit define. Instead, use GENMASK, FIELD_PREP etc.

drivers/accel/amdxdna/aie4_msg_priv.h
```diff
@@ -46,4 +46,25 @@ struct aie4_msg_destroy_vfs_resp {
...
+	} f;
+} __packed;
+
+#define AIE4_WORK_BUFFER_MIN_SIZE	(4 * 1024 * 1024) /* 4 MB */
```
houlz0507: SZ_4M

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -234,9 +236,40 @@ static int aie4_fw_load(struct amdxdna_dev_hdl *ndev)
 	return ret;
 }
 
+static int aie4_alloc_work_buffer(struct amdxdna_dev_hdl *ndev)
+{
+	struct amdxdna_dev *xdna = ndev->aie.xdna;
+	struct pci_dev *pdev = to_pci_dev(xdna->ddev.dev);
+	struct aie_dma_hdl *buf;
+
+	if (pdev->is_virtfn)
```
houlz0507: vf is not added yet. Please remove this for now.

drivers/accel/amdxdna/aie4_message.c
```diff
@@ -25,3 +25,27 @@ int aie4_suspend_fw(struct amdxdna_dev_hdl *ndev)
...
 }
+
+int aie4_attach_work_buffer(struct amdxdna_dev_hdl *ndev, u32 pasid,
+			    dma_addr_t addr, u32 size)
+{
+	DECLARE_AIE_MSG(aie4_msg_attach_work_buffer, AIE4_MSG_OP_ATTACH_WORK_BUFFER);
+	struct amdxdna_dev *xdna = ndev->aie.xdna;
+	int ret;
+
+	if (size < AIE4_WORK_BUFFER_MIN_SIZE || !addr) {
```
houlz0507: you can remove the check.

drivers/accel/amdxdna/aie4_msg_priv.h
NishadSaraf: Ah. This was a miss. I'll fix it.

drivers/accel/amdxdna/aie.h
NishadSaraf: The upcoming debug features planned for upstreaming follow a similar pattern where buffers need to be retained across suspend and resume cycles. These buffers are allocated during initialization and freed only when the driver is removed. Different buffer attributes are accessed at various locations in the code base. As there is significant overlap in these patterns, this approach helps reduce redundancy and simplifies the maintenance across different generations. 

When freeing the buffer, both the `xdna` and `size` attributes are required when only the handle is provided. I will merge this information with the next patch in the series.

houlz0507: The change for upcoming patch is not allowed for upstreaming. Could you add this part to your debug features patches?

xdavidz:
@NishadSaraf  I thought that we could make the mgmt buffer common. Thus all buffers including work buffer should be allocated by the amdxdna APIs, maybe move code in aie.c but API name as amdxdna_xxx for buffer alloc and free.

The workbuffer alloc/free might be no-op for aie2. I discussed this with Max, we even bringstrom to have the workbuffer for aie2 even we don't use it today. Thus, we can make the code look simpler for the linux maintainer to review.
But for VE2 or embedded, this is really not needed, therefore this is questionable. I don't have a deeper thought further. Lizhi may have suggestions to make this area look nicer.

After that the extra ops is the attach or register (I think those different naming APIs are doing the same thing, just using different mailbox messages across aie2, aie4)

Given the current situation, I am not sure you plan. Are you going to make it done in next week? If yes, please arrange a meeting instead of chatting here for better communication. 

Please also include @hlaccabu, I asked him to get familiar with this area too.

drivers/accel/amdxdna/aie.h
NishadSaraf: Will do!

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -474,15 +521,23 @@ static int aie4_init(struct amdxdna_dev *xdna)
 	ndev->aie.xdna = xdna;
 	xdna->dev_handle = ndev;
 
+	ret = aie4_alloc_work_buffer(ndev);
```
houlz0507: Should be called after pcidev_init() which enable device and set dma mask.

xdavidz:
After reviewing your code. I think I should take both "move message buffer to common layer" and "add dram work buffer for AIE4". And the 3rd is already fixed in upcoming patches.

I don't see a reason why we have to get your current PR merged in 1-2 days. I will need to fully tested on top my upcoming patchsets, and also reviewed by Lizhi. I think it should be over all in good shape. but we don't plan it in 1-2 days. Just to be clear.

Your name will appear on both the staging driver PR and the linux upstreaming requests.

xdavidz:
@maxzhen I added these 2 PR into another PR here: https://github.com/amd/xdna-driver/pull/1264


could we use the new PR to review the changes as a full set of next upstreaming request, and close this one?

maxzhen:
> @maxzhen I added these 2 PR into another PR here: #1264
> 
> could we use the new PR to review the changes as a full set of next upstreaming request, and close this one?

Yes, I'll close this one.

PR #1259, opened 2026-04-15
amdxdna: Add OOT to module description to disinguish oot and staging …

hlaccabu:
…drivers

- Create shim_test dev filter for OOT vs staging driver


hlaccabu:
use -k option instead in shim_test

PR #1269, opened 2026-04-18
opt: add GCC optimize pragmas to amdxdna_ubuf.c

theonlychant:


maxzhen:
Thanks for the PR.
However, I don't see any in-tree Linux driver enabled these opts. It seems risky to me to enable them here. I decide not to take this change.

PR #1281, opened 2026-04-29
Fix building with clang 22.x

berolinux:
Without this patch, clang fails on shim_debug.h when a std::string (as opposed to a POD like char*) is passed through varargs. This probably causes a runtime crash with gcc, and definitely causes a compile time error with clang.

It also fails on virtio/platform_virtio.cpp because of the attempt to initialize a variable length array to nothing (which gcc seems to treat as a no-op).

### DIFF
--- a/src/shim/shim_debug.h
+++ b/src/shim/shim_debug.h
@@ -12,19 +12,29 @@
 
 namespace {
 
+template <typename T>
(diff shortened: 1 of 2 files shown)


maxzhen (approved):

maxzhen:
ok to test

PR #1285, opened 2026-05-02
Build using the configuration of the chosen kernel, not /proc/config.gz

AngryLoki:
This fixes compilation when the compilers for the running kernel and the target kernel differ.

`/proc/config.gz` should not be used as the primary source of configuration values for several reasons:
- A module may be compiled for a kernel that will be booted later
- `/proc/config.gz` may not belong to the target system (e.g., an Arch/Gentoo host with a Clang-built kernel building a deb inside an Ubuntu Docker container with a GCC-based config in `/lib/modules/6.8.0-111-generic/build/.config`)


maxzhen:
ok to test

maxzhen (approved):

PR #1286, opened 2026-05-04
accel/amdxdna: move get PMF data and query sensors to common code + AIE4 temp support

hlaccabu:


drivers/accel/amdxdna/aie.c
```diff
@@ -214,6 +214,79 @@ int amdxdna_get_metadata(struct aie_device *aie,
...
+
+	if (copy_to_user(u64_to_user_ptr(args->buffer), &sensor, sizeof(sensor)))
+		return -EFAULT;
+
+	args->buffer_size -= sizeof(sensor);
+	sensors_count++;
+
+	memset(&sensor, 0, sizeof(sensor));
+	sensor.type = AMDXDNA_SENSOR_TYPE_TEMPERATURE;
+	sensor.input = npu_metrics.npu_temp;
```
maxzhen: This is a new field, you need to wrap it by a new HAVE_xxx macro.

drivers/accel/amdxdna/aie.h
```diff
@@ -11,6 +11,34 @@
 #define AIE_INTERVAL	20000	/* us */
 #define AIE_TIMEOUT	1000000	/* us */
 
+#if IS_ENABLED(CONFIG_AMD_PMF) && defined(HAVE_7_0_amd_pmf_get_npu_data)
+#include <linux/amd-pmf-io.h>
+#define AIE_GET_PMF_NPU_METRICS(metrics) amd_pmf_get_npu_data(metrics)
+#define AIE_GET_PMF_NPU_DATA(field, val)				\
```
maxzhen: I don't see it is ever called?

maxzhen (changes requested):

drivers/accel/amdxdna/aie.c
```diff
@@ -214,6 +214,81 @@ int amdxdna_get_metadata(struct aie_device *aie,
...
+
+	args->buffer_size -= sizeof(sensor);
+	sensors_count++;
+
+#ifdef HAVE_7_2_amd_pmf_npu_metrics_npu_temp
+	memset(&sensor, 0, sizeof(sensor));
+	sensor.type = AMDXDNA_SENSOR_TYPE_TEMPERATURE;
+	sensor.input = npu_metrics.npu_temp;
+	sensor.unitm = 0;
+	scnprintf(sensor.label, sizeof(sensor.label), "Total Temperature");
```
maxzhen: "Total Temperature" sounds strange. Is this really something that xrt-smi is looking for?

maxzhen (dismissed):

drivers/accel/amdxdna/aie.c
hlaccabu: Good point, I think I got that from "Total Power" from xrt-smi but for temp it is just reported as "Temperature (C)". I will fix

drivers/accel/amdxdna/aie.c
```diff
@@ -214,6 +214,87 @@ int amdxdna_get_metadata(struct aie_device *aie,
...
+			  struct amdxdna_drm_get_info *args, u32 total_col)
+{
+#if IS_ENABLED(CONFIG_AMD_PMF) && defined(HAVE_7_0_amd_pmf_get_npu_data)
+	struct amdxdna_drm_query_sensor sensor = {};
+	struct amd_pmf_npu_metrics npu_metrics;
+	u32 sensors_count = 0, i;
+	int ret;
+
+#ifdef HAVE_7_2_amd_pmf_npu_metrics_npu_temp
+	npu_metrics.npu_temp = U16_MAX;
```
houlz0507, line 228: If PMF would not update the unsupported counters, maybe we should do the same for other counters as well? e.g. memset npu_metrics to 0xff?

hlaccabu: We only return a total of 3 of the counters, and of those 3 temp is the only one supported by one device series (AIE4) and not the other, hence why we set it to U16_MAX. The other unsupported npu_metrics we don't even return to user... stylistically, is it better to leave them unset or set to 0xff?

houlz0507: > We only return a total of 3 of the counters, and of those 3 temp is the only one supported by one device series (AIE4) and not the other, hence why we set it to U16_MAX. The other unsupported npu_metrics we don't even return to user... stylistically, is it better to leave them unset or set to 0xff?

My questions is the unsupported counters. Will PMF set it to an invalid value like U16_MAX? Or it will not change it? If PMF set it to invalid value, then the driver does not initial it with U16_MAX. If PMF does not change the unsupported counter, then the driver should initialize all the counters.

hlaccabu: I init here now: 
struct amd_pmf_npu_metrics npu_metrics = {};

Max and I discussed, values of 0 work for the rest of the fields, we just want to set npu_temp to U16_MAX because 0 is a valid temperature.

hlaccabu:
retest this please

PR #1290, opened 2026-05-05
accel/amdxdna: add AIE tile register/memory read/write access

NishadSaraf:
Add a secure debug interface that lets userspace read or write
individual AIE tile registers (size == 4) and tile memory blocks
(size > 4) via firmware mailbox messages. Reads are exposed through
DRM_AMDXDNA_AIE_TILE_READ on get_array; writes through
DRM_AMDXDNA_AIE_TILE_WRITE on set_state. Access is gated by EUID
match against the context creator (or CAP_SYS_ADMIN), replacing the
previous PID-based check.

Patches:
  1. accel/amdxdna: backport upstream coredump cleanup
     - aligns the existing AIE coredump path with the upstream-bound
       version: direct callback in place of dev_ops indirection,
       inlined CAP_SYS_ADMIN gate, goto-style error paths, kerneldoc
       reformat.
  2. accel/amdxdna: add AIE tile read/write access
     - core tile r/w implementation plus a shared
       amdxdna_hwctx_lookup_locked() helper (also retrofitted into
       the coredump path), AIE2 enablement, UAPI, version bump 0.12.
  3. accel/amdxdna: enable tile read/write for AIE4 devices
     - AIE4 firmware-message + dispatch wiring with PASID forwarding.
  4. test/shim_test: fix tile read/write test drv_ver to always pass
     - lets the new tile r/w shim tests run as positive tests.

### DIFF
--- a/drivers/accel/amdxdna/aie.c
+++ b/drivers/accel/amdxdna/aie.c
@@ -311,17 +311,6 @@ void amdxdna_hmm_invalidate(struct amdxdna_gem_obj *abo,
 		XDNA_DBG(xdna, "Wait for bo interrupted by signal");
 }
 
-bool amdxdna_hwctx_access_allowed(struct amdxdna_hwctx *hwctx, bool root_only)
-{
-	if (amdxdna_is_admin())
-		return true;
-
-	if (root_only)
-		return false;
-
-	return task_tgid_nr(current) == hwctx->client->pid;
-}
-
 struct amdxdna_msg_buf_hdl *amdxdna_alloc_msg_buff(struct amdxdna_dev *xdna, u32 size)
 {
 	struct amdxdna_msg_buf_hdl *hdl;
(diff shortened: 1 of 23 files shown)


drivers/accel/amdxdna/aie.c
```diff
@@ -339,36 +375,9 @@ int amdxdna_get_coredump(struct aie_device *aie,
...
-	}
-
-	if (!capable(CAP_SYS_ADMIN) &&
-	    task_tgid_nr(current) != hwctx->client->pid) {
-		XDNA_ERR(xdna, "Permission denied for context %u",
-			 config.context_id);
-		ret = -EPERM;
-		goto unlock_srcu;
-	}
+	hwctx = amdxdna_hwctx_lookup_locked(xdna, config.pid, config.context_id, &idx);
```
houlz0507: It is very confuse for taking a lock in the function and unlock outside.  Please consider  amdxdna_hwctx_walk

amd-sgannu:
retest this please

amd-sgannu:
retest this please

drivers/accel/amdxdna/aie2_pci.c
```diff
@@ -812,7 +813,7 @@ static int aie2_get_hwctx_status(struct amdxdna_client *client,
 	struct amdxdna_client *tmp_client;
 	int ret;
 
-	if (!amdxdna_is_admin())
+	if (!capable(CAP_SYS_ADMIN))
```
houlz0507: Should we allow user to dump its own hwctx?

maxzhen: I thought we allow this in OOT driver, no? The check in OOT driver is either admin or owner of the ctx.

NishadSaraf: OOT driver doesn't have access control. Gated the access to the admin and the client owner/user.

houlz0507 (approved):

maxzhen (approved):

PR #1292, opened 2026-05-07
xrt submodule update and shim changes for xrt-smi cleanup

aktondak:
This PR moves the smi_xdna.cpp implementation to top level xrt which is now common with windows so we dont duplicate the config writing logic.
This PR also updates the xrt submodule to pull in the code changes for that code cleanup.
The xdna shim now links to smi_ryzen.cpp through xrt_coreutil and passes the config through device query accordingly


NishadSaraf:
retest this please

maxzhen:
It is not a clean solution for sharing a .cpp file in SHIM from XRT repo. The PR will be updated to expose proper internal APIs for SHIM to call instead of sharing the .cpp file. Marking it as do-not-merge for now. Will review when the new solution is implemented and PR is updated.

NishadSaraf:
retest this please

aktondak:
retest this please

maxzhen (approved):

PR #1296, opened 2026-05-08
accel/amdxdna: Add suspend/resume support for AIE4 platform

xdavidz:
On suspend, hardware is stopped in a way that matches each device flavor (PF, VF, or classic).

VF suspend walks all clients and tears down hardware contexts, then tears down the mailbox;

PF suspend should take care of firmware shutdown, which will disable services for VFs if any as well; Thus, VF only needs to tear down the mailbox.

Classic suspend takes care of everything, including hardware contexts, firmware shutdown, and mailbox.

On resume, PCI is re-enabled, bus mastering restored, the appropriate hardware start routine is run, and hardware contexts are recreated for VF/classic.

PF resume also re-sends CREATE_VFS when the device has VFs so virtual functions are brought back consistently.

### DIFF
--- a/drivers/accel/amdxdna/aie4_ctx.c
+++ b/drivers/accel/amdxdna/aie4_ctx.c
@@ -116,7 +116,7 @@ static int aie4_msg_destroy_context(struct amdxdna_dev_hdl *ndev, u32 hw_context
 	return aie_send_mgmt_msg_wait(&ndev->aie, &msg);
 }
 
-static int aie4_hwctx_create(struct amdxdna_hwctx *hwctx)
+int aie4_hwctx_create(struct amdxdna_hwctx *hwctx)
 {
 	DECLARE_AIE_MSG(aie4_msg_create_hw_context, AIE4_MSG_OP_CREATE_HW_CONTEXT);
 	struct amdxdna_client *client = hwctx->client;
@@ -132,6 +132,9 @@ static int aie4_hwctx_create(struct amdxdna_hwctx *hwctx)
 			 ndev->partition_id, hwctx->num_tiles);
 		return -EINVAL;
 	}
+
+	if (priv->state == CTX_STATE_CONNECTED)
+		return 0;
 
 	req.partition_id = ndev->partition_id;
 	req.request_num_tiles = hwctx->num_tiles;
@@ -165,21 +168,37 @@ static int aie4_hwctx_create(struct amdxdna_hwctx *hwctx)
 
 	priv->hw_ctx_id = resp.hw_context_id;
 	hwctx->doorbell_offset = resp.doorbell_offset;
+	priv->state = CTX_STATE_CONNECTED;
 
 	return 0;
 }
 
-static void aie4_hwctx_destroy(struct amdxdna_hwctx *hwctx)
+int aie4_hwctx_destroy(struct amdxdna_hwctx *hwctx)
 {
 	struct amdxdna_client *client = hwctx->client;
 	struct amdxdna_hwctx_priv *priv = hwctx->priv;
 	struct amdxdna_dev *xdna = client->xdna;
 	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+	int ret;
 
 	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
 
-	aie4_msg_destroy_context(ndev, priv->hw_ctx_id);
+	if (priv->state == CTX_STATE_DISCONNECTED)
+		return 0;
+
+	ret = aie4_msg_destroy_context(ndev, priv->hw_ctx_id);
+	if (ret)
+		return ret;
+
+	/* wake up if there are still pending hwctx(s) */
+	priv->state = CTX_STATE_DISCONNECTED;
+	wake_up_all(&priv->cert_comp->waitq);
+
+	/* release cert comp */
 	aie4_put_cert_comp(priv->cert_comp);
+	priv->cert_comp = NULL;
+
+	return ret;
 }
 
 static void aie4_hwctx_umq_fini(struct amdxdna_hwctx *hwctx)
--- a/drivers/accel/amdxdna/aie4_pci.c
+++ b/drivers/accel/amdxdna/aie4_pci.c
@@ -697,6 +697,207 @@ dev_exit:
 	return ret;
 }
 
+static void aie4_hwctx_suspend_all(struct amdxdna_dev_hdl *ndev)
+{
+	struct amdxdna_dev *xdna = ndev->aie.xdna;
+	struct amdxdna_client *client;
+	struct amdxdna_hwctx *hwctx;
+	unsigned long hwctx_id;
+	int idx;
+
+	amdxdna_for_each_client(xdna, client) {
+		idx = srcu_read_lock(&client->hwctx_srcu);
+		amdxdna_for_each_hwctx(client, hwctx_id, hwctx)
+			aie4_hwctx_destroy(hwctx);
+		srcu_read_unlock(&client->hwctx_srcu, idx);
+	}
+
+	XDNA_DBG(xdna, "Finished hwctx suspend");
+}
+
+static int aie4_hwctx_resume_all(struct amdxdna_dev_hdl *ndev)
+{
+	struct amdxdna_dev *xdna = ndev->aie.xdna;
+	struct amdxdna_client *client;
+	struct amdxdna_hwctx *hwctx;
+	unsigned long hwctx_id;
+	int ret, idx;
+
+	amdxdna_for_each_client(xdna, client) {
+		idx = srcu_read_lock(&client->hwctx_srcu);
+		amdxdna_for_each_hwctx(client, hwctx_id, hwctx) {
+			ret = aie4_hwctx_create(hwctx);
+			if (ret)
+				goto error;
+		}
+		srcu_read_unlock(&client->hwctx_srcu, idx);
+	}
+
+	XDNA_DBG(xdna, "Finished hwctx resume");
+	return 0;
+error:
+	srcu_read_unlock(&client->hwctx_srcu, idx);
+	XDNA_DBG(xdna, "Failed hwctx resume");
+	return ret;
+}
+
+static int aie4_pf_suspend(struct amdxdna_dev *xdna)
+{
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+	/* PF suspend will cleanup fw status */
+	aie4_pf_hw_stop(ndev);
+
+	XDNA_DBG(xdna, "pf suspend done");
+	return 0;
+}
+
+static int aie4_vf_suspend(struct amdxdna_dev *xdna)
+{
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+	aie4_hwctx_suspend_all(ndev);
+	/* when PF and VF both present, PF suspend will do cleanup for all VFs */
+	aie4_mailbox_fini(ndev);
+
+	XDNA_DBG(xdna, "vf suspend done");
+	return 0;
+}
+
+static int aie4_classic_suspend(struct amdxdna_dev *xdna)
+{
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+	aie4_hwctx_suspend_all(ndev);
+	aie4_classic_hw_stop(ndev);
+
+	XDNA_DBG(xdna, "classic suspend done");
+	return 0;
+}
+
+static int aie4_pf_resume(struct amdxdna_dev *xdna)
+{
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+	struct pci_dev *pdev = to_pci_dev(xdna->ddev.dev);
+	int ret;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+
+	ret = pci_enable_device(pdev);
+	if (ret) {
+		XDNA_ERR(xdna, "enable pci device failed %d", ret);
+		return ret;
+	}
+	pci_set_master(pdev);
+
+	ret = aie4_pf_hw_start(ndev);
+	if (ret) {
+		XDNA_ERR(xdna, "hw_start failed %d", ret);
+		goto pci_disable;
+	}
+
+	/* restore fw status */
+	if (ndev->num_vfs) {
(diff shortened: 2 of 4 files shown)


drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -694,6 +694,180 @@ static int aie4_get_array(struct amdxdna_client *client,
...
+	int idx;
+
+	list_for_each_entry(client, &xdna->client_list, node) {
+		idx = srcu_read_lock(&client->hwctx_srcu);
+		amdxdna_for_each_hwctx(client, hwctx_id, hwctx)
+			aie4_hwctx_resume(hwctx);
+		srcu_read_unlock(&client->hwctx_srcu, idx);
+	}
+
+	XDNA_DBG(xdna, "Finished hwctx resume");
```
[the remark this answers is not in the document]
xdavidz: yes, added return value check

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -694,6 +694,180 @@ static int aie4_get_array(struct amdxdna_client *client,
...
+
+	XDNA_DBG(xdna, "vf suspend done");
+	return 0;
+}
+
+static int aie4_classic_suspend(struct amdxdna_dev *xdna)
+{
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
```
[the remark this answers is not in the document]
xdavidz: fixed for classic suspend

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -694,6 +694,180 @@ static int aie4_get_array(struct amdxdna_client *client,
...
+}
+
+static int aie4_vf_suspend(struct amdxdna_dev *xdna)
+{
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+	aie4_hwctx_suspend_all(ndev);
+	/* when PF and VF both present, PF suspend will do cleanup for all VFs */
+	aie4_mailbox_fini(ndev);
```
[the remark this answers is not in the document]
xdavidz: invalid comment, with pf and vf, the clear up work is done in PF. VF doesn't need to send mailbox messages in case the PF already tear down the firmware which is providing the service of mailbox.

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -785,6 +961,8 @@ const struct amdxdna_dev_ops aie4_vf_ops = {
 	.get_aie_info		= aie4_get_info,
 	.get_array		= aie4_get_array,
 	.get_coredump		= aie4_get_aie_coredump,
+	.resume			= aie4_vf_resume,
+	.suspend		= aie4_vf_suspend,
 };
```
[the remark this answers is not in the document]
xdavidz: runtime-PM is not supported yet, it should be handled well when KMQ+TDR is in the picture.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -334,3 +334,22 @@ int aie4_hwctx_valid_doorbell(struct amdxdna_client *client, u32 vm_pgoff)
...
 	return 0;
 }
+
+void aie4_hwctx_suspend(struct amdxdna_hwctx *hwctx)
+{
+	struct amdxdna_dev *xdna = hwctx->client->xdna;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+
+	aie4_hwctx_destroy(hwctx);
```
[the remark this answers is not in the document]
xdavidz: cert_comp and doorbell_offset should be set to invalid.
priv will be freed, thus hw_ctx_id is not needed to be set to invalid.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -334,3 +334,22 @@ int aie4_hwctx_valid_doorbell(struct amdxdna_client *client, u32 vm_pgoff)
...
+}
+
+int aie4_hwctx_resume(struct amdxdna_hwctx *hwctx)
+{
+	struct amdxdna_dev *xdna = hwctx->client->xdna;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+
+	return aie4_hwctx_create(hwctx);
+}
```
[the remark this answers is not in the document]
xdavidz: fixed with ret checking

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -694,6 +694,202 @@ static int aie4_get_array(struct amdxdna_client *client,
...
+		ret = aie_send_mgmt_msg_wait(&ndev->aie, &msg);
+		if (ret) {
+			XDNA_ERR(xdna, "create vfs op failed %d", ret);
+			goto pci_disable;
+		}
+	}
+
+	XDNA_DBG(xdna, "pf resume done");
+	return 0;
+
```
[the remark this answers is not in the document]
xdavidz: fixed.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -180,6 +180,8 @@ static void aie4_hwctx_destroy(struct amdxdna_hwctx *hwctx)
 
 	aie4_msg_destroy_context(ndev, priv->hw_ctx_id);
 	aie4_put_cert_comp(priv->cert_comp);
+	priv->cert_comp = NULL;
+	hwctx->doorbell_offset = AMDXDNA_INVALID_DOORBELL_OFFSET; //invalid offset
```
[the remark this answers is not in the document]
xdavidz: resolved by adding hwctx state.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -334,3 +354,22 @@ int aie4_hwctx_valid_doorbell(struct amdxdna_client *client, u32 vm_pgoff)
...
 	return 0;
 }
+
+void aie4_hwctx_suspend(struct amdxdna_hwctx *hwctx)
+{
+	struct amdxdna_dev *xdna = hwctx->client->xdna;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+
+	aie4_hwctx_destroy(hwctx);
```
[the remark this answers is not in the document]
xdavidz: once disconnected, it cannot be run. But calling multiple times is allowed.

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -697,6 +697,234 @@ static int aie4_get_array(struct amdxdna_client *client,
...
+	struct amdxdna_client *client;
+	struct amdxdna_hwctx *hwctx;
+	unsigned long hwctx_id;
+	int idx;
+
+	list_for_each_entry(client, &xdna->client_list, node) {
+		idx = srcu_read_lock(&client->hwctx_srcu);
+		amdxdna_for_each_hwctx(client, hwctx_id, hwctx)
+			aie4_hwctx_suspend(hwctx);
+		srcu_read_unlock(&client->hwctx_srcu, idx);
```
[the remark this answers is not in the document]
xdavidz: this is expected. VF should be able to suspend without interacting with PF, the PF tear downs the firmware.

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -697,6 +697,234 @@ static int aie4_get_array(struct amdxdna_client *client,
 	return ret;
 }
 
+static int aie4_set_state(struct amdxdna_client *client,
```
houlz0507: why this is part of suspend and resume?

drivers/accel/amdxdna/aie4_ctx.c
houlz0507: Please merge aie4_hwctx_destroy() and aie4_hwctx_suspend(). Having a function just directly calling the other function is redundant.

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -697,6 +697,234 @@ static int aie4_get_array(struct amdxdna_client *client,
...
+
+static void aie4_hwctx_suspend_all(struct amdxdna_dev_hdl *ndev)
+{
+	struct amdxdna_dev *xdna = ndev->aie.xdna;
+	struct amdxdna_client *client;
+	struct amdxdna_hwctx *hwctx;
+	unsigned long hwctx_id;
+	int idx;
+
+	list_for_each_entry(client, &xdna->client_list, node) {
```
houlz0507: Please use amdxdna_for_each_client.

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -697,6 +697,234 @@ static int aie4_get_array(struct amdxdna_client *client,
...
+	XDNA_DBG(xdna, "Failed hwctx resume");
+	return ret;
+}
+
+static int aie4_pf_suspend(struct amdxdna_dev *xdna)
+{
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+	aie4_pf_hw_stop(ndev);
```
houlz0507, line 750: Do you need to call aie4_sriov_stop() first?

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -697,6 +697,234 @@ static int aie4_get_array(struct amdxdna_client *client,
...
+
+static int aie4_pf_resume(struct amdxdna_dev *xdna)
+{
+	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
+	struct pci_dev *pdev = to_pci_dev(xdna->ddev.dev);
+	int ret, num_vf;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+
+	ret = pci_enable_device(pdev);
```
houlz0507, line 789: Add pci_disable_device to *_stop() and add pci_enable_device to *_start()?

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -697,6 +697,234 @@ static int aie4_get_array(struct amdxdna_client *client,
...
+	}
+	pci_set_master(pdev);
+
+	ret = aie4_pf_hw_start(ndev);
+	if (ret) {
+		XDNA_ERR(xdna, "hw_start failed %d", ret);
+		goto pci_disable;
+	}
+
+	num_vf = pci_num_vf(pdev);
```
houlz0507: Should this be in pf_hw_start?

drivers/accel/amdxdna/aie4_pci.c
xdavidz: > why this is part of suspend and resume?

This is not part of the change, but the single function squeezed into the functions of init/fini. It looks not well organized. We might reorg the code function later. I move it back so no changes will be in this patch.

drivers/accel/amdxdna/aie4_ctx.c
xdavidz: deleted the suspend/resume functions for now.

drivers/accel/amdxdna/aie4_pci.c
xdavidz: fixed.

drivers/accel/amdxdna/aie4_pci.c
xdavidz: > Do you need to call aie4_sriov_stop() first?

For pf suspend? This behavior needs to be documented.
For now, we don't clear existing configured VF by sriov_stop, and during resume we just resend mailbox msg to setup VF, everything works good. We can start running end-to-end test after sleep.

But by checking some referrence design docs, it seems depends on the firmware deisgn and system.
clearing up the existing sriovs is always the most reliable way for suspend.
But for resume, the user will have to recreate the VFs becuase those VFs won't be preserved after the sriov_stop.

On our system, both works for now. We might pick one approach and also document the expected behavior.

maxzhen: > > Do you need to call aie4_sriov_stop() first?
> 
> For pf suspend? This behavior needs to be documented. For now, we don't clear existing configured VF by sriov_stop, and during resume we just resend mailbox msg to setup VF, everything works good. We can start running end-to-end test after sleep.
> 
> But by checking some referrence design docs, it seems depends on the firmware deisgn and system. clearing up the existing sriovs is always the most reliable way for suspend. But for resume, the user will have to recreate the VFs becuase those VFs won't be preserved after the sriov_stop.
> 
> On our system, both works for now. We might pick one approach and also document the expected behavior.

After suspending, the system may or may not be powered off. If system is not powered off (s3) and PF does not clear all VFs before suspending, FW will still remember all VFs after resume, but VFs may not be visible on PCIe bus. So, it seems better to always clear VFs before suspending and recreate them after resume. Today, it works because you call aie4_pf_hw_stop() which will send suspend command to FW, which may clear up all VFs, I guess. But, we don't have to rely on this specific FW behavior to make driver work stably, I think.

xdavidz: After talking with Lizhi and Max, we can try the following flow for now for both S3 and S4.

During suspend, the VF is indeed suspended prior to PF is suspended, however a pci_sriov_disable on Linux will trigger VF driver removal while will tear down VF driver thus resume won't be able to resume hwctx etc.

A reliable way today is to not issue pci_sriov_disable during suspend. 
The fw is stateless, thus after resume.

The PF driver should tell fw if there are existing vfs during suspend by sending a mailbox message.
The pci_sriov_enable should only be called if current VFs is zero in case of S4 poweroff system and forget this config even. For S3, this is not needed, the VFs preserved across each s3 sleep request.

Tested on SRIOV board.

drivers/accel/amdxdna/aie4_pci.c
xdavidz: The hw_start() and hw_stop() are shared between_init/_fini and _resume/_suspend.
For _init and _fini it has been called already in aie4m_pcidev_init (pcim_enable_device and pci_set_master).

drivers/accel/amdxdna/aie4_pci.c
xdavidz: refactored to new approach, please re-review.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -255,6 +274,7 @@ void aie4_hwctx_fini(struct amdxdna_hwctx *hwctx)
 	aie4_hwctx_destroy(hwctx);
 	aie4_hwctx_umq_fini(hwctx);
 	kfree(hwctx->priv);
+	hwctx->priv = NULL;
```
houlz0507: Why is this needed?

xdavidz: fixed.

houlz0507 (approved):

PR #1297, opened 2026-05-08
accel/amdxdna: Add iommu_bypass debugfs for aie4 only

xdavidz:
Add debug only debugfs for aie4.

### DIFF
--- a/drivers/accel/amdxdna/Kbuild
+++ b/drivers/accel/amdxdna/Kbuild
@@ -60,3 +60,4 @@ amdxdna-$(CONFIG_PCI_IOV) += \
 
 amdxdna-$(CONFIG_DEBUG_FS) += \
 	amdxdna_debugfs.o \
+	aie4_debugfs.o \
--- a/drivers/accel/amdxdna/aie4_debugfs.c
+++ b/drivers/accel/amdxdna/aie4_debugfs.c
@@ -3,15 +3,19 @@
  * Copyright (C) 2026, Advanced Micro Devices, Inc.
  */
 
-#include "amdxdna_cbuf.h"
-#include "amdxdna_debugfs.h"
-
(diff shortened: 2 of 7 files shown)


maxzhen (commented):
Is the bypass mode value differing from system to system? If not, you don't really need a debugfs node for this. You can have a global macro to define the mode and make compile time decision. This is much simpler and easier to upstream, I think.

xdavidz:
> Is the bypass mode value differing from system to system? If not, you don't really need a debugfs node for this. You can have a global macro to define the mode and make compile time decision. This is much simpler and easier to upstream, I think.

I would keep this debug only for bringup. We do not need to upstream this. 
@houlz0507 you are ok with this not being up-streamed right? This is bringup only feature, that the very first step is alawys the iommu_bypass mode, but it requires ipulib(jtag) to setup a register when mpnpufw is already loaded by the driver. We are just trying to make this a little user-friendly. Any end-user of repleased product would not need to use this feature.

xdavidz:
> > Is the bypass mode value differing from system to system? If not, you don't really need a debugfs node for this. You can have a global macro to define the mode and make compile time decision. This is much simpler and easier to upstream, I think.
> 
> I would keep this debug only for bringup. We do not need to upstream this. @houlz0507 you are ok with this not being up-streamed right? This is bringup only feature, that the very first step is alawys the iommu_bypass mode, but it requires ipulib(jtag) to setup a register when mpnpufw is already loaded by the driver. We are just trying to make this a little user-friendly. Any end-user of repleased product would not need to use this feature.

after talking with Max. I think we need to do 2 things:
1) covert debugfs to a ifdef macro to minimize the engineering complicity.
2) using OOT driver for using cma for now. I will update this PR later.

xdavidz:
use 1466 to address this, close it now.

PR #1312, opened 2026-05-14
XRT Update XRT submodule to latest commit

ManojTakasi:
XRT Update XRT submodule to latest commit


chvamshi-xilinx:
Looks like io_uring changes are causing the build failure. Please revert it in XRT master and update the commit here.

PR #1343, opened 2026-05-20
accel/amdxdna: update aie4_msg_priv.h and callers for protocol major v6 fw

hlaccabu:


drivers/accel/amdxdna/aie4_msg_priv.h
```diff
@@ -103,7 +105,9 @@ struct aie4_msg_create_hw_context_req {
 #define AIE4_MSG_PASID GENMASK(19, 0)
 #define AIE4_MSG_PASID_VLD GENMASK(31, 31)
 	__u32 pasid;
-	__u32 priority_band;
+	__u32 priority_band  : 8;
+	__u32 priority_level : 8;
+	__u32 restore_id     : 16;
```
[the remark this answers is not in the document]
hlaccabu: UAPI values like AMDXDNA_QOS_NORMAL_PRIORITY will be parsed by my other open PR here. I will use __u8 and __u16 for these fields

maxzhen (dismissed):

drivers/accel/amdxdna/aie4_msg_priv.h
```diff
@@ -115,11 +119,14 @@ struct aie4_msg_create_hw_context_resp {
...
 struct aie4_msg_destroy_hw_context_req {
 	__u32 hw_context_id;
-	__u32 resvd1;
+	__u32 graceful_flag : 1;
+	__u32 resvd1        : 31;
 } __packed;
 
 struct aie4_msg_destroy_hw_context_resp {
 	enum aie4_msg_status status;
+	__u32 restore_id : 16;
```
NishadSaraf: better to use __u16 instead of bitfield.

drivers/accel/amdxdna/aie4_msg_priv.h
```diff
@@ -189,13 +198,13 @@ struct aie4_msg_power_override_resp {
 
 #define AIE4_WORK_BUFFER_MIN_SIZE      SZ_4M
 
-struct aie4_msg_attach_work_buffer_req {
+struct aie4_msg_dram_work_buffer_req {
```
NishadSaraf: It’s better not to rename these structs, as they align with the function names.

drivers/accel/amdxdna/aie4_msg_priv.h
```diff
@@ -189,13 +198,13 @@ struct aie4_msg_power_override_resp {
...
 	__u64 buff_addr;
 	__u32 pasid;
 	__u32 buff_size;
 } __packed;
 
-struct aie4_msg_attach_work_buffer_resp {
+struct aie4_msg_dram_work_buffer_resp {
```
NishadSaraf: ditto

drivers/accel/amdxdna/aie4_msg_priv.h
```diff
@@ -32,6 +30,10 @@ enum aie4_msg_opcode {
 	AIE4_MSG_OP_POWER_OVERRIDE                   = 0x3000B,
 	AIE4_MSG_OP_AIE_RW_ACCESS                    = 0x3000E,
 	AIE4_MSG_OP_AIE_COREDUMP                     = 0x30010,
+
+	/* System Control PF/VF Opcodes */
+	AIE4_MSG_OP_DRAM_WORK_BUFFER                 = 0x40001,
```
NishadSaraf: It's best not to rename this enum since it aligns with the function names.

drivers/accel/amdxdna/aie4_msg_priv.h
```diff
@@ -115,11 +119,14 @@ struct aie4_msg_create_hw_context_resp {
 
 struct aie4_msg_destroy_hw_context_req {
 	__u32 hw_context_id;
-	__u32 resvd1;
+	__u32 graceful_flag : 1;
+	__u32 resvd1        : 31;
```
NishadSaraf: Bitfield widths are not recommended. Need to switch this to using bitmask macros.

drivers/accel/amdxdna/aie4_message.c
```diff
@@ -92,8 +92,8 @@ int aie4_query_npu_firmware_version(struct amdxdna_dev_hdl *ndev,
 int aie4_query_cert_firmware_version(struct amdxdna_dev_hdl *ndev,
 				     struct amdxdna_drm_query_firmware_version *cert_version)
 {
-	DECLARE_AIE_MSG(aie4_msg_query_cert_firmware_version,
-			AIE4_MSG_OP_QUERY_CERT_FIRMWARE_VERSION);
+	DECLARE_AIE_MSG(aie4_msg_get_uc_version,
```
NishadSaraf: Based on the comments in priv.h, no changes are necessary for this file.

NishadSaraf (changes requested):

drivers/accel/amdxdna/aie4_msg_priv.h
hlaccabu: I understand that but fw renamed the OP codes... Would you rather be aligned with FW or the current function names?

Edit: I spoke with Max, it sounds like we should keep OP code alignment with FW and change function names.

drivers/accel/amdxdna/aie4_msg_priv.h
hlaccabu: FW also renamed...

xdavidz: > FW also renamed...

No, we don't do this often. Because this is for upstream ready, once we upstream the code, we cannot change the code name here and there often even though we think they look "better". This may make the maintainers annoying it really depends. We'd rather spend more time at the fist time to make the code design clean and keep those unchanged. 

This is also part of the reason why the old driver is being obsolated, we cannot merge code to the new driver because of those unnecessary "enhancements".

I'd like @houlz0507 to comment, because he is the last step for sending out upstream patch.

houlz0507: I agree. Please avoid changing the names unless there is a strong reason to do so.

hlaccabu: Got it, thanks

drivers/accel/amdxdna/npu3_regs.c
```diff
@@ -38,10 +38,10 @@
...
 
 static const struct amdxdna_fw_feature_tbl npu3_fw_feature_table[] = {
-	{ .major = 5, .min_minor = 10 },
-	{ .features = BIT_U64(AIE4_GET_COREDUMP), .major = 5, .min_minor = 24 },
-	{ .features = BIT_U64(AIE4_RW_ACCESS), .major = 5, .min_minor = 24 },
-	{ .features = BIT_U64(AIE4_CALIBRATE_CLOCK), .major = 5, .min_minor = 24 },
+	{ .major = 6, .min_minor = 1 },
+	{ .features = BIT_U64(AIE4_GET_COREDUMP), .major = 6, .min_minor = 0 },
+	{ .features = BIT_U64(AIE4_RW_ACCESS), .major = 6, .min_minor = 0 },
+	{ .features = BIT_U64(AIE4_CALIBRATE_CLOCK), .major = 6, .min_minor = 0 },
```
houlz0507, line 44: Why do we need change the version check?

hlaccabu: None of those features would be enabled if their majors were left at 5, aie_check_protocol() requires exact major version match to enable features, not >=.  Since they are all supported in v6, should I just remove these checks instead?

xdavidz: The rule is to add version feature check together as the code change for adding this feature.

I am wondering how those feautes (core_dump, rw_access and calibrate_clock) are tested for aie4? My best guess is that those are not enabled, then please remove those check now.

Only add those when the feature is introduced with the same PR. Do not mix them, it is very hard for upstreaming patch maintainence.

drivers/accel/amdxdna/aie4_msg_priv.h
```diff
@@ -14,8 +14,6 @@ enum aie4_msg_opcode {
 	/* Classic/PF/VF common */
 	AIE4_MSG_OP_IDENTIFY                         = 0x10002,
 	AIE4_MSG_OP_SUSPEND                          = 0x10003,
-	AIE4_MSG_OP_CALIBRATE_CLOCK                  = 0x10009,
```
xdavidz, line 17: This is not acceptable to me. Why remove old op codes?
I think the old op code remain there for older fw version.

For newer fw version, we should add feature table to support newer opcode. 

I will take over this PR and drill into the details.

xdavidz (changes requested):

drivers/accel/amdxdna/aie4_msg_priv.h
maxzhen: The idea is that this version of FW is not backward compatible with old ones. The firmware team wants to start from scratch. I think this is OK since there has been no official release of the FW yet.

xdavidz:
closing this with a new combined PR

PR #1352, opened 2026-05-21
Fix imported dma-buf mmap EINVAL error in qemu virtio path

dkotipal-xilinx:
Fixes mmap failure for imported BOs seen in qemu virtio (Failed to mmap dma buf -22).

Root cause:  imported dma-buf mmap was using DRM’s fake vm_pgoff instead of a buffer-relative offset.

Fix:  set vma->vm_pgoff = 0 before dma_buf_mmap() in the imported mmap path.

Scope: imported dma-buf mapping only; non-imported shmem path is unchanged.

Result: imported mmap succeeds, no -EINVAL in this flow.

### DIFF
--- a/src/driver/amdxdna/amdxdna_gem.c
+++ b/src/driver/amdxdna/amdxdna_gem.c
@@ -611,6 +611,7 @@ static int amdxdna_gem_shmem_insert_pages(struct amdxdna_gem_obj *abo,
 
 	vma->vm_private_data = NULL;
 	vma->vm_ops = NULL;
+	vma->vm_pgoff = 0;
 	ret = dma_buf_mmap(abo->dma_buf, vma, 0);
 	if (ret) {
 		XDNA_ERR(xdna, "Failed to mmap dma buf %d", ret);


src/driver/amdxdna/amdxdna_gem.c
```diff
@@ -611,6 +611,7 @@ static int amdxdna_gem_shmem_insert_pages(struct amdxdna_gem_obj *abo,
 
 	vma->vm_private_data = NULL;
 	vma->vm_ops = NULL;
+	vma->vm_pgoff = 0;
```
wendyliang25, line 614: when user calls mmap() it will provide the vm_pgoff.
Could you provide details on "imported dma-buf mmap was using DRM’s fake vm_pgoff instead of a buffer-relative offset" ? 
drm used the offset passed from userspace to look up the gem object, and then calls the gem object mmap() function. 
`ret = dma_buf_mmap(abo->dma_buf, vma, 0);` does the mmap() from offset 0.


thanks

wendyliang25:
Hi @dkotipal-xilinx , recently, there is a change in xdna-driver and vxdna to disable mmap() for user pointer backed bo, could you try with latest kernel module in drivers/accel and latest vxdna? thanks

src/driver/amdxdna/amdxdna_gem.c
dkotipal-xilinx: getting following error
 [ 4988.850732] amdxdna 0000:c5:00.1: [drm] ERROR amdxdna_insert_pages: Failed to mmap dma buf -22
[ 4988.850746] amdxdna 0000:c5:00.1: [drm] ERROR amdxdna_gem_obj_mmap: Failed insert pages, ret -22

wendyliang25: > Failed to mmap dma buf
Hi @dkotipal-xilinx , however, amdxdna_insert_pages() passed `0` as pgoff to `dma_buf_map()`, and `dma_buf_mmap()` doesn't use `vma->vm_pgoff`, instead, it sets `vma->vm_pgoff` to the input `pgoff` which is `0`.

Recently, there is a change to amdxdna_ubuf.c to remove the mmap(). The PR to remove the mmap() from kernel module are merged before the shim RP to remove the need to call mmap() from shim, and the PR to remove mmap() from vxdna.
Have you tried the unchanged xdna-driver main branch for the .ko and the shim, vxdna to see if there is still such an issue?
If issue is still there, could you provide which linux kernel you are using?

dkotipal-xilinx: Hi @wendyliang25 , sure i will verify with latest drivers and update the status accordingly.

PR #1369, opened 2026-05-28
Refactor 0f658b79046e ("accel/amdxdna: fix drv_cmd UAF in debug BO sc…

houlz0507:
Revert the memory allocation part which is useless. The fix is actually clean drv_cmd pointer before job done. drv_cmd is not use-able in job free callback.


maxzhen (dismissed):

maxzhen:
retest this please

wendyliang25 (approved):

PR #1393, opened 2026-06-03
accel/amdxdna: drop running-kernel LLVM probe from DKMS Makefile

### LINKED ISSUE #1377: amdxdna DKMS Makefile detects the kernel compiler from the running kernel, not the build target — spurious LLVM=1 breaks builds for GCC kernels
### Summary

`src/driver/amdxdna/Makefile` decides whether to pass `LLVM=1` by inspecting the **currently running** kernel rather than the kernel being **built against**. On a host that boots a Clang/LLVM-built kernel but also has a GCC-built kernel installed, a DKMS autoinstall for the GCC kernel is wrongly given `LLVM=1`. Clang is then handed GCC-only flags baked into the target kernel's build config and the module fails to compile.

### Environment

- Primary/running kernel: locally built with Clang + LTO (`CONFIG_CC_IS_CLANG=y`), built without `CONFIG_IKCONFIG_PROC`, so **`/proc/config.gz` does not exist**.
- Also installed: a stock distro kernel built with GCC (`CONFIG_CC_IS_GCC=y`).
- `xrt-amdxdna` 2.23.0 via DKMS; the GCC kernel's package `postinst` runs `dkms autoinstall` for it.

### Root cause

Two interacting issues in `src/driver/amdxdna/Makefile`:

**1. `KERNEL_VER` ignores DKMS's `$kernelver` (line 6):**
```make
KERNEL_VER ?= $(shell uname -r)
```
DKMS exports `kernelver` for the *target* kernel, but the Makefile falls straight back to `uname -r` (the running kernel). The sibling `configure_kernel.sh` already does this correctly (see #1137):
```sh
KERNEL_VER="${KERNEL_VER:-${kernelver:-$(uname -r)}}"
```
The Makefile never received the same precedence fix.

**2. The `USE_LLVM` probe reads the running kernel (lines 40-45):**
```make
USE_LLVM ?= $(shell \
    if [ -e /proc/config.gz ]; then \
        zgrep -q "CONFIG_CC_IS_CLANG=y" /proc/config.gz 2>/dev/null && echo -n "LLVM=1"; \
    elif [ -e /boot/config-$(KERNEL_VER) ]; then \
        grep -q "CONFIG_CC_IS_CLANG=y" /boot/config-$(KERNEL_VER) 2>/dev/null && echo -n "LLVM=1"; \
    fi)
```
`/proc/config.gz` *always* describes the running kernel, never the build target — so the first branch is wrong by construction whenever you build for a kernel other than the one you booted. The `elif` would be correct only if `KERNEL_VER` were the target, which issue 1 prevents.

### Reproduction

1. Boot a Clang-built kernel (so the running kernel reports `CONFIG_CC_IS_CLANG=y`).
2. Have a GCC-built kernel's headers installed.
3. Trigger a DKMS build of amdxdna for the **GCC** kernel (e.g. `dkms autoinstall -k <gcc-kernel>`, or a kernel package upgrade).

The build invokes `... LLVM=1 modules` against the GCC kernel and fails:
```
warning: the compiler differs from the one used to build the kernel
  The kernel was built by: x86_64-linux-gnu-gcc 15.2.0
  You are using:           clang version 21.1.8
clang: error: unknown argument: '-mpreferred-stack-boundary=3'
clang: error: unknown argument: '-mindirect-branch=thunk-extern'
clang: error: unknown argument: '-fmin-function-alignment=16'
clang: error: unknown argument: '-fconserve-stack'
clang: error: unsupported argument 'bounds-strict' to option '-fsanitize='
clang: error: unsupported option '-mrecord-mcount' for target 'x86_64-unknown-linux-gnu'
```
These are all GCC flags from the target kernel's Kbuild config; Clang rejects them. Under DKMS this makes the kernel package's `postinst` exit non-zero, leaving kernel meta-packages unconfigured.

### Suggested direction

Detect the compiler from the **target** kernel, tied to `KERNEL_SRC`:

- Give `KERNEL_VER` the same precedence as `configure_kernel.sh`: prefer `$(kernelver)`, then `uname -r`.
- In the probe, drop the `/proc/config.gz` branch (it's the running kernel) and read the target's config. The most robust source, since it's bound directly to `KERNEL_SRC`, is the target build tree itself:
```make
USE_LLVM ?= $(shell grep -q "CONFIG_CC_IS_CLANG=y" $(KERNEL_SRC)/include/config/auto.conf 2>/dev/null && echo -n "LLVM=1")
```
(falling back to `/boot/config-$(KERNEL_VER)` if `auto.conf` isn't present).

**Verified locally** — applying exactly this change (target-aware `KERNEL_VER` plus the `auto.conf`-based probe) builds cleanly for the GCC target, signs and installs the module, and resolves the cascading dpkg failure.

Happy to send a PR if that direction looks right.

FIM43-Redeye:
Fixes #1377.

The `USE_LLVM` probe in `drivers/accel/amdxdna/Makefile` detected the compiler from the **running** kernel (`/proc/config.gz`, else `/boot/config-$(uname -r)`), not the **build target**. On a host booted on a clang-built kernel that builds the module for a gcc target (e.g. a DKMS autoinstall during a distro kernel upgrade), the probe forced `LLVM=1` onto the gcc build and clang rejected the kernel's gcc-only flags.

The probe is also redundant: DKMS appends `LLVM=1` to its `make` invocation when the *target* kernel is clang-built, and that command-line variable propagates to the inner kernel build via `MAKEFLAGS`. This drops the probe and the two `$(USE_LLVM)` tokens, letting the standard `LLVM=` kbuild variable flow through. Manual builds against a clang kernel pass `LLVM=1` by hand, per the usual out-of-tree convention.

Per @maxzhen's note on the issue, this targets the active `drivers/accel/amdxdna/Makefile`; the obsolete `src/driver` copy is left untouched.

### Testing
Real builds on the reporter's configuration — clang-booted host, no `/proc/config.gz`, both clang and gcc target kernels installed (gcc 15.2.0 / clang 21.1.8):

| Case | Result |
|---|---|
| gcc target, no `LLVM=1` (as DKMS does) | builds with gcc, produces `amdxdna.ko` |
| clang target, `LLVM=1` (as DKMS adds) | builds with clang |
| gcc target + spurious `LLVM=1` (old probe's behavior) | reproduces the report's exact failure (`-mpreferred-stack-boundary=3`, `-fconserve-stack`, `-mrecord-mcount`, …) |

### DIFF
--- a/drivers/accel/amdxdna/Makefile
+++ b/drivers/accel/amdxdna/Makefile
@@ -37,13 +37,6 @@ DEFINES := -DAMDXDNA_DEVEL
 XDNA_DRIVER_VERSION ?= 2.21.0
 XDNA_DATE ?= $(shell date +%Y%m%d)
 XDNA_HASH ?= $(shell git rev-parse HEAD)
-# Detect if kernel was built with Clang - check /proc/config.gz first, then /boot/config-*
-USE_LLVM ?= $(shell \
-    if [ -e /proc/config.gz ]; then \
-        zgrep -q "CONFIG_CC_IS_CLANG=y" /proc/config.gz 2>/dev/null && echo -n "LLVM=1"; \
-    elif [ -e /boot/config-$(KERNEL_VER) ]; then \
-        grep -q "CONFIG_CC_IS_CLANG=y" /boot/config-$(KERNEL_VER) 2>/dev/null && echo -n "LLVM=1"; \
-    fi)
 
 ifndef MODULE_VER_STR
 MODULE_VER_STR := $(XDNA_DRIVER_VERSION)_$(XDNA_DATE),$(XDNA_HASH)
@@ -51,11 +44,12 @@ endif
 
 DEFINES += -DMODULE_VER_STR='\"$(MODULE_VER_STR)\"'
 
+# Compiler (gcc/clang) comes from the LLVM= kbuild var set by DKMS/caller, not probed here (#1377)
 ifeq ($(XDNA_BUS_TYPE), of)
 DEFINES += -DAMDXDNA_OF
 # Generate config_kernel.h when building directly (Yocto/DKMS flow)
 modules: $(SRC_DIR)/config_kernel.h
-	$(MAKE) -C $(KERNEL_SRC) M=$(SRC_DIR) CFLAGS_MODULE="$(DEFINES)" OFT_CONFIG_AMDXDNA_PCI=$(PCI) OFT_CONFIG_AMDXDNA_OF=$(OF) $(USE_LLVM) modules
+	$(MAKE) -C $(KERNEL_SRC) M=$(SRC_DIR) CFLAGS_MODULE="$(DEFINES)" OFT_CONFIG_AMDXDNA_PCI=$(PCI) OFT_CONFIG_AMDXDNA_OF=$(OF) modules
 
 $(SRC_DIR)/config_kernel.h:
 	@echo "[INFO] Generating config_kernel.h for OF build..."
@@ -63,7 +57,7 @@ $(SRC_DIR)/config_kernel.h:
 else
 # PCI builds: CMake generates config_kernel.h
 modules:
-	$(MAKE) -C $(KERNEL_SRC) M=$(SRC_DIR) CFLAGS_MODULE="$(DEFINES)" OFT_CONFIG_AMDXDNA_PCI=$(PCI) OFT_CONFIG_AMDXDNA_OF=$(OF) $(USE_LLVM) modules
+	$(MAKE) -C $(KERNEL_SRC) M=$(SRC_DIR) CFLAGS_MODULE="$(DEFINES)" OFT_CONFIG_AMDXDNA_PCI=$(PCI) OFT_CONFIG_AMDXDNA_OF=$(OF) modules
 endif
 
 modules_install:
--- a/src/driver/amdxdna/Makefile
+++ b/src/driver/amdxdna/Makefile
@@ -36,13 +36,6 @@ DEFINES := -DAMDXDNA_DEVEL
 # Driver version
 XDNA_DATE ?= $(shell date +%Y%m%d)
 XDNA_HASH ?= $(shell git rev-parse HEAD)
-# Detect if kernel was built with Clang - check /proc/config.gz first, then /boot/config-*
-USE_LLVM ?= $(shell \
-    if [ -e /proc/config.gz ]; then \
-        zgrep -q "CONFIG_CC_IS_CLANG=y" /proc/config.gz 2>/dev/null && echo -n "LLVM=1"; \
-    elif [ -e /boot/config-$(KERNEL_VER) ]; then \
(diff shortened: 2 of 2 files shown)


maxzhen:
@kashif , @X-Ryl669, could you review? I can't set you as reviewer, not sure why...

X-Ryl669:
I don't want to bother with the already done work, but since the old driver is also in the repository, wouldn't the file `src/driver/amdxdna/Makefile` get the same change?

FIM43-Redeye:
I just scoped it to drivers/accel since @maxzhen said src/driver is obsolete and it's not wired into the build anymore. I can drop it in if he confirms but I didn't want to overdo it on obsolete code.

X-Ryl669:
@maxzhen What do you think?

kashif:
Tested this on Arch (kernel 7.0.9-arch2-1, gcc 16.1.1, NPU Strix) and it works well. The module builds clean and the reasoning holds up here.

Two things I confirmed locally that back the approach:
- Arch's `/usr/bin/dkms` already appends `LLVM=1` when the target kernel is clang-built (it reads `$kernel_source_dir/vmlinux` `.comment`, else the target `kernel_config`). So the in-Makefile probe is redundant on this distro, and more importantly dkms keys off the target, not the running kernel, which is exactly the failure mode this PR removes.
- With the probe dropped, a DKMS-style build (`make -C drivers/accel/amdxdna ...` with no `LLVM=`) produces a valid gcc `amdxdna.ko` matching the running kernel's vermagic. No regression on the gcc path.

On @X-Ryl669's question about `src/driver/amdxdna/Makefile`: it does carry the identical probe, and it isn't entirely dead code. It's still pulled in by the `legacy_driver` CMake target and as `EXTERNALSRC` in the ve2 Yocto flow (`build/build_ve2.sh`). So if those paths are still supported they have the same latent bug and could take the same one-liner. If they're being retired, probably cleaner to drop the files than keep a divergent probe. Either way I'd defer to @maxzhen on whether legacy/ve2 is in scope here. Keeping this PR focused on the active driver is fine by me.

LGTM for the accel driver.

maxzhen:
Fine by me to fix src/driver/amdxdna/Makefile if it is useful for ve2 yocto build on ARM platform.

FIM43-Redeye:
Added the same one-liner to `src/driver/amdxdna/Makefile` per maxzhen's OK — thanks to kashif for catching it's still live via the ve2 Yocto build (`EXTERNALSRC`), I'd wrongly assumed it was dead. Verified it builds clean for a gcc target the same way. The probe was actually more broken there: ve2 cross-compiles for ARM on an x86 host, so the old `/proc/config.gz` check read the build host's kernel, not the target.

maxzhen:
Added @chvamshi-xilinx for confirming the fix for VE2 build.

maxzhen:
ok to test

chvamshi-xilinx (approved):

maxzhen (approved):

X-Ryl669:
Thanks all! :+1:

PR #1404, opened 2026-06-08
accel/amdxdna: Do not autorestart syscall on SIGINT

amd-bvladino:
By default signal handler of userspace is installed with *SA_RESTART*, and on Ctrl+C wait_event_interruptible returns -ERESTARTSYS, which the Kernel will silently restart.

Convert to -EINTR so the ioctl will always propagates the signal to userspace and tools can exit on Ctrl+C interruption regardless of how they installed their handler.

### DIFF
--- a/drivers/accel/amdxdna/amdxdna_dpt.c
+++ b/drivers/accel/amdxdna/amdxdna_dpt.c
@@ -497,6 +497,7 @@ static int amdxdna_dpt_get_data(struct amdxdna_dpt *dpt,
 			if (ret) {
 				XDNA_DPT_DBG(dpt, "Wait interrupted by signal: %d", ret);
 				footer.size = 0;
+				ret = -EINTR;
 				goto exit;
 			}
 		} else {
@@ -516,10 +517,10 @@ static int amdxdna_dpt_get_data(struct amdxdna_dpt *dpt,
 	}
 
 exit:
-	if (ret == 0 || ret == -ESHUTDOWN) {
+	if (ret == 0 || ret == -ESHUTDOWN || ret == -EINTR) {
 		if (copy_to_user(buf + offset, &footer, sizeof(footer))) {
 			/*
-			 * On -ESHUTDOWN preserve the original error: user
+			 * On -ESHUTDOWN / -EINTR preserve the original error: user
 			 * space still gets the zero-size sentinel via the
 			 * footer.size = 0 already set on the shutdown
 			 * branch above, and even if the writeback fails


NishadSaraf (changes requested):
Thanks for chasing this down, @amd-bvladino — the root-cause analysis is spot on. `wait_event_interruptible()` returning `-ERESTARTSYS`, combined with the userspace handler being installed with `SA_RESTART` (glibc's `signal()` default), is exactly why Ctrl+C silently re-enters the blocking watch ioctl.

That said, I'd push back on fixing this in the driver and suggest we fix it in XRT instead.

**Why keep the kernel returning `-ERESTARTSYS`:**
- `-ERESTARTSYS` is the idiomatic return for an interruptible wait. Its whole purpose is to delegate the restart-vs-interrupt decision to userspace via `SA_RESTART`. Converting it to `-EINTR` in the driver hard-codes that policy for every caller and removes the ability of a well-behaved consumer to opt into auto-restart.
- The thing that's actually wrong here lives in userspace: the watch loop installs its SIGINT handler with `signal()` (implicit `SA_RESTART`), which is the wrong choice for a loop that wants to be interruptible. That's the bug to fix.

**Proposed XRT fix** — in [`setup()`/`restore()`](https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/tools/common/SmiWatchMode.cpp#L37-L68), install the handler via `sigaction()` without `SA_RESTART` on POSIX (Windows keeps the portable `signal()` path):

```cpp
void setup() {
  if (signal_handler_set)
    return;
#ifdef _WIN32
  old_signal_handler = signal(SIGINT, watch);
#else
  struct sigaction sa{};
  sa.sa_handler = watch;
  sigemptyset(&sa.sa_mask);
  sa.sa_flags = 0;  // no SA_RESTART: interrupted ioctl returns EINTR
  sigaction(SIGINT, &sa, &old_action);
#endif
  signal_handler_set = true;
}

void restore() {
  if (!signal_handler_set)
    return;
#ifdef _WIN32
  static_cast<void>(signal(SIGINT, old_signal_handler));
#else
  static_cast<void>(sigaction(SIGINT, &old_action, nullptr));
#endif
  signal_handler_set = false;
}
```

The shim already maps `EINTR` to a clean unwind ("Query interrupted by user"), so the watch loop exits as soon as the ioctl returns.

amd-bvladino:
@NishadSaraf , since it is a regression, proposed fix aligns with older driver implementation as well:
https://github.com/amd/xdna-driver/blob/5e4eb2173ec2acfaf6bcc5a9cd4a05161f149aec/src/driver/amdxdna/amdxdna_dpt.c#L523

NishadSaraf:
> @NishadSaraf , since it is a regression, proposed fix aligns with older driver implementation as well:
> 
> https://github.com/amd/xdna-driver/blob/5e4eb2173ec2acfaf6bcc5a9cd4a05161f149aec/src/driver/amdxdna/amdxdna_dpt.c#L523

At that time, we provided a workaround in the legacy driver. However, as mentioned above, the actual fix needs to be implemented in XRT.

PR #1417, opened 2026-06-12
accel/amdxdna: update fw versions

gyang1099:


hlaccabu (approved):

tools/info.json
```diff
@@ -63,66 +63,66 @@
...
 			"pci_device_id": "17f1",
 			"pci_revision_id": "15",
-			"version": "1.7.1.339",
+			"version": "2.5.0.172",
 			"fw_name": "npu.dev.sbin"
 		},
 		{
 			"device": "npu11",
-			"url": "https://gitlab.com/kernel-firmware/drm-firmware/-/raw/amd-ipu-staging/amdnpu/17f1_10/cert.sbin.1.2.0.32",
+			"url": "https://gitlab.com/kernel-firmware/drm-firmware/-/raw/amd-ipu-staging/amdnpu/17f1_13/cert.sbin.1.5.0.39",
```
NishadSaraf, line 106: It's preferable to create a symlink named 17f*_15 that points to 17f*_13 instead of hardcoding it in this manner.

NishadSaraf (changes requested):

tools/info.json
gyang1099: good suggestion, we will deal with this after 1.8

PR #1422, opened 2026-06-15
Bug bounty

saifuddin-xilinx:
Fixed a bug details as follws:
fix null deref in ve2/aie4 hwctx and src tree payload inline

### DIFF
--- a/drivers/accel/amdxdna/amdxdna_ctx.c
+++ b/drivers/accel/amdxdna/amdxdna_ctx.c
@@ -142,6 +142,9 @@ void *amdxdna_cmd_get_payload(struct amdxdna_gem_obj *abo, u32 *size)
 	else
 		num_masks = 1 + FIELD_GET(AMDXDNA_CMD_EXTRA_CU_MASK, cmd->header);
 
+	if (abo->mem.size < offsetof(struct amdxdna_cmd, data[num_masks]))
+		return NULL;
+
 	if (size) {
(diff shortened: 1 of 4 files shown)


drivers/accel/amdxdna/amdxdna_ctx.c
```diff
@@ -142,6 +142,9 @@ void *amdxdna_cmd_get_payload(struct amdxdna_gem_obj *abo, u32 *size)
 	else
 		num_masks = 1 + FIELD_GET(AMDXDNA_CMD_EXTRA_CU_MASK, cmd->header);
 
+	if (abo->mem.size < offsetof(struct amdxdna_cmd, data[num_masks]))
```
houlz0507, line 145: abo->mem.size is 4K aligned. So it will never smaller than offsetof(struct amdxdna_cmd, data[num_masks]).
How about adding a comment to avoid the future AI scans?

drivers/accel/amdxdna/amdxdna_ctx.c
```diff
@@ -195,6 +198,8 @@ int amdxdna_cmd_set_error(struct amdxdna_gem_obj *abo,
 
 	if (amdxdna_cmd_get_op(abo) == ERT_CMD_CHAIN) {
 		cc = amdxdna_cmd_get_payload(abo, NULL);
+		if (!cc)
```
houlz0507, line 201: here, cmd != NULL and size == NULL. So amdxdna_cmd_get_payload() will never return NULL, correct?

PR #1424, opened 2026-06-16
cmake: add Void Linux to recognized package flavors

AndreNijman:
## What

Add Void Linux to the recognized Linux package flavors in `CMake/pkg.cmake`.

On Void, `build.sh -release` aborts at configure time:

```
CMake Error at CMake/pkg.cmake:137 (message):
  Unknown Linux package flavor: void.  Supported distributions: Debian/Ubuntu
  (deb), Fedora/RHEL (rpm), Arch Linux (TGZ).
```

Void's `/etc/os-release` `ID` is `void` (with empty `ID_LIKE`), so it matched no branch.

## Fix

Route `void` to the same `TGZ` generator already used for Arch — both are
source-based distros with no native `deb`/`rpm` in this flow — and make the
comment/status message distro-neutral. The behavioral change is one line
(`MATCHES "arch"` → `MATCHES "arch|void"`).

## Validation

Tested on Void Linux (glibc 2.41, kernel 7.0.11, in-tree `amdxdna` driver).
With this change `build.sh -release -nokmod` configures cleanly and produces
`xrt_plugin.*-amdxdna.tar.gz`. Installed alongside an XRT base built from the
`xrt` submodule, `xrt-smi examine` then enumerates the Phoenix NPU
(`RyzenAI-npu1`) and mlir-aie example kernels run on it — i.e. the SHIM-only
path the README documents for kernel >= 6.14 (in-tree driver) now works on Void.

`checkpatch.pl` is not applicable (CMake-only change).

### DIFF
--- a/CMake/pkg.cmake
+++ b/CMake/pkg.cmake
@@ -123,9 +123,10 @@ elseif("${XDNA_CPACK_LINUX_PKG_FLAVOR}" MATCHES "fedora")
   endif()
 elseif("${XDNA_CPACK_LINUX_PKG_FLAVOR}" MATCHES "arch|void")
   set(CPACK_GENERATOR "TGZ")
-  # Source-based distros (Arch, Void) have no native deb/rpm in this flow, so we
-  # generate a tarball that can be repackaged into a native package (e.g. the
-  # provided Arch PKGBUILD, whose install hooks handle post-install/pre-remove).
+  # Arch and Void are binary distros, but this packaging flow has no native
+  # deb/rpm generator for them, so we emit a tarball that can be repackaged into
+  # a native package (e.g. the provided Arch PKGBUILD, whose install hooks handle
+  # post-install/pre-remove).
   set(CPACK_ARCHIVE_COMPONENT_INSTALL ON)
   message(STATUS "Arch/Void Linux detected - generating TGZ package")
   if(NOT SKIP_KMOD)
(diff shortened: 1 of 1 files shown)


maxzhen:
ok to test

maxzhen:
retest this please

maxzhen:
retest this please

maxzhen:
@AndreNijman , could you address Copilot review comments?

AndreNijman:
Thanks @maxzhen — addressed both:

**`:128` (source-based wording)** — fixed in the new commit (1cc97e1); reworded to note Arch/Void are binary distros that just lack a native deb/rpm generator in this flow.

**`:124` (regex anchoring)** — I dug into this and don't think a change is warranted: `XDNA_CPACK_LINUX_PKG_FLAVOR` holds the distro `$ID $ID_LIKE` (e.g. `void`, or `manjaro arch`), **not** the machine architecture — that's `XDNA_CPACK_ARCH`, set separately via `uname -m`/`dpkg` — so `aarch64` can't reach this branch. I also kept it unanchored deliberately: `^(arch|void)$` would drop Arch derivatives that identify via `ID_LIKE` (Manjaro → `manjaro arch`), and it stays consistent with the `debian`/`fedora` branches above, which match `ID_LIKE` the same way (Ubuntu → `ubuntu debian`). Glad to switch all branches to per-token/`STREQUAL` matching as a separate cleanup if you'd prefer.

Issue #1418, opened 2026-06-12
Question: Does this support the AMD Ryzen™ 5 7640HS  which uses xdna1?


famewolf:
Question: Does this support the AMD Ryzen™ 5 7640HS  which uses xdna1?

If not to your knowledge does anything or is the NPU in these basically a dead end?

maxzhen:
I think that xdna1 is the PHX platform and should be supported by this driver. Do you see any problem?

famewolf:
I'll know soon.  I have a mini pc enroute with that cpu however gemini told
me the driver only supported xdna2 which would have left the xdna1 series
out in the cold but I've learned the hard way to get a second opinion on
anything it tells me.

PR #1438, opened 2026-06-22
docs: clarify XDNA1 platform support

Rohithmatham12:
Summary\n- Clarify in the README that Phoenix / Hawk Point XDNA1 NPUs are supported through PCI device ID 0x1502.\n- Mention the newer Strix Point and related XDNA PCI device IDs already handled by the driver.\n\nWhy\nIssue #1418 asks whether Ryzen 5 7640HS / XDNA1 is supported. The source already binds 0x1502 and tools/info.json ships firmware metadata for that device, but the README only says "RyzenAI processor". This small docs update makes the support boundary easier to find.\n\nTesting\n- git diff --check\n\nAddresses #1418

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ On a machine with XDNA device, with both XRT and XRT plugin packages installed, 
 To run AI applications, your system needs
 * Processor:
   - To run AI applications (test machine): RyzenAI processor
+    - Supported platforms are determined by the PCI IDs compiled into the
+      driver and the matching firmware metadata included with this repository.
   - To build this repository (build machine): Any x86 processors, but recommend AMD processor :wink:
 * Operating System:
   - Ubuntu >= 22.04


maxzhen (changes requested):
1. I don't really see the need to call out XDNA1. What is special with it? Only because someone has ever asked about it?
2. It appears to be a small update, but requires long term maintenance effort. We need to update this section whenever we add any PCI device ID.

Based on the above concerns, I am not convinced that we want to take this change.

Rohithmatham12:
Thanks for the review. That makes sense — enumerating XDNA generations / PCI IDs in the README would create a stale list to maintain.

I revised the PR to remove the XDNA1 and PCI-ID callouts. The updated wording is intentionally generic: supported platforms are determined by the PCI IDs compiled into the driver and the matching firmware metadata included in the repository. That should answer the discoverability problem without requiring this section to be updated each time a device ID is added.

maxzhen:
> Thanks for the review. That makes sense — enumerating XDNA generations / PCI IDs in the README would create a stale list to maintain.
> 
> I revised the PR to remove the XDNA1 and PCI-ID callouts. The updated wording is intentionally generic: supported platforms are determined by the PCI IDs compiled into the driver and the matching firmware metadata included in the repository. That should answer the discoverability problem without requiring this section to be updated each time a device ID is added.

The updated wording does not add any value. The driver is a pci device driver which, of course, will only work if pci ID matches. I'm not going to take this change.

Rohithmatham12:
Thanks for taking another look. I understand your point that, from a driver-maintainer perspective, PCI ID matching is obvious and does not need to be explained as a driver mechanism.

The reason I still think there is a documentation gap is that the README is read by users who are not starting from the PCI driver model. They usually arrive with a product name like "Ryzen 5 7640HS", "Phoenix", "Hawk Point", or "Ryzen AI", and the current wording does not give them a clear way to determine whether their platform is expected to work. That was exactly the confusion behind #1418.

I agree that listing every XDNA generation or PCI ID in the README is the wrong long-term maintenance tradeoff. My revised intent was not to create a device-ID matrix, but to document the support-discovery rule: support is determined by what the driver binds to and what firmware metadata is present in this repo, not by broad marketing names alone.

If the current wording still feels too obvious, would you be open to a narrower user-facing sentence such as:

> To check whether a specific Ryzen AI platform is supported, compare its PCI device ID with the IDs handled by the driver and the firmware metadata shipped in this repository.

That avoids a stale list, avoids calling out XDNA1, and gives users a concrete way to self-answer compatibility questions without opening repeat issues. The distinction matters: what is obvious to driver maintainers is often exactly the missing bridge for users trying to decide whether the project supports their hardware.

If you still feel the README should not mention this at all, I understand and will leave it closed. My goal here is only to reduce repeated support questions like #1418 without adding maintenance burden.

maxzhen:
Thanks for your clarification. To be honest, we support NPU devices on ALL RyzenAI processors as we documented today, be it PHX, HPT, or STX, or any newer ones. There is really no need to ask, but I can't really stop people from asking since it's their choice.

Even if I list all products, say "PHX", or "STX", people may still ask about "XDNA1", or "7640HS", or any other possible names that may be found in various marketing documentations. Since we're engineering team, not marketing or product support team, I'd like to leave the accurate marketing product brand names to the appropriate teams. I can't really provide a detailed and accurate list of names here.

PR #1442, opened 2026-06-22
accel/amdxdna: skip PASID tag in non-SVA mode

NishadSaraf:
In carveout / physical-address mode (e.g. VFIO passthrough into a VM with no IOMMU SVA) the client has no PASID, yet the AIE4 hwctx and mem-access mailbox requests still set PASID_VLD with an invalid PASID. The NPU then issues PASID-tagged DMA that the host IOMMU rejects with INVALID_DEVICE_REQUEST, stalling the device. Add aie4_msg_pasid() to set the field only when the client actually has a PASID.

### DIFF
--- a/drivers/accel/amdxdna/aie4_ctx.c
+++ b/drivers/accel/amdxdna/aie4_ctx.c
@@ -169,7 +169,7 @@ int aie4_hwctx_create(struct amdxdna_hwctx *hwctx)
 	DECLARE_AIE_MSG(aie4_msg_create_hw_context, AIE4_MSG_OP_CREATE_HW_CONTEXT);
 	struct amdxdna_client *client = hwctx->client;
 	struct amdxdna_hwctx_priv *priv = hwctx->priv;
-	struct amdxdna_dev *xdna = hwctx->client->xdna;
+	struct amdxdna_dev *xdna = client->xdna;
 	struct amdxdna_dev_hdl *ndev = xdna->dev_handle;
 	int ret;
 
@@ -183,10 +183,8 @@ int aie4_hwctx_create(struct amdxdna_hwctx *hwctx)
 
 	req.partition_id = ndev->partition_id;
 	req.request_num_tiles = hwctx->num_tiles;
-	req.pasid = FIELD_PREP(AIE4_MSG_PASID, client->pasid) |
-		FIELD_PREP(AIE4_MSG_PASID_VLD, 1);
+	req.pasid = aie4_msg_pasid(client);
 	req.priority_band = aie4_parse_priority_to_dev(hwctx->qos.priority);
-
 	req.hsa_addr_high = upper_32_bits(amdxdna_gem_dev_addr(priv->umq_bo));
 	req.hsa_addr_low = lower_32_bits(amdxdna_gem_dev_addr(priv->umq_bo));
 
@@ -200,8 +198,7 @@ int aie4_hwctx_create(struct amdxdna_hwctx *hwctx)
 	}
 
 	XDNA_DBG(xdna, "resp msix: %d, ctx id: %d, doorbell: %d",
-		 resp.job_complete_msix_idx,
-		 resp.hw_context_id,
+		 resp.job_complete_msix_idx, resp.hw_context_id,
 		 resp.doorbell_offset);
 
 	/* setup interrupt completion per msix index */
--- a/drivers/accel/amdxdna/aie4_message.c
+++ b/drivers/accel/amdxdna/aie4_message.c
@@ -19,6 +19,15 @@
 #include "amdxdna_mailbox_helper.h"
 #include "amdxdna_pci_drv.h"
 
+u32 aie4_msg_pasid(struct amdxdna_client *client)
+{
+	if (!amdxdna_pasid_on(client))
+		return 0;
+
+	return FIELD_PREP(AIE4_MSG_PASID, client->pasid) |
+	       FIELD_PREP(AIE4_MSG_PASID_VLD, 1);
+}
+
 int aie4_suspend_fw(struct amdxdna_dev_hdl *ndev)
 {
 	DECLARE_AIE_MSG(aie4_msg_suspend, AIE4_MSG_OP_SUSPEND);
@@ -217,8 +226,7 @@ int aie4_get_aie_coredump(struct amdxdna_hwctx *hwctx,
 	int ret;
 
 	req.context_id = hwctx->fw_ctx_id;
-	req.pasid = FIELD_PREP(AIE4_MSG_PASID, hwctx->client->pasid) |
-		    FIELD_PREP(AIE4_MSG_PASID_VLD, 1);
+	req.pasid = aie4_msg_pasid(hwctx->client);
 	req.num_buffers = num_bufs;
 	req.buffer_list_addr = to_dma_addr(list_hdl, 0);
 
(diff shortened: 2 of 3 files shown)


drivers/accel/amdxdna/aie4_msg_priv.h
```diff
@@ -330,4 +332,13 @@ struct aie4_msg_calibrate_clock_resp {
 	enum aie4_msg_status status;
 } __packed;
 
+static inline u32 aie4_msg_pasid(struct amdxdna_client *client)
+{
+	if (!amdxdna_pasid_on(client))
```
xdavidz: Why we add inline function inside this msg header? Can this amdxdna_pasid_on can be aie4 specific or amdxdna generic?

If it is aie4 specific, can it be in the aie4_ctx.c only？
if it is amdxdna specific, can it be in the aie.c?

xdavidz (changes requested):

drivers/accel/amdxdna/aie4_msg_priv.h
NishadSaraf: Good point. `amdxdna_pasid_on()` is generic (stays in `amdxdna_pci_drv.h`, shared with aie2), but the PASID field layout (`AIE4_MSG_PASID`/`PASID_VLD`) is aie4-specific. I moved `aie4_msg_pasid()` into `aie4_message.c` as a `static` helper and relocated the aie4 hwctx message senders (`aie4_msg_create_context`/`aie4_msg_destroy_context`) there too, so the helper no longer lives in the protocol header. Updated in 72a7e837.

drivers/accel/amdxdna/aie4_msg_priv.h
```diff
@@ -10,6 +10,8 @@
 #include <linux/sizes.h>
 #include <linux/types.h>
 
+#include "amdxdna_pci_drv.h"
+
```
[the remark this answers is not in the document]
NishadSaraf: Done. `aie4_msg_pasid()` moved into `aie4_message.c` and the `#include "amdxdna_pci_drv.h"` removed from `aie4_msg_priv.h`, so the header is protocol-only again (consistent with `aie2_msg_priv.h`).

drivers/accel/amdxdna/aie4_msg_priv.h
```diff
@@ -330,4 +332,13 @@ struct aie4_msg_calibrate_clock_resp {
...
 } __packed;
 
+static inline u32 aie4_msg_pasid(struct amdxdna_client *client)
+{
+	if (!amdxdna_pasid_on(client))
+		return 0;
+
+	return FIELD_PREP(AIE4_MSG_PASID, client->pasid) |
+	       FIELD_PREP(AIE4_MSG_PASID_VLD, 1);
+}
```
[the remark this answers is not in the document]
NishadSaraf: Good catch. This PR intentionally scopes to the upstream/staging tree (`drivers/accel/amdxdna`). The same `PASID_VLD`-only-when-valid fix will be mirrored into the legacy tree (`src/driver/amdxdna`) in a separate PR to keep the two modules consistent. Tracking that separately.

drivers/accel/amdxdna/aie4_pci.h
```diff
@@ -78,6 +78,9 @@ int aie4_cmd_wait(struct amdxdna_hwctx *hwctx, u64 seq, u32 timeout);
 int aie4_hwctx_valid_doorbell(struct amdxdna_client *client, u32 vm_pgoff);
 int aie4_hwctx_create(struct amdxdna_hwctx *hwctx);
 void aie4_hwctx_destroy(struct amdxdna_hwctx *hwctx);
+int aie4_msg_create_context(struct amdxdna_hwctx *hwctx, u32 *hw_context_id,
+			    u32 *msix_idx, u32 *doorbell_offset);
+int aie4_msg_destroy_context(struct amdxdna_dev_hdl *ndev, u32 hw_context_id);
```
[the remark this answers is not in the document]
NishadSaraf: Good catch. Moved both prototypes under the existing `/* aie4_message.c */` section so they sit with their implementations. Updated in 1e3b670f.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -140,79 +140,30 @@ static void aie4_unlink_cert_comp(struct amdxdna_hwctx *hwctx)
 	}
 }
 
-static int aie4_msg_destroy_context(struct amdxdna_dev_hdl *ndev, u32 hw_context_id)
```
xdavidz, line 143: @NishadSaraf can you please only add the code that needed for this bug fix but do NOT move those code that is not necessary for this fix please?

Currently, it seems clean to move around the hwctx create and destroy but it will be hard in the future for adding KMQ etc.

Thus, please keep the context create and destroy remains in the aie4_ctx.c. This is different design between aie2 and aie4. We don't do everything in the aie4_message.c.

xdavidz (changes requested):

drivers/accel/amdxdna/aie4_ctx.c
NishadSaraf: Done. Kept `aie4_msg_destroy_context()` in `aie4_ctx.c` (reverted that move). `aie4_msg_create_context()` is new code introduced by this fix and builds the PASID field, so it lives in `aie4_message.c` next to the `static aie4_msg_pasid()` helper (also used by coredump/mem-access). `aie4_hwctx_create`/`destroy` themselves stay in `aie4_ctx.c`. Updated in f427766a.

drivers/accel/amdxdna/aie4_message.c
```diff
@@ -15,10 +15,79 @@
...
+	case AMDXDNA_QOS_HIGH_PRIORITY:
+		return AIE4_CONTEXT_PRIORITY_BAND_FOCUS;
+	case AMDXDNA_QOS_REALTIME_PRIORITY:
+		return AIE4_CONTEXT_PRIORITY_BAND_REAL_TIME;
+	default:
+		return AIE4_CONTEXT_PRIORITY_BAND_NORMAL;
+	}
+}
+
+int aie4_msg_create_context(struct amdxdna_hwctx *hwctx, u32 *hw_context_id,
```
xdavidz: please move those code back to the aie4_ctx.c

xdavidz (changes requested):

drivers/accel/amdxdna/aie4_message.c
NishadSaraf: Not sure how the KMQ implementation affected these cleanups, but I have made the change. Please take a look

drivers/accel/amdxdna/aie4_message.c
```diff
@@ -15,10 +15,20 @@
 #include "aie4_msg_priv.h"
 #include "aie4_pci.h"
 #include "amdxdna_ctx.h"
+#include "amdxdna_gem.h"
 #include "amdxdna_mailbox.h"
```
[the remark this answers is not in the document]
NishadSaraf: Good catch — removed the now-unused `#include "amdxdna_gem.h"` (the only user, `aie4_msg_create_context`, was reverted out). Updated.

xdavidz (approved):
Thanks. Is that possible to add a shim_test for this? I know it depends on the grub, but we don't want this to be broken again.

NishadSaraf:
> Thanks. Is that possible to add a shim_test for this? I know it depends on the grub, but we don't want this to be broken again.

The existing shim_tests and xrt_tests already cover this code path.

xdavidz:
> > Thanks. Is that possible to add a shim_test for this? I know it depends on the grub, but we don't want this to be broken again.
> 
> The existing shim_tests and xrt_tests already cover this code path.

No, I don't think any shim_test or xrt_test cover the physical carvdout test. Please specify.

NishadSaraf:
> > > Thanks. Is that possible to add a shim_test for this? I know it depends on the grub, but we don't want this to be broken again.
> > 
> > 
> > The existing shim_tests and xrt_tests already cover this code path.
> 
> No, I don't think any shim_test or xrt_test cover the physical carvdout test. Please specify.

The shim tests and xrt tests are designed to be agnostic to the addressing mode. They do not, and should not, differentiate between whether the driver is using SVA/PASID or a physical carveout. The carveout is determined by the platform/boot configuration (reserving memory using memmap=, without IOMMU SVA), not something that the tests configure. Once the system is set up for a carveout, the same unmodified tests can run through this code path.

Issue #1447, opened 2026-06-24
Mainline `amdxdna` (kernel ≤6.17) missing GET_ARRAY / AIE2 telemetry ioctls the current XRT SHIM expects


Scottcjn:
**Summary.** On a stock mainline kernel, the in-tree `amdxdna` driver does not implement the `GET_ARRAY` and AIE2 telemetry ioctls that the current XRT XDNA SHIM (XRT 2.25 / this repo's HEAD plugin) issues. Telemetry / array queries fail with `EOPNOTSUPP`/`EINVAL`, and a fully functional NPU requires building and side-loading the staging `amdxdna.ko` from this repo. This isn't surfaced by the error, so it's a non-obvious adoption papercut on gen-1 (XDNA1 / Phoenix) hardware.

### Environment
| | |
|---|---|
| Hardware | Ryzen 7 8845HS — XDNA1 / Phoenix, PCI `1022:1502`, `RyzenAI-npu1` |
| OS / kernel | Ubuntu 25.10, **kernel 6.17.0-6-generic** (mainline `amdxdna`) |
| In-tree driver | `/lib/modules/6.17.0-6-generic/kernel/drivers/accel/amdxdna/amdxdna.ko` |
| Userspace | XRT **2.25.0** + XDNA SHIM built from this repo (HEAD) |

### Symptom (mainline in-tree driver loaded)
`xrt-smi examine` enumerates the device and basic BO/exec paths work, but the SHIM's telemetry / array-info queries fail. From `shim_test`:
```
ioctl(QUERY_TELEMETRY) failed: Operation not supported
DRM_IOCTL_AMDXDNA_GET_ARRAY IOCTL failed (err=-22): Invalid argument
QUERY_TELEMETRY with header-only buffer should fail EINVAL, got 95     # 95 = EOPNOTSUPP
```
`shim_test`: **25 subtests pass** on mainline; the telemetry/get_array subtests fail.

### Root cause
The HEAD SHIM calls newer driver entry points — `amdxdna_drm_get_array_ioctl` / `aie2_get_array` and the AIE2 telemetry query (`aie2_query_telemetry` / `amdxdna_get_telemetry`) — that the in-tree 6.17 `amdxdna` predates. The in-tree driver returns `EOPNOTSUPP` (95) / `EINVAL` (-22) instead of servicing them.

### Evidence the staging driver fixes it
```
$ strings build/Release/bins/driver/amdxdna.ko | grep -iE 'get_array|telemetry'
[...]

PR #1449, opened 2026-06-24
shim: gracefully skip optional telemetry queries when the ioctl is unsupported (#1447 follow-up)

Scottcjn:
Follow-up to #1447 (in-tree `amdxdna` ≤6.17 missing `GET_ARRAY`/AIE2 telemetry ioctls).

**Problem.** When the loaded kernel driver predates `DRM_AMDXDNA_QUERY_TELEMETRY`, the ioctl returns `EOPNOTSUPP`/`EINVAL`. The SHIM's telemetry query handlers call it unconditionally, so `shim_err` throws and the **entire** `xrt-smi` report aborts — even though basic BO/exec paths work fine on that driver.

**Fix.** A small helper (`telemetry_handler::telemetry_query`) wraps the optional telemetry ioctl: on `EOPNOTSUPP`/`ENOTSUP`/`EINVAL`/`ENOTTY` it logs a one-time warning and returns `false`, so the caller returns an empty result and the rest of the report continues. Any other (real) error is rethrown. Applied to all 9 optional telemetry query handlers (aie/misc/opcode/rtos/stream_buffer, incl. the aie4 overrides). Required paths are untouched.

This turns #1447 from "this ioctl is missing" into "the SHIM handles it safely on older drivers."

**Testing.**
- Compiles cleanly under the repo's `-Werror` release build (`cmake --build ... --target xrt_driver_xdna`). The `-Werror=maybe-uninitialized` check caught that `misc_telemetry`'s scalar result needed value-initialization on the new early-return path — fixed (`result_type output{}`).
- Surgical/no-regression by construction: on a telemetry-capable driver the `try` succeeds and behavior is unchanged.
- **Disclosure:** I could not live-reproduce the unsupported-ioctl path on my test box — the in-tree 6.17 driver had been replaced by the DKMS staging build, and the other installed in-tree modules are for different kernels (vermagic mismatch). The runtime degradation is straightforward (standard errno catch) but is best confirmed against an older in-tree driver / in CI. Happy to iterate.

Surfaced during an open-source XDNA1 (Phoenix) Linux bring-up.

### DIFF
--- a/src/shim/device.cpp
+++ b/src/shim/device.cpp
@@ -18,10 +18,14 @@
 #include <sys/syscall.h>
 
 #include <algorithm>
+#include <cerrno>
 #include <climits>
 #include <cstdio>
+#include <cstdlib>
+#include <mutex>
 #include <sstream>
 #include <string>
+#include <system_error>
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


maxzhen:
ok to test

maxzhen (changes requested):
It seems that the fix should be in xrt-smi, not in SHIM. The SHIM should faithfully signaling the caller and return whatever error it sees. It is the caller to decide if this is optional or required.

Scottcjn:
You're right, thanks. Moved it out of the shim.

The shim goes back to just returning whatever errno the ioctl gives, no swallowing. The optional decision now lives in the caller in XRT: Xilinx/XRT#9907.

It skips telemetry only when the driver reports it unsupported (EOPNOTSUPP, or ENOTTY if the ioctl is not registered), which is what aie2_get_info returns from its default case on an older driver. Everything else including EINVAL still surfaces so real errors are not hidden. I gave the preemption path the same handling since it hits the same rtos_telemetry query.

That leaves nothing in this shim PR, so closing it in favor of the XRT one.

PR #1453, opened 2026-06-25
accel/amdxdna: Increase max ctx id for aie4/umq

hlaccabu:
@NishadSaraf fw for umq supports > 255 ctxs, I am assuming we want to keep 255 as max for kmq? Let me know.

### DIFF
--- a/drivers/accel/amdxdna/amdxdna_ctx.c
+++ b/drivers/accel/amdxdna/amdxdna_ctx.c
@@ -19,7 +19,7 @@
 #include "amdxdna_pci_drv.h"
 #include "amdxdna_pm.h"
 
-#define MAX_HWCTX_ID		255
+#define MAX_HWCTX_ID		1024
 #define MAX_ARG_COUNT		4095
 
 struct amdxdna_fence {


drivers/accel/amdxdna/amdxdna_ctx.c
```diff
@@ -19,7 +20,8 @@
 #include "amdxdna_pci_drv.h"
 #include "amdxdna_pm.h"
 
-#define MAX_HWCTX_ID		255
+/* KMQ has a smaller hw context id space than UMQ */
+#define MAX_HWCTX_ID(xdna)	((xdna)->dev_info->device_type == AMDXDNA_DEV_TYPE_KMQ ? 255 : SZ_1K)
```
xdavidz: @hlaccabu this might not work well, because aie4 will support KMQ soon. Let's ask Mr. @NishadSaraf, what number is good for aie2, or keep it 255 as of now, only use 1k for aie4.

I am also doubt why it is 255 not 256, if the id is start from 0, then the aie4 number should be 1023 instead of 1024.

hlaccabu: Implementation of kernel mode submission in OOT driver doesn't change the device type between AMDXDNA_DEV_TYPE_UMQ and AMDXDNA_DEV_TYPE_KMQ though, that is statically set in regs.c files.

drivers/accel/amdxdna/amdxdna_ctx.c
```diff
@@ -19,7 +19,8 @@
 #include "amdxdna_pci_drv.h"
 #include "amdxdna_pm.h"
 
-#define MAX_HWCTX_ID		255
+/* KMQ has a smaller hw context id space than UMQ */
+#define MAX_HWCTX_ID(xdna)	((xdna)->dev_info->device_type == AMDXDNA_DEV_TYPE_KMQ ? 255 : 1023)
```
xdavidz: This looks better but it doesn't address my another concern.
The KMQ will be enabled soon for aie4.

My question is:
Should aie2 also keep 255 and aie4 use 1023? If yes, the new code won't work.

xdavidz (changes requested):

drivers/accel/amdxdna/amdxdna_ctx.c
hlaccabu: I said before the device type of aie4 won't change to KMQ though, it will still have device type UMQ with kernel mode submission enabled.

xdavidz (dismissed):

xdavidz:
can this be merged?

drivers/accel/amdxdna/amdxdna_ctx.c
```diff
@@ -12,14 +12,15 @@
...
+#include <linux/sizes.h>
 #include "trace/events/amdxdna.h"
 
 #include "amdxdna_ctx.h"
 #include "amdxdna_gem.h"
 #include "amdxdna_pci_drv.h"
 #include "amdxdna_pm.h"
 
-#define MAX_HWCTX_ID		255
+#define MAX_HWCTX_ID		SZ_1K
```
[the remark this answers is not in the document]
NishadSaraf: Agreeing with the Copilot's comment, using SZ_1K can be misleading. It would be better to assign 1024 and remove the new header include.

NishadSaraf (changes requested):

drivers/accel/amdxdna/amdxdna_ctx.c
NishadSaraf: Hi @xdavidz , @hlaccabu, it is acceptable to increase this device generation agnostic value, as it is not constrained by the firmware context limit. However, since this value is exported to userspace, it is recommended to increment the driver’s minor version number to reflect this change.

NishadSaraf (dismissed):
Hi @xdavidz , @hlaccabu, it is acceptable to increase this device generation agnostic value, as it is not constrained by the firmware context limit. However, since this value is exported to userspace, it is recommended to increment the driver’s minor version number to reflect this change.

maxzhen:
@xdavidz , @NishadSaraf , Can we just decide that, on Linux, we support max 255 context? Increasing the max number may have negative impact on performance even when most of the time you will only run much less number of contexts than 255. In our critical path, we need to always perform a ctx lookup by the ID. If the ID range is too big, the lookup will take longer time. It is related to the max number of the range, not current number of IDs allocated.

xdavidz:
> @xdavidz , @NishadSaraf , Can we just decide that, on Linux, we support max 255 context? Increasing the max number may have negative impact on performance even when most of the time you will only run much less number of contexts than 255. In our critical path, we need to always perform a ctx lookup by the ID. If the ID range is too big, the lookup will take longer time. It is related to the max number of the range, not current number of IDs allocated.

@hlaccabu historical story is the code was designed to lookup the hwctx_hdl for each submit cmd in hot path, it was designed as IDR and IDR is not good if the max range is too big (based on Linux document), later it was changed to xa_load.

We don't have too much concern of the xa_load but we also not sure how much difference for each xa_load if we change the max size from 255 to 1024.

Can you measure the time usage difference between using 255 and 1024 and see how much difference in us or ms. Then we can continue to just simply switch to 1024 or may have to re-investigate how to handle 1024+ range with a better data structure.

maxzhen (approved):

PR #1457, opened 2026-06-28
amdxdna/ve2 : Fixed partition Initialization issue.

saifuddin-xilinx:
**Issue :**
- The driver was calling ve2_partition_initialize (partition init) unnecessarily, even when the same hardware context was already running on the partition.

**Root Cause 1:** 
- In ve2_response_ctx_switch_req, the condition to detect a context switch compared two adjacent FIFO entries (c_ctx vs t_ctx) instead of the FIFO head against the currently running context (active_ctx). Since t_ctx in a kernel linked list is never NULL, this comparison always evaluated incorrectly, triggering a context switch request even when no real switch was needed.

**Root Cause 2:** 
- In ve2_check_context_req, when the scheduler ran, it blindly promoted is_context_req → is_idle_due_to_context without verifying whether a different context was still waiting in the FIFO. By the time the scheduler ran, the FIFO may have already drained back to the same context, making the switch request stale.

**Fix 1:** 
- Corrected the comparison in ve2_response_ctx_switch_req from c_ctx->ctx != t_ctx->ctx to c_ctx->ctx != mgmtctx->active_ctx — so a context switch is only requested when the FIFO head is genuinely a different context from the one currently running.

**Fix 2:** 
- Added a FIFO head check in ve2_check_context_req before promoting the flag — is_idle_due_to_context is now only set when the FIFO head is actually a different context at the moment the scheduler runs, preventing stale promotions from triggering unnecessary partition re-inits.

### DIFF
--- a/src/driver/amdxdna/ve2_hwctx.c
+++ b/src/driver/amdxdna/ve2_hwctx.c
@@ -38,6 +38,10 @@ int max_col;
 module_param(max_col, int, 0644);
 MODULE_PARM_DESC(max_col, "Max column supported by this driver");
 
+int ve2_perf_optimization;;
+module_param(ve2_perf_optimization, int, 0644);
+MODULE_PARM_DESC(ve2_perf_optimization, "Enable perf mode. disabled by default.");
+
 #define CTX_TIMER	(nsecs_to_jiffies(1))
 
 /*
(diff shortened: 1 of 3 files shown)


chvamshi-xilinx (approved):

saifuddin-xilinx:
Created a new one
https://github.com/amd/xdna-driver/pull/1459

PR #1458, opened 2026-06-28
fix: reorder NPU init — PSP before SMU on Strix Halo

bong-water-water-bong:
On Strix Point/Halo (NPU4, PCI \`0x17f0\` rev \`0x10\`), the NPU SMU is embedded in the firmware package loaded by PSP. The current init order (SMU → PSP) always fails because SMU firmware hasn't been loaded yet.

Swap to PSP → SMU. Also adds a \`stop_psp_no_smu\` cleanup label that skips SMU fini when SMU was never initialized.

The community confirmed this bug affects all Strix Halo Linux users — existing workaround bypasses SMU entirely, leaving the NPU without power management. See: https://github.com/Peterc3-dev/amdxdna-strix-fix

Compile-tested on kernel 7.0.0.

### DIFF
--- a/drivers/accel/amdxdna/aie2_pci.c
+++ b/drivers/accel/amdxdna/aie2_pci.c
@@ -372,16 +372,16 @@ static int aie2_hw_start(struct amdxdna_dev *xdna)
 		goto disable_dev;
 	}
 
-	ret = aie_smu_init(ndev->aie.smu_hdl);
-	if (ret) {
-		XDNA_ERR(xdna, "failed to init smu, ret %d", ret);
-		goto free_channel;
-	}
-
 	ret = aie_psp_start(ndev->aie.psp_hdl);
 	if (ret) {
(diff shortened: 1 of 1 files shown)


houlz0507:
The issue you mentioned can not be reproduced locally. And the patch causes the hardware we are using stops to work. Starting psp failed without turning on power through SMU first.
[446812.366764] amdxdna 0000:c5:00.1: [drm] *ERROR* PSP is not ready, ret 0xffffff92
[446812.366967] amdxdna 0000:c5:00.1: [drm] *ERROR* failed to validate fw, ret -110
[446812.367039] amdxdna 0000:c5:00.1: [drm] *ERROR* aie2_hw_start: failed to start psp, ret -110

maxzhen:
@bong-water-water-bong, could you clarify how this issue can be reproduced? Is this a race/timing related issue? We test STX daily for a couple of years and has never seen such problem so far. It is impossible for us to miss it since it is a nothing-works problem.

PR #1465, opened 2026-06-30
accel/amdxdna: add query DPM level and DPM frequency table support

donwalkarsoham:


drivers/accel/amdxdna/npu3_regs.c
```diff
@@ -97,14 +97,47 @@ static int npu3_update_counters(struct aie_device *aie)
...
+	 * The AIE and NPU-H domains can sit at different levels, so each is
+	 * resolved with its own level.
+	 */
+	ret = aie4_query_dpm_level(ndev, &aieclk_level, &npuhclk_level);
 	if (ret)
 		return ret;
 
-	aie->npuclk_freq = npu_metrics.mpnpuclk_freq;
-	aie->hclk_freq = npu_metrics.npuclk_freq;
+	if (ndev->dpm_aie_levels && ndev->dpm_npuh_levels) {
```
xdavidz: @donwalkarsoham this code... generated by AI?

donwalkarsoham: Hi @xdavidz,
As discussed, this is an initial copy for giving an idea of the changes and this needs further review, changes and testing. As suggested, using some AI tools for workflow.

drivers/accel/amdxdna/aie4_message.c
```diff
@@ -117,6 +117,70 @@ int aie4_query_cert_firmware_version(struct amdxdna_dev_hdl *ndev,
...
+
+	ndev->dpm_aie_levels = 0;
+	ndev->dpm_npuh_levels = 0;
+
+	ret = aie_send_mgmt_msg_wait(&ndev->aie, &msg);
+	if (ret) {
+		XDNA_WARN(xdna, "Get DPM freq table failed (ret %d); using static table",
+			  ret);
+		return ret;
+	}
```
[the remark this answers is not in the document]
donwalkarsoham: done

drivers/accel/amdxdna/npu3_regs.c
```diff
@@ -97,14 +97,47 @@ static int npu3_update_counters(struct aie_device *aie)
...
+	 * PMF clock telemetry isn't wired up on Linux yet, and aie4 only asks
+	 * FW for a DPM *level* without programming the clock itself, so the
+	 * cached freq would otherwise be stale (the level-0 boot value). Query
+	 * the level FW most recently sent to PMFW and map it to a frequency.
+	 * The AIE and NPU-H domains can sit at different levels, so each is
+	 * resolved with its own level.
+	 */
+	ret = aie4_query_dpm_level(ndev, &aieclk_level, &npuhclk_level);
 	if (ret)
 		return ret;
```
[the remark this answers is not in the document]
donwalkarsoham: done

drivers/accel/amdxdna/aie4_pci.h
```diff
@@ -57,6 +58,12 @@ struct amdxdna_dev_hdl {
 	u32				total_col;
 	u32				max_dpm_level;
 
+	/* FW-provided DPM frequency tables (MHz); *_levels == 0 if unavailable */
```
xdavidz: There is a dpm_clk_tbl in amdxdna_dev_priv, but you add another in the amdxdna_dev_hdl.

I understand this is a cache of current dpm level, thus the other one in the priv is obsolated?

I think the logic is not.

We do have 2 dpm_clk_freq arrrays, one is hardcoded, one is dynamically quired from the fw. if dynamic query failed, then use the hardcode one. 

you might consider adding it to the amdxdna_dev_priv for dpm_clk_tbl, and only update it when query is successful. So the logic is a little different than windows driver. 

We always query from the dpm_clk_tbl, but we do update the dpm_clk_tbl once by querying the fw when aie4_pci init.

Do this make better sense?

xdavidz (changes requested):

drivers/accel/amdxdna/aie4_pci.h
donwalkarsoham: ok got it, makes sense. So, we query the FW table at aie4_pci initialization and update the dpm_clk_tbl. After this, we only query the dpm_clk_tbl during run time.

xdavidz: > ok got it, makes sense. So, we query the FW table at aie4_pci initialization and update the dpm_clk_tbl. After this, we only query the dpm_clk_tbl during run time.

1) We query dpm freq from FW during initializaioon and update the dpm_clk_tbl, if query is failed, use the default freq.
2) We query the dpm_clk_tbl freq during run time.
 2.1) During the run time query, we also call update_dpm from pmf, if pmf is not ready, we skip updating the final freq.
 
 In this way, we have good dpm_clk_tbl and also accurate freq when pmf is working. I think current code structure support this flow.

donwalkarsoham: Thanks, that flow makes sense.
One implementation detail: Since priv->dpm_clk_tbl is const and shared across devices, we can't update it in place. We can leave it as the read-only default template and add a small writable array in the per-device amdxdna_dev_hdl. At init, we can copy the defaults into that per-device array, then overwrite it with the FW values if the query succeeds (if it fails, it just keeps the defaults). At runtime we will read frequencies from this per-device table if PMF is not available. That way priv stays immutable, each device has its own correct table, and there's a single table to index.

donwalkarsoham: Added the changes.

xdavidz: > Added the changes.

I will review this today.

drivers/accel/amdxdna/aie4_message.c
```diff
@@ -126,6 +126,57 @@ int aie4_query_cert_firmware_version(struct amdxdna_dev_hdl *ndev,
...
+	struct amdxdna_dev *xdna = ndev->aie.xdna;
+	u32 i, n;
+	int ret;
+
+	memset(ndev->dpm_clk_tbl, 0, sizeof(ndev->dpm_clk_tbl));
+	for (i = 0; i < AIE4_MAX_DPM_LEVEL_COUNT && ndev->priv->dpm_clk_tbl[i].hclk; i++)
+		ndev->dpm_clk_tbl[i] = ndev->priv->dpm_clk_tbl[i];
+
+	ret = aie_send_mgmt_msg_wait(&ndev->aie, &msg);
+	if (ret) {
```
xdavidz, line 141: I think a cleaner way is to always return 0, but warn the error code from fw.

drivers/accel/amdxdna/npu3_regs.c
```diff
@@ -71,7 +71,7 @@ static int npu3_set_dpm(struct aie_device *aie, u32 dpm_level)
 	struct amdxdna_dev_hdl *ndev = aie->xdna->dev_handle;
 	int max_dpm_level = 0;
 
-	while (ndev->priv->dpm_clk_tbl[max_dpm_level].hclk)
+	while (ndev->dpm_clk_tbl[max_dpm_level].hclk)
```
xdavidz: this is wrong, you get ride of the priv->dpm_clk_tbl.

drivers/accel/amdxdna/npu3_regs.c
```diff
@@ -97,14 +97,21 @@ static int npu3_update_counters(struct aie_device *aie)
...
+	u32 aieclk_level, npuhclk_level;
 	int ret;
 
 	ret = amdxdna_get_sensors(&npu_metrics);
-	if (ret)
-		return ret;
+	if (!ret) {
+		aie->npuclk_freq = npu_metrics.mpnpuclk_freq;
+		aie->hclk_freq = npu_metrics.npuclk_freq;
+	} else if (!aie4_query_dpm_level(ndev, &aieclk_level, &npuhclk_level)) {
```
xdavidz: I do not think this logic is right, let's sync up offline.

the logic here is:
1) during fw boot (load fw), query dpm level and dpm freq
2) each time query will call_get_sensors to get most accurate data
   2.0) if query success, use the pmf sensor data
   2.1) if the query failed, maybe not support this etc, use the dpm level data. Given the dpm level data is either hardcoded or set correctly by No.1 step, we don't need to redo the query, just get what ever date from the dpm level.

Thus, I think you don't have to use windows driver's example (if 0 then not query).

I think you should:
1) still use the priv->dpm_level to update the dpm_table
2) if dpm_query failed, warn but skip updating the dpm_table

That's it.

xdavidz (changes requested):

drivers/accel/amdxdna/npu3_regs.c
donwalkarsoham: ok

xdavidz (approved):

xdavidz:
@donwalkarsoham why this is still pending? I need this to test SR-IOV TOPs.

xdavidz:
retest this please

drivers/accel/amdxdna/aie4_message.c
xdavidz: I was wrong given that we fallback to static table when it is non-zer.
Thus if (ret) return ret is right.

drivers/accel/amdxdna/aie4_message.c
```diff
@@ -126,6 +126,68 @@ int aie4_query_cert_firmware_version(struct amdxdna_dev_hdl *ndev,
...
+	struct amdxdna_dev *xdna = ndev->aie.xdna;
+	u32 i;
+	int ret;
+
+	ret = aie_send_mgmt_msg_wait(&ndev->aie, &msg);
+	if (ret) {
+		XDNA_WARN(xdna, "Get DPM freq table failed, ret %d status 0x%x",
+			  ret, resp.status);
+		return ret;
+	}
```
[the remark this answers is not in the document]
xdavidz: given that we fallback to static table when it is non-zero.
Thus if (ret) return ret is right.

NishadSaraf (dismissed):

xdavidz (dismissed):

maxzhen (dismissed):

drivers/accel/amdxdna/aie4_message.c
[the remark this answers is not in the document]
donwalkarsoham: As @xdavidz noted, we intentionally fall back to the static table on any non-zero status, so if (ret) return ret is correct here. The per-device table is now populated from the static defaults before this query, so a failed/NOTSUPP query leaves valid clocks. Keeping as-is.

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -349,11 +349,13 @@ static int aie4_query_aie(struct amdxdna_dev_hdl *ndev)
 	if (ret)
 		return ret;
 
+	ret = aie4_init_dpm_freq_table(ndev);
+	if (ret)
+		/* if query dpm from fw failed, using default value */
+		(void)ndev->priv->hw_ops->set_dpm(&ndev->aie, 0);
+
```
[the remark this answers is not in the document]
donwalkarsoham: Fixed. total_col is now set before the DPM init/fallback so max_tops/curr_tops aren't computed with total_col == 0. Also seeded ndev->dpm_clk_tbl/max_dpm_level from the static defaults at init (FW values override on success), so the fallback path has valid clocks.

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -843,17 +845,15 @@ static int aie4_query_resource_info(struct amdxdna_client *client,
...
 	u32 buf_sz;
 
 	xdna = client->xdna;
 	ndev = xdna->dev_handle;
-	priv = ndev->priv;
 
 	aie_update_counters(ndev);
-	res_info.npu_clk_max = priv->dpm_clk_tbl[ndev->max_dpm_level].hclk;
+	res_info.npu_clk_max = ndev->dpm_clk_tbl[ndev->max_dpm_level].hclk;
 	res_info.npu_tops_max = ndev->aie.max_tops;
```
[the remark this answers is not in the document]
donwalkarsoham: Fixed. ndev->dpm_clk_tbl is now populated from the static table during init, so npu_clk_max is always valid even when the FW DPM query fails or returns NOTSUPP.

drivers/accel/amdxdna/npu3_regs.c
```diff
@@ -76,7 +75,7 @@ static int npu3_set_dpm(struct aie_device *aie, u32 dpm_level)
 
 	while (ndev->priv->dpm_clk_tbl[max_dpm_level].hclk)
 		max_dpm_level++;
-	max_dpm_level--;
+	max_dpm_level--; /* last element is 0 */
```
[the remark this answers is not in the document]
donwalkarsoham: set_dpm() still uses the static priv table; in practice it has the same level count as the FW table, so max_dpm_level/max_tops match. With seeding, the per-device table equals the static one on the fallback path, so there's no divergence. Leaving the set_dpm() refactor as a possible follow-up to avoid re-touching the shared power-mode path.

drivers/accel/amdxdna/npu3_regs.c
```diff
@@ -99,15 +98,20 @@ static int npu3_set_dpm(struct aie_device *aie, u32 dpm_level)
...
+	ret = aie4_query_dpm_level(ndev, &aieclk_level, &npuhclk_level);
+	if (!ret) {
+		aie->npuclk_freq = ndev->dpm_clk_tbl[aieclk_level].npuclk;
+		aie->hclk_freq = ndev->dpm_clk_tbl[npuhclk_level].hclk;
+		aie->max_tops = NPU3_DPM_TOPS(ndev, ndev->dpm_clk_tbl[ndev->max_dpm_level].hclk);
+		if (!aie->hclk_freq)
+			XDNA_WARN(aie->xdna, "dpm freq table not populated, clk is 0");
+	} else {
+		XDNA_WARN(aie->xdna, "cannot get dpm level from fw, using default");
+	}
```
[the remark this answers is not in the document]
donwalkarsoham: Per the agreed design, on a failed runtime query we warn and keep the last-known-good clocks rather than re-deriving defaults each call. With the new init-time seeding + set_dpm(0), there's always a valid initial value, so clocks/curr_tops can't be stuck at 0. Keeping the warn-and-keep behavior.

xdavidz (approved):

NishadSaraf (approved):

PR #1466, opened 2026-06-30
accel/amdxdna: Add iommu_bypass(npu3a) option to build.sh + support cma buf

hlaccabu:


drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -25,6 +25,49 @@
 #define PSP_NOTIFY_INTR		0xD007BE11
 #define AIE4_TOTAL_COLUMN	3
 
+/* Not upstreamed: echo a bypass magic to FW right after work-buffer attach. */
+#ifdef AMDXDNA_IOMMU_BYPASS
```
xdavidz: This is good that only the IOMMU_BYPASS can enable this.
I think the cma_buf.c and .h are new thus no need to use this MACRO.

But the code added in the amdxdna_gem.c should have this MACRO.

xdavidz (changes requested):

xdavidz (dismissed):

drivers/accel/amdxdna/amdxdna_gem.c
```diff
@@ -942,6 +981,11 @@ amdxdna_gem_prime_import(struct drm_device *dev, struct dma_buf *dma_buf)
 	abo->type = AMDXDNA_BO_SHARE;
 	gobj->resv = dma_buf->resv;
 
+	abo->mem.dma_addr = AMDXDNA_INVALID_ADDR;
+	if (!amdxdna_iova_on(to_xdna_dev(dev)) &&
+	    drm_prime_get_contiguous_size(sgt) >= gobj->size)
+		abo->mem.dma_addr = sg_dma_address(sgt->sgl);
```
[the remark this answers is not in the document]
hlaccabu: By the time we reach sg_dma_address(sgt->sgl) on line 987, drm_prime_get_contiguous_size(sgt) has already dereferenced sgt->sgl. Guarding this line is redundant.

xdavidz (dismissed):

maxzhen (dismissed):
Since both iommu bypassing and cma support is for npu3a platform only, could you rename the option name in build.sh to something like -npu3a? Please rename the new macro name to NPU3A as well.

xdavidz:
> Since both iommu bypassing and cma support is for npu3a platform only, could you rename the option name in build.sh to something like -npu3a? Please rename the new macro name to NPU3A as well.

Hi @hlaccabu, please address Max's comments, change the IOMMU_BYPASSING to NPU3A so that it is very clear for NPU3A only.

drivers/accel/amdxdna/amdxdna_pci_drv.c
```diff
@@ -101,9 +106,9 @@ static const struct amdxdna_device_id amdxdna_ids[] = {
 	{ 0x17f1, 0x10, &dev_npu3_classic_info },
 	{ 0x17f2, 0x10, &dev_npu3_pf_info },
 	{ 0x17f3, 0x10, &dev_npu3_vf_info },
-	{ 0x1B0A, 0x10, &dev_npu3_classic_info },
-	{ 0x1B0B, 0x10, &dev_npu3_pf_info },
-	{ 0x1B0C, 0x10, &dev_npu3_vf_info },
+	{ 0x1B0A, 0x00, &dev_npu3_classic_info },
+	{ 0x1B0B, 0x00, &dev_npu3_pf_info },
+	{ 0x1B0C, 0x00, &dev_npu3_vf_info },
```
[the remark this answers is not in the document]
hlaccabu: 0x10 was wrong

drivers/accel/amdxdna/amdxdna_gem.c
```diff
@@ -942,6 +981,11 @@ amdxdna_gem_prime_import(struct drm_device *dev, struct dma_buf *dma_buf)
...
+
```
maxzhen: Not sure why you need these changes?

drivers/accel/amdxdna/amdxdna_pci_drv.c
```diff
@@ -163,7 +168,7 @@ static int amdxdna_drm_open(struct drm_device *ddev, struct drm_file *filp)
 	client->pasid = IOMMU_PASID_INVALID;
 	client->mm = current->mm;
 
-	if (!amdxdna_iova_on(xdna)) {
+	if (!amdxdna_iova_on(xdna) && !amdxdna_use_cma()) {
```
maxzhen: Why adding !amdxdna_use_cma() here?

maxzhen (changes requested):

drivers/accel/amdxdna/amdxdna_gem.c
hlaccabu: Copilot suggested this fix, I felt the comment was reasonable: 

"amdxdna_gem_prime_import() fails the import if the dma-buf isn't physically contiguous, even though in PASID/SVA mode the device uses UVA (amdxdna_obj_dma_addr()) and doesn't require physical contiguity. This makes importing common non-contiguous dma-bufs fail unnecessarily. Only cache a base DMA address when the import is contiguous; let the non-PASID open path reject non-contiguous imports later if a DMA address is actually needed."

drivers/accel/amdxdna/amdxdna_pci_drv.c
hlaccabu: Because amdxdna_iova_on only skips sva_init when force_iova=1, but for npu3a we want to skip this as well when we are using cma_addressing. I could make cma a fallback like carveout is, but the driver log would show "SVA bind device failed" every time. It makes sense for npu3 to have that XDNA_WARN as SVA/PASID should work but for npu3a it is expected to fail.

drivers/accel/amdxdna/amdxdna_gem.c
maxzhen: It does not make sense to me. Could you revert the change and test it?

drivers/accel/amdxdna/amdxdna_pci_drv.c
maxzhen: Please revert the change and let it show the error log. Please also remove the line#20 above as well.

drivers/accel/amdxdna/amdxdna_pci_drv.c
```diff
@@ -231,7 +237,8 @@ static void amdxdna_client_cleanup(struct amdxdna_client *client)
 	xa_for_each(&client->dev_heap_xa, heap_id, heap)
 		drm_gem_object_put(to_gobj(heap));
 	xa_destroy(&client->dev_heap_xa);
-	drm_mm_takedown(&client->dev_heap_mm);
+	if (client->xdna->dev_info->dev_heap_max_size)
+		drm_mm_takedown(&client->dev_heap_mm);
```
maxzhen: This change makes sense. This is a generic change and should be in its own commit. Make the commit message ready for upstream, please.

drivers/accel/amdxdna/amdxdna_gem.c
hlaccabu: Sure, code works without this change it's just an extra guard copilot suggested. I will test along with other suggestions from you.

xdavidz (dismissed):

drivers/accel/amdxdna/amdxdna_gem.c
```diff
@@ -964,6 +1003,10 @@ amdxdna_drm_create_share_bo(struct drm_device *dev,
 		abo = amdxdna_gem_create_ubuf_object(dev, args);
 	else if (amdxdna_use_carveout(to_xdna_dev(dev)))
 		abo = amdxdna_gem_create_cbuf_object(dev, args);
+#ifdef AMDXDNA_NPU3A
+	else if (amdxdna_use_cma())
```
maxzhen: do you still need to check if cma is used? on npu3a platform, do we have use cases where cma is not used?

drivers/accel/amdxdna/amdxdna_pci_drv.c
```diff
@@ -163,7 +163,7 @@ static int amdxdna_drm_open(struct drm_device *ddev, struct drm_file *filp)
 	client->pasid = IOMMU_PASID_INVALID;
 	client->mm = current->mm;
 
-	if (!amdxdna_iova_on(xdna)) {
+	if (!amdxdna_iova_on(xdna) && !AMDXDNA_NPU3A) {
```
maxzhen: please use macro to enable/disable line of codes, not in if statements.

drivers/accel/amdxdna/amdxdna_gem.c
hlaccabu: No, my mistake leaving that in there I realized right after I pushed. Removing that now.

drivers/accel/amdxdna/amdxdna_gem.c
```diff
@@ -8,6 +8,7 @@
 #include <drm/drm_device.h>
 #include <drm/drm_gem.h>
 #include <drm/drm_gem_shmem_helper.h>
+#include <drm/drm_prime.h>
```
maxzhen: do we really need this header here?

maxzhen (approved):

xdavidz (approved):

xdavidz:
@hlaccabu Max has approved this 4 days ago, please manage to enable this in pipeline after the PR can be merged. We will need to run basic coverage including xrt-smi validate with reasonable performance numbers.

PR #1468, opened 2026-07-01
test/shim_test: add cmd_submit EOPNOTSUPP test for aie4

donwalkarsoham:
test/shim_test: add cmd_submit EOPNOTSUPP test for aie4

AIE4 does not implement the cmd_submit callback, but a user can still
issue the EXEC_CMD ioctl directly. Add a negative test that issues the
ioctl and verifies the driver rejects it with EOPNOTSUPP instead of
dereferencing the NULL cmd_submit op.

Guards the kernel fix "accel/amdxdna: fix null pointer dereference for
cmd_submit on aie4".

### DIFF
--- a/test/shim_test/shim_test.cpp
+++ b/test/shim_test/shim_test.cpp
@@ -1277,6 +1277,26 @@ TEST_query_telemetry_short_buf(device::id_type id, std::shared_ptr<device>& sdev
       + std::to_string(err));
 }
 
+void
+TEST_cmd_submit_unsupported(device::id_type id, std::shared_ptr<device>& sdev, arg_type& arg)
+{
+  // Negative case: AIE4 has no cmd_submit op, EXEC_CMD must be rejected with EOPNOTSUPP.
+  int err = fork_query<int>(sdev.get(), [](int fd) {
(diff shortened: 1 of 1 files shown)


donwalkarsoham:
retest this please

xdavidz (approved):

xdavidz (changes requested):
Hi @donwalkarsoham , I think this test becomes invalid, because the aie4 does support submit_cmd now.

What are still missing is the we don't have decent coverage of testing all supported IOCTL.
Maybe you can close this and recreate a new shim_test to make sure all IOCTLs can be issues with no crash. or even more that supported IOCTL should only return supported expected values.

donwalkarsoham:
> Hi @donwalkarsoham , I think this test becomes invalid, because the aie4 does support submit_cmd now.
> 
> What are still missing is the we don't have decent coverage of testing all supported IOCTL. Maybe you can close this and recreate a new shim_test to make sure all IOCTLs can be issues with no crash. or even more that supported IOCTL should only return supported expected values.

ok sure, will create new tests. Closing this as suggested

PR #1487, opened 2026-07-13
accel/amdxdna: fix VE2 host queue header layout for CERT firmware

rajkumar-xilinx:
-Fix the host queue header to align with latest CERT firmware queue header format.
-Latest kernel is 6.16+ in xdna, so old timer API names were removed/renamed:
    from_timer() → timer_container_of()
    del_timer_sync() → timer_delete_sync()
-Initialise the host-queue version in the header, program the host clock time in the partition handshake
-Save per-column CERT firmware status on hwctx teardown, and expose hwctx/ctx limit fields.

### DIFF
--- a/drivers/accel/amdxdna/amdxdna_ctx.h
+++ b/drivers/accel/amdxdna/amdxdna_ctx.h
@@ -129,6 +129,55 @@ struct amdxdna_cmd {
 	u32 data[];
 };
 
+/*
+ * Context health data written into a command BO payload when a command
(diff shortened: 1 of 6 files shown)


rajkumar-xilinx:
Except 6 tests, all hw tests are passing now with this PR. Compiled accel driver against latest kernel (default kernel in xdna-driver).

chvamshi-xilinx (approved):

PR #1517, opened 2026-07-22
accel/amdxdna: remove artificial timeout from HMM range fault retry loop

maxzhen:
Remove the 1-second timeout (HMM_RANGE_DEFAULT_TIMEOUT) that caused EXEC_CMD to return -ETIME when hmm_range_fault() repeatedly returned -EBUSY or when map_invalid was repeatedly set during command submission.

This artificial timeout does not work reliably during testing and may cause failures uncessarily.


maxzhen:
retest this please - infra issue

wendyliang25 (approved):

houlz0507 (approved):

PR #1527, opened 2026-07-24
accel/amdxdna: aie4 TDR

wendyliang25:
Summary

  Hardens the aie4 kernel-mode queue (KMQ) submission path against context teardown/recreate races — both the firmware-error TDR reset and S3/S4 suspend/resume — so an in-flight submitter can never orphan a job, desync the HSA queue, or be spuriously failed across a suspend. Also
   fixes the faulting-job reporting on reset and adds/enables shim test coverage.

  On aie4 there is no driver-side TDR poller: a control-code timeout or firmware fault is delivered as an async CTX_ERROR event, which drives aie4_ctx_reset() (destroy → drain → recreate). A submitter holds only hwctx_srcu (not dev_lock), so it races that teardown; suspend/resume
   tears the same context down and back up while the app is frozen. These commits make both paths safe.

  Commits

  accel/amdxdna: aie4: time out only the faulting job on context reset — On the critical-error reset path, only the head (oldest) in-flight job is reported ERT_CMD_STATE_TIMEOUT with the cached firmware health report; jobs queued behind it never ran and are reported
  ERT_CMD_STATE_ABORT. Drains the running list in submission order and fills the health payload once.

  accel/amdxdna: aie4: harden ctx reset against concurrent kernel-mode submit — Four related fixes:

  - a. Publish/clear doorbell_addr under io_lock in create/destroy so a concurrent submitter can't ring a NULL/stale doorbell or orphan a job past the drain. A submit that finds the doorbell NULLed by a TDR returns -EAGAIN before publishing anything, so the command can be safely
  resubmitted on the recreated context.
  - b. On teardown/reset, discard any straggler's published-but-unowned queue prefix by advancing read_index up to write_index (CERT treats read == write as empty), keeping the Linux-allocated HSA queue indices monotonic — TDR/fini only, never suspend.
  - c. Distinguish TDR from suspend/resume in the submit wait via a per-ctx ctx_error_gen (bumped only when a ctx error is cached): the submitter re-acquires the fresh cert_comp and continues across a suspend/resume, and only returns -EAGAIN on a real TDR. Required because the
  shim does not retry a submit that returns -EAGAIN.
  - d. Unwind a chain interrupted mid-publish by a TDR instead of orphaning its unterminated prefix on the running list: a command that published nothing returns -EAGAIN (safe to resubmit); a chain that already published a prefix is reported ERT_CMD_STATE_ABORT and completed
  inline (its prefix may already have executed on the AIE, so it cannot be resubmitted). Both TDR-disconnect signals — a NULLed doorbell (a) and a cert_comp broken during the HSA-full wait (c) — now surface uniformly as -EAGAIN, so the unwind is driven by a single condition.

  test/shim_test: run the timeout context-health test on aie4 — Widen TEST_io_timeout's filter from dev_filter_is_npu4 to dev_filter_is_aie4_or_npu4 so the bad_timeout context-health-report test runs on npu3/npu3a (the machinery and health-data verify are already aie4-aware).

  test/shim_test: always disable force preemption on io_test exit — Force preemption is a device-global sticky setting; if a step after enabling it threw (e.g. hw-ctx creation rejected by firmware), it stayed on and broke every subsequent context-creating test. Add an RAII guard
  that clears it on any early/exception exit.

  test/shim_test: add aie4 TDR reset reap/unwind tests — Cover the previously untested aie4 KMQ TDR reset paths. All three drive a bad_timeout command to hang CERT so the queue cannot drain, making a blocked submit unblockable except by the ensuing context-error reset
  (deterministic, no timing race): TEST_tdr_timeout_and_abort (head job TIMEOUT, queued jobs ABORT), TEST_tdr_partial_chain_abort (over-long runlist blocks mid-chain → ABORT, not resubmitted or orphaned), and TEST_tdr_single_submit_eagain (a not-yet-published submit returns
  -EAGAIN, surfaced as a system_error by the shim). Gated to aie4 via dev_filter_is_aie4.

### DIFF
--- a/drivers/accel/amdxdna/aie4_ctx.c
+++ b/drivers/accel/amdxdna/aie4_ctx.c
@@ -308,11 +308,11 @@ void aie4_hwctx_destroy(struct amdxdna_hwctx *hwctx)
  * struct amdxdna_ctx_health_data (version 1) so user space can read it after
  * the command is marked timed out. If no report was cached for this context,
  * the per-generation fields are zeroed. The cached report is not consumed here:
- * every job completed by a single context reset must observe the same report,
- * so the reset cleanup (aie4_hwctx_cleanup_running_jobs) clears it exactly once
- * after all parked jobs have been served. Caller must hold io_lock, which
- * serializes this multi-word read against the async error worker overwriting
- * the cache concurrently.
+ * the errored abort path (aie4_hwctx_cleanup_running_jobs) attaches it to the faulting
+ * head job and clears it exactly once after all parked jobs have been served, so
+ * a later unrelated reset cannot observe a stale report. Caller must hold
+ * io_lock, which serializes this multi-word read against the async error worker
+ * overwriting the cache concurrently.
  */
 static void aie4_fill_health_data_locked(struct amdxdna_gem_obj *cmd_abo,
 					 struct amdxdna_hwctx *hwctx)
@@ -1117,37 +1117,46 @@ static void job_worker(struct work_struct *work)
 }
 
 /*
- * Reap the in-flight jobs parked by the worker on disconnect. On a plain
- * teardown (@errored false) they are aborted. On the critical-error reset path
- * (@errored true) they are completed as TIMEOUT with the cached firmware health
- * report attached, so waiters observe the timeout (and its per-command health
- * data) instead of a bare abort.
+ * Abort every in-flight job on this kernel-mode context's queue. On a plain
+ * teardown (@errored false) all jobs are ABORTed. On the critical-error path
+ * (@errored true) only the head job - the oldest in-flight job, i.e. the one
+ * that actually faulted - is reported as TIMEOUT with the cached firmware health
+ * report attached; the jobs queued behind it never ran, so they are ABORTed too.
+ * next_running_job() pops the head first, so the first job drained here is the
+ * faulting one.
+ *
+ * The caller must have already disconnected the context (aie4_hwctx_destroy()):
+ * that keeps the running list stable and makes the pending submitters self-abort
+ * via the doorbell_addr==NULL gate, so only the parked running list is drained
+ * here. destroy() also quiesces the worker, so the cancel_work_sync() below is a
+ * defensive no-op in that expected ordering.
  */
 void aie4_hwctx_cleanup_running_jobs(struct amdxdna_hwctx *hwctx, bool errored)
 {
 	struct amdxdna_hwctx_priv *priv = hwctx->priv;
 	struct amdxdna_sched_job *job;
+	bool timeout = errored;
 
-	/*
-	 * The worker parks (preserves) in-flight jobs on disconnect instead of
-	 * reaping them, so on teardown/reset stop it and complete the preserved
-	 * jobs here.  cancel_work_sync() ensures the worker is not touching the
-	 * list.
+	/* The worker parks in-flight jobs on disconnect; make sure it is not
+	 * touching the running list before we drain it.
 	 */
 	cancel_work_sync(&priv->job_work);
 	while ((job = next_running_job(hwctx))) {
-		if (errored)
+		if (timeout) {
 			job_timeout(job);
-		else
+			/* Only the faulting head job carries the report. */
+			timeout = false;
+		} else {
 			job_abort(job);
+		}
 	}
 
 	if (errored) {
 		/*
-		 * Every job above was filled from the same cached health report;
-		 * consume it exactly once now that all parked jobs for this reset
-		 * have been served, so a later unrelated reset cannot observe a
-		 * stale report.
+		 * The head job was filled from the cached health report; consume
+		 * it exactly once now that all parked jobs for this reset have
+		 * been served, so a later unrelated reset cannot observe a stale
+		 * report.
 		 */
 		mutex_lock(&priv->io_lock);
 		priv->cached_ctx_error_valid = false;
--- a/drivers/accel/amdxdna/aie4_ctx.c
+++ b/drivers/accel/amdxdna/aie4_ctx.c
@@ -29,8 +29,14 @@
 #define CTX_INVALID_ID			(~0U)
 #define CTX_INVALID_DOORBELL		AMDXDNA_INVALID_DOORBELL_OFFSET
 
-/* Max sub-commands in one ERT_CMD_CHAIN, matching the user-mode shim cap. */
-#define MAX_CHAINED_SUB_CMD		64
+/*
+ * Poison value for priv->doorbell_addr: an unmapped kernel address, not NULL.
+ * A valid doorbell is doorbell_base (a non-NULL pcim_iomap() of the BAR) plus a
+ * small offset, so it can never be 0; the driver still uses an explicit INVALID
+ * sentinel (mirroring CTX_INVALID_DOORBELL) rather than overloading 0/NULL, and
+ * a stray ring on a poisoned pointer faults loudly instead of writing to 0.
+ */
+#define DOORBELL_ADDR_INVALID		((void __iomem *)~0UL)
 
 static void job_worker(struct work_struct *work);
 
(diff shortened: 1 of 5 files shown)


drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -278,14 +285,28 @@ void aie4_hwctx_destroy(struct amdxdna_hwctx *hwctx)
...
+	 * moves its job onto running_job_list before this proceeds (so the reset
+	 * drain reaps it). This prevents ringing a NULL/dead doorbell and orphaning
+	 * a late-submitted job past the drain. doorbell_base is a device-level
+	 * managed mapping; just drop the pointer. User-mode has no driver doorbell.
+	 */
+	if (priv->kernel_submit) {
+		mutex_lock(&priv->io_lock);
+		priv->doorbell_addr = NULL;
+		mutex_unlock(&priv->io_lock);
+	}
```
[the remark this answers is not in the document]
wendyliang25: fixed, moved the ctx_error_gen increasement to when unlink `cert_comp` which is under `cert_comp` lock, when reading ctx_error_gen, it is before getting the cert_comp.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -734,16 +768,43 @@ static int wait_till_seq_completed(struct amdxdna_hwctx *hwctx, u64 seq)
...
+	 * Only a TDR reset bumps ctx_error_gen (in aie4_unlink_cert_comp(), before
+	 * the wake); that is what turns a broken wait into -EAGAIN so the caller
+	 * unwinds. The entry snapshot needs no lock: wait_event's schedule() commits
+	 * this read before any teardown can wake us, and ctx_error_gen only ever
+	 * increases, so a TDR that breaks this wait always leaves it != err_gen. A
+	 * benign suspend/resume caches no error and does not bump, so err_gen is
+	 * unchanged and the wait re-acquires the fresh cert_comp and continues.
 	 */
-	ret = wait_event_freezable(cert_comp->waitq, check_cmd_done(hwctx, seq, cert_comp));
+	for (;;) {
```
maxzhen: not sure why we need to busy loop in driver? shim will come back again when it sees -EAGAIN anyway.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -1106,42 +1174,87 @@ static void job_worker(struct work_struct *work)
...
 		mutex_unlock(&priv->io_lock);
 	}
+
+	/*
+	 * All jobs are drained; put the queue back to an empty, consistent state.
+	 * This runs on the TDR reset (before recreate) and on ctx fini, but not on
+	 * the normal suspend path (which does not call this and must keep its
+	 * indices so aie4_hwctx_resume_jobs() can re-drive the preserved jobs).
+	 */
+	aie4_hwctx_reset_umq_indices(hwctx);
```
maxzhen, line 1321: I'm not sure if this is the right place to do this. This function is for job clean-up and should just deal with jobs only. The queue index state belongs to the hwctx handling code. Making it a side effect of job cleaning up does not feel right.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -1354,12 +1472,38 @@ int aie4_cmd_submit(struct amdxdna_hwctx *hwctx, struct amdxdna_sched_job *job,
...
+	if (job->aie4_job_state == AIE4_JOB_STATE_SUBMITTING &&
+	    ret == -EAGAIN) {
+		/*
+		 * A chain published a prefix but a TDR reset tore the context down
+		 * mid-stream (submit_one_cmd hit the NULLed doorbell, or the
+		 * ctx_error_gen-broken HSA-full wait - both surface as -EAGAIN). Its
+		 * prefix may already have executed on the AIE, so - unlike a
+		 * not-yet-published command - it cannot be safely resubmitted. Report
+		 * it ABORTed (a completion the app observes) rather than -EAGAIN.
+		 * Complete it inline rather than moving it to the running list: the
```
maxzhen: We are assuming that the timeout will never be triggered by a sub-cmd from this job. This requires the len of runlist < len of queue. It is better to enforce / check this in submission code path with comment about the reason for checking.

maxzhen (changes requested):

drivers/accel/amdxdna/aie4_ctx.c
xdavidz: > not sure why we need to busy loop in driver? shim will come back again when it sees -EAGAIN anyway.

same question, The EAGAIN will be retried by shim code infinitely from user-land. The kernel busy loop will sleep on the waitq and hold a cert completion. In the wait_cmd we have same example code. I think we should apply the same code patten and explain how the things get handled by shim code.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -883,9 +944,16 @@ static int submit_one_cmd(struct amdxdna_hwctx *hwctx,
...
-	/* doorbell_addr is dropped under io_lock on teardown; NULL => disconnected. */
+	/*
+	 * doorbell_addr is dropped under io_lock on teardown; NULL => disconnected
+	 * by a TDR reset. Nothing has been published for this command yet, so
+	 * return -EAGAIN: a single command (or the first sub-command of a chain)
+	 * leaves the job PENDING and the caller can resubmit it on the recreated
+	 * context; a chain that already published a prefix is unwound as ABORT by
+	 * the caller (see aie4_cmd_submit).
+	 */
 	if (!priv->doorbell_addr)
```
xdavidz: You are implying the doorbell_addr is 0 or NULL is invalid. I think we have doobell_addr set to INVALID as ~0U. https://github.com/amd/xdna-driver/blob/main/include/uapi/drm/amdxdna_accel.h#L26

Why invalid doorbell_addr ~0U is not enough? Or, there is a small window that the doorbell_addr can be 0? We really should set the doorbell_addr to INVALD unless it is valid.

drivers/accel/amdxdna/aie4_ctx.c
maxzhen: After a 2nd thought, it does make sense to reset the queue here. The job state and queue content is, indeed, coupled.

NishadSaraf:
retest this please

drivers/accel/amdxdna/aie4_ctx.c
wendyliang25: Thanks @xdavidz , fixed

drivers/accel/amdxdna/aie4_ctx.c
wendyliang25: checked against number of max commands in the queue

drivers/accel/amdxdna/aie4_ctx.c
wendyliang25: as discussed, remove the for loop here, instead of for chained commands, we check if it is due to TDR or resume in the `submit_job_cmds()`, if it is resume, we will do again for the subcommand within submit_job_cmds; if it is TDR, returns, we abort.

wendyliang25:
> ## Pull request overview
> Copilot reviewed 5 out of 5 changed files in this pull request and generated no new comments.
> 
> Comments suppressed due to low confidence (1)
> **drivers/accel/amdxdna/aie4_ctx.c:1404**
> 
> * Kernel-mode submission now rejects ERT_CMD_CHAIN with command_count > CTX_MAX_CMDS (32). User-space still allows up to 64 sub-commands (e.g. src/shim/umq/hwq.cpp:339 and src/driver/amdxdna/amdxdna_ctx.h:367), so an application/shim can construct runlists that will now be rejected (-EINVAL) on aie4. Please align the user-space cap/validation with CTX_MAX_CMDS for aie4 (or plumb the queue capacity into the shim) so runlists can’t exceed the hardware queue capacity.
> 
> ```
> 	/*
> 	 * A chain (runlist) must fit within the queue. CERT advances the host-visible
> 	 * read_index only once per runlist - at the last sub-command (CHAIN_FLG_LAST_CMD),
> 	 * not per sub-command - so a chain's own entries never free a queue slot until
> 	 * the whole chain has been published and run. A chain longer than the queue
> 	 * could therefore never publish its tail: it would block forever in
> 	 * wait_till_hsa_not_full() waiting for a slot that only frees at chain end.
> 	 * Reject ccnt > CTX_MAX_CMDS. Also validate against the payload size before
> 	 * walking the handle array so a bogus count cannot drive an out-of-bounds read.
> 	 */
> 	if (!ccnt || ccnt > CTX_MAX_CMDS ||
> 	    payload_len < struct_size(payload, data, ccnt)) {
> 		XDNA_ERR(xdna, "Invalid command count %u", ccnt);
> 		return -EINVAL;
> ```

CERT doesn't allow number of sub-commands larger than the command queue, this patch set only change the kernel submission mode.

maxzhen (approved):

xdavidz (approved):

PR #1528, opened 2026-07-25
amdxdna: ve2: add per-hwctx auto coredump capture when cmd timeouts

rajkumar-xilinx:
-Auto-capture an AIE coredump when a command times out, and cache it for the next coredump request for that context.
-The coredump cache is per-hardware-context (lives in amdxdna_ctx_priv). Capture is always on: on timeout the driver snapshots the partition into a vmalloc'd buffer (keep-latest, a newer timeout overwrites the previous).
-Retrieval is consume-on-read: a coredump request is served from the hwctx cache and it is then cleared.
-If the cache is empty, the driver falls back to a live coredump capture req.

### DIFF
--- a/src/driver/amdxdna/ve2_debug.c
+++ b/src/driver/amdxdna/ve2_debug.c
@@ -109,10 +109,13 @@ static int ve2_get_array_hwctx(struct amdxdna_client *client,
 		list_for_each_entry(tmp_client, &xdna->client_list, node) {
 			unsigned long id;
 			struct amdxdna_ctx *ctx;
+			int srcu_idx;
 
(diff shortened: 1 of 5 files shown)


rajkumar-xilinx:
verification is in progress

rajkumar-xilinx:
amd-edf:/mnt/all_hw_tests/coredump/test_files# ./host.exe 
COREDUMP START
====> time: 543588 us
COREDUMP end
Coredump written to aie_coredump.txt (29360144 bytes)
AIE core status @0x338004: 0x201 (expected 0x201)
TEST PASSED

NishadSaraf:
retest this please

PR #1529, opened 2026-07-26
accel/amdxdna: label Total Power sensor in W, not mW

atassis:
`DRM_AMDXDNA_QUERY_SENSORS` describes the NPU power sensor with units that do not match the value it reports. `amdxdna_query_sensors()` sets `unitm = -3` on `npu_power` and names the units `mW`, so per the UAPI (`pow(10, unitm) * input`) a raw reading of 1118 is described as 1.118 mW when it is 1.118 W.

## Root cause

`npu_power` comes from `struct amd_pmf_npu_metrics`, which `include/linux/amd-pmf-io.h` documents as `@npu_power`: NPU power [mW]. `unitm = -3` therefore converts mW to W, which is exactly what the consumer expects: XRT's `read_data_driven_electrical()` (`xrt/src/runtime_src/core/common/sensor.cpp`) computes `input * pow(10, unitm)` and stores the result as `power_consumption_watts`. The scale factor is right; the units string labelling it is not.

## Fix

Set `units` to `W` and leave `unitm = -3` alone. The sensor becomes self-consistent under the UAPI contract, and the value every consumer sees is unchanged, so no paired userspace change is needed. `units` is not read on the XRT power path at all: only the `Total Power` label and `unitm` are.

Dropping `unitm` to 0 instead would keep `input`, `max`, `average` and `highest` exact `__u32`, but it would break the end-to-end report, since `xrt-smi` would then show 1118 W. v1 of this PR did that; see the review thread.

A second commit fixes the kernel-doc that made this easy to get wrong. `@units` was described only as "The sensor units", with nothing saying which value it names, which is what let `mW` sit next to a raw mW `input` and look right. It now names the scaled value:

```
 * @units: The unit of the scaled value (pow(10, unitm) * value), not of the raw
 *         @input, @max, @average and @highest members.
```

Comment only, no ABI or behaviour change.

## Evidence

Read on a Strix part (npu4) through `DRM_AMDXDNA_QUERY_SENSORS`, sampling the Total Power sensor at 50 Hz while dispatching a prebuilt design back-to-back:

| state | raw `input` |
| --- | --- |
| idle | 0.209 mean, max 1 |
| sustained load | 979.82 mean, peak 1118 |

At `unitm = -3` that is ~0 W idle and ~1.1 W under load, a plausible figure for this part, which is what establishes the reported quantity is watts and not milliwatts.

## hwmon

`amdxdna_hwmon_read()` converts `npu_power` with `MICROWATT_PER_MILLIWATT` and does not go through `unitm`, so hwmon `power1_input` is unaffected by this change.

### DIFF
--- a/drivers/accel/amdxdna/amdxdna_sensors.c
+++ b/drivers/accel/amdxdna/amdxdna_sensors.c
@@ -61,7 +61,7 @@ int amdxdna_query_sensors(struct amdxdna_drm_get_info *args, u32 total_col)
 	sensor.input = npu_metrics.npu_power;
 	sensor.unitm = -3;
 	scnprintf(sensor.label, sizeof(sensor.label), "Total Power");
-	scnprintf(sensor.units, sizeof(sensor.units), "mW");
+	scnprintf(sensor.units, sizeof(sensor.units), "W");
 
 	if (args->buffer_size < sizeof(sensor))
 		goto out;
--- a/include/uapi/drm/amdxdna_accel.h
+++ b/include/uapi/drm/amdxdna_accel.h
@@ -441,7 +441,8 @@ enum amdxdna_sensor_type {
  * @average: The average value of the sensor.
  * @highest: The highest recorded sensor value for this driver load for the sensor.
  * @status: The sensor status.
- * @units: The sensor units.
+ * @units: The unit of the scaled value (pow(10, unitm) * value), not of the raw
+ *         @input, @max, @average and @highest members.
  * @unitm: Translates value member variables into the correct unit via (pow(10, unitm) * value).
  * @type: The sensor type from enum amdxdna_sensor_type.
  * @pad: Structure padding.


maxzhen:
ok to test

NishadSaraf:
retest this please

NishadSaraf (changes requested):
Thanks for the detailed write-up and for checking the shim before proposing this.

I dug into the full `unitm` path and I'd like to suggest a different fix. The short version: `unitm = -3` is not actually wrong, and `unitm = 0` would regress `xrt-smi`. The cleaner minimal change is to keep `unitm = -3` and change the units string from `mW` to `W`.

**Why `unitm = 0` breaks the real consumer**

The shim (`src/shim/device.cpp`) does copy `unitm` through unmodified, but the value is consumed one layer up in XRT common, in `read_data_driven_electrical()` (`xrt/src/runtime_src/core/common/sensor.cpp`). That code computes `input * pow(10, unitm)` and stores the result directly into `power_consumption_watts` — i.e. it interprets the scaled value as **watts**.

So for a Strix reading of ~1118:

- Today (`unitm = -3`): `1118 * 10^-3 = 1.118` → `xrt-smi` reports **1.118 W** (correct).
- With `unitm = 0`: `1118 * 10^0 = 1118` → `xrt-smi` reports **1118 W** (1000x too high).

In other words, the `-3` is doing the mW→W conversion that the consumer expects; it isn't a bug, it's what makes the end-to-end power report come out right. The `input` value is in mW, but the pipeline's contract is "`pow(10, unitm) * input` yields the value the consumer treats as watts."

**Suggested change**

Keep `unitm = -3` and set the units string to `W`:

```c
sensor.type = AMDXDNA_SENSOR_TYPE_POWER;
sensor.input = npu_metrics.npu_power;
sensor.unitm = -3;
snprintf(sensor.label, sizeof(sensor.label), "Total Power");
snprintf(sensor.units, sizeof(sensor.units), "W");
```

This makes the sensor self-consistent per the UAPI contract (`pow(10, unitm) * input` = value in `units` = `1.118 W`), requires no paired XRT change, and keeps `xrt-smi` reporting the correct value. It's numerically equivalent to what you already noted in the PR description.

**Please retarget to the staging driver**

We're no longer maintaining the out-of-tree driver for aie2/aie4 (`src/driver/amdxdna/aie2_pci.c` / `aie4_pci.c`); development has moved to the upstream/staging driver. The equivalent code now lives in `drivers/accel/amdxdna/amdxdna_sensors.c`, where `amdxdna_query_sensors()` has the same `unitm = -3` / `mW` on the power sensor. Could you please submit the `units = "W"` change there instead?

atassis:
@NishadSaraf Hi! I have forgot to look into xrt, thanks for finding that out! Recording why v1 went the other way, since it may be worth a doc fix.

I read the UAPI as self-describing: `units` names the unit of `input`, `unitm` scales it. `input` is already mW, so `unitm = 0 / units = "mW"` was the natural encoding, and integer sub-units are what kernel interfaces normally hand out (hwmon `power*_input` in uW, ACPI power meter in mW). W as a base unit is unusual for a kernel interface, so mW read as the conventional choice rather than an XRT-specific one.

What settles it is not which unit is nicer: `unitm` is released UAPI with a live consumer, so the effective value cannot move. My precision argument does not survive either, since `input` stays an exact `__u32` either way and XRT renders both through `precision(double, 3)`.

The header is what misled me. `@unitm: Translates value member variables into the correct unit via (pow(10, unitm) * value)` never says what `units` is relative to. Would you take a separate doc patch making it explicit that `units` names the unit of the scaled value?

NishadSaraf:
> The header is what misled me. `@unitm: Translates value member variables into the correct unit via (pow(10, unitm) * value)` never says what `units` is relative to. Would you take a separate doc patch making it explicit that `units` names the unit of the scaled value?

Agreed on the wording; making `units` explicit that it names the unit of the scaled value `(pow(10, unitm) * value)` resolves the ambiguity cleanly. Since `unitm` is released UAPI, the values don't move, so this is purely a comment fix.

Rather than a separate doc patch, could you fold it into this same patch? It's touching the exact struct we're already discussing, so keeping the clarification with it keeps the context together and avoids a standalone one-line-comment patch.

While you're at it, please add your `Signed-off-by` and rebase onto the latest tip. Happy to re-review once it's in. Thanks!

atassis:
Did the doc patch, too

NishadSaraf (approved):

PR #1532, opened 2026-07-27
accel/amdxdna: annotate PF/VF dev_lock nesting for lockdep

NishadSaraf:
pci_enable_sriov() is called from amdxdna_sriov_configure() with the PF's dev_lock held and synchronously probes each VF, whose amdxdna_probe() takes that VF's own dev_lock. Both are initialized from the same drmm_mutex_init() call site, so they share one lock class and lockdep misreports the strictly hierarchical PF -> VF acquisition as "possible recursive locking".

The PF and VF dev_locks are distinct instances and a VF never takes the PF's dev_lock, so there is no real deadlock. Place the VF instance in a deeper lockdep subclass to express the valid nesting.

### DIFF
--- a/drivers/accel/amdxdna/amdxdna_pci_drv.c
+++ b/drivers/accel/amdxdna/amdxdna_pci_drv.c
@@ -458,7 +458,18 @@ static int amdxdna_probe(struct pci_dev *pdev, const struct pci_device_id *id)
 	if (ret)
 		return ret;
 
-	drmm_mutex_init(ddev, &xdna->dev_lock);
+	ret = drmm_mutex_init(ddev, &xdna->dev_lock);
+	if (ret)
+		return ret;
+
+	/*
+	 * A VF's dev_lock is acquired in probe() while the PF still holds its
+	 * own dev_lock across pci_enable_sriov(). They are distinct instances
+	 * sharing one lock class, so nest the VF at a deeper subclass to keep
+	 * lockdep from misreporting the hierarchical PF -> VF order.
+	 */
+	lockdep_set_subclass(&xdna->dev_lock,
+			     pdev->is_virtfn ? SINGLE_DEPTH_NESTING : 0);
 	init_rwsem(&xdna->notifier_lock);
 	INIT_LIST_HEAD(&xdna->client_list);
 	ida_init(&xdna->hwctx_ida);


xdavidz (commented):
I don't understand what issue is fixed by this?
Can you please elaborate?

xdavidz:
> I don't understand what issue is fixed by this? Can you please elaborate?

@houlz0507 do you have any thoughts on this? Basically it makes one of the locking checking tool happy, but it is not a real bug.

drivers/accel/amdxdna/amdxdna_pci_drv.c
```diff
@@ -457,7 +460,18 @@ static int amdxdna_probe(struct pci_dev *pdev, const struct pci_device_id *id)
...
+		return ret;
+
+	/*
+	 * A VF's dev_lock is acquired in probe() while the PF still holds its
+	 * own dev_lock across pci_enable_sriov(). They are distinct instances
+	 * sharing one lock class, so nest the VF at a deeper subclass to keep
+	 * lockdep from misreporting the hierarchical PF -> VF order.
+	 */
+	if (pdev->is_virtfn)
+		lockdep_set_subclass(&xdna->dev_lock, AMDXDNA_DEV_LOCK_SUBCLASS_VF);
```
houlz0507: It should be:
lockdep_set_subclass(&xdna->dev_lock, pdev->is_virtfn ? SINGLE_DEPTH_NESTING : 0);

NishadSaraf: Done, updated to use the standard macro:

```c
lockdep_set_subclass(&xdna->dev_lock,
		     pdev->is_virtfn ? SINGLE_DEPTH_NESTING : 0);
```

This drops the custom `AMDXDNA_DEV_LOCK_SUBCLASS_VF` define in favor of `SINGLE_DEPTH_NESTING` and applies the annotation unconditionally (subclass 0 for the PF, which is the default).

NishadSaraf:
@xdavidz to elaborate: this is not fixing a real deadlock, it only silences a lockdep false-positive that shows up when running a `CONFIG_PROVE_LOCKING` (debug) kernel with SR-IOV enabled.

The sequence is:

1. `amdxdna_sriov_configure()` takes the **PF**'s `dev_lock` and calls `pci_enable_sriov()`.
2. `pci_enable_sriov()` synchronously probes each **VF**, and `amdxdna_probe()` takes that **VF**'s own `dev_lock`.

The PF and VF `dev_lock`s are distinct mutex instances, and a VF never takes the PF's `dev_lock`, so there is no real recursion or deadlock. But both are initialized from the same `drmm_mutex_init()` call site, so lockdep assigns them the **same lock class** and reports the PF -> VF acquisition as "possible recursive locking":

```
============================================
WARNING: possible recursive locking detected
--------------------------------------------
  lock(&xdna->dev_lock);
  lock(&xdna->dev_lock);
```

Putting the VF instance in a deeper subclass (`SINGLE_DEPTH_NESTING`) tells lockdep the nesting is intentional and hierarchical, which clears the warning. On production kernels (lockdep off) `lockdep_set_subclass()` compiles to a no-op, so there is no runtime cost — it just keeps the debug/CI logs clean and avoids masking any genuine future lock-ordering bugs.

drivers/accel/amdxdna/amdxdna_pci_drv.c
```diff
@@ -458,7 +458,18 @@ static int amdxdna_probe(struct pci_dev *pdev, const struct pci_device_id *id)
...
+		return ret;
+
+	/*
+	 * A VF's dev_lock is acquired in probe() while the PF still holds its
+	 * own dev_lock across pci_enable_sriov(). They are distinct instances
+	 * sharing one lock class, so nest the VF at a deeper subclass to keep
+	 * lockdep from misreporting the hierarchical PF -> VF order.
+	 */
+	lockdep_set_subclass(&xdna->dev_lock,
+			     pdev->is_virtfn ? SINGLE_DEPTH_NESTING : 0);
```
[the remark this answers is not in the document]
NishadSaraf: This suggestion is incorrect and would actually break the build. `lockdep_set_subclass(lock, sub)` takes the lock object and dereferences `->dep_map` internally:

```c
#define lockdep_set_subclass(lock, sub)					\
	lockdep_init_map_type(&(lock)->dep_map, (lock)->dep_map.name, ...)
```

So `&xdna->dev_lock` (the `struct mutex *`) is the correct argument — this matches the canonical usage across the tree (e.g. `lockdep_set_subclass(&rq->lock, ...)`). Passing `&xdna->dev_lock.dep_map` would expand to `dep_map.dep_map`, which does not exist. (And with `CONFIG_LOCKDEP` disabled the macro is a no-op, so there is no type mismatch in the first place.) Keeping the current code.

maxzhen (approved):

PR #1539, opened 2026-07-31
shim_test: enable VE2 edge build and board-verified test coverage

bisingha-xilinx:
Add opt-in shim_test support for the VE2 embedded platform (npu_ve2 / DT-bound amdxdna). Wire up XDNA_BUILD_SHIM_TEST CMake, minimal shim_ve2 pcie_device/pcie_vendor queries, and npu_ve2 vadd/nop test binaries.

Introduce dev_filter_npu_ve2 for positive opt-in of board-verified tests and dev_filter_xdna_not_npu_ve2 to skip unsupported cases (debug BO, cmd fence, RO map). Skip the zocl userpf device where pcie_device is absent.

Forty-seven tests pass on VEK385 edge hardware

### DIFF
--- a/CMake/native.cmake
+++ b/CMake/native.cmake
@@ -10,6 +10,11 @@
 # as amdxdna_legacy.ko.
 option(PACKAGE_LEGACY_DRIVER "Package legacy driver source" OFF)
 
+# The VE2 shim test (test/shim_test) is not part of the normal VE2 runtime
+# build. It is opt-in (e.g. built on demand by a dedicated packaging recipe)
+# so the default xrt/amdxdna build does not compile or install test artifacts.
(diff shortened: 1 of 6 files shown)


chvamshi-xilinx (approved):

maxzhen:
@bisingha-xilinx @chvamshi-xilinx , this PR added quite some test cases which only works on VE2. Are we 100% sure they are not applicable to other platforms (aie2, aie2p, aie4)?

We're trying hard to minimize the skipped test set on ALL platforms in shim test. Any test case added for just one platform requires strong justification to make sure we don't blindly add duplicated test cases and we don't miss testing on a platform with new test cases. If the effort to go through existing test cases and see if they can be enhanced for ve2 is considered too big, I'm OK for you to create a separate shim_test for ve2 only.

PR #1541, opened 2026-08-01
accel/amdxdna: flush only the requested range in amdxdna_flush_bo

atassis:
## Problem

`SYNC_BO` carries an offset and a size, but `amdxdna_flush_bo()` honours them on one path out of
four. An imported BO flushes its whole scatterlist, and the `base.pages` / `mem.pages` fallbacks
flush every page of the BO. The cost of a sync is set by the size of the buffer rather than by the
range the caller asked to maintain.

## Fix

`amdxdna_gem_obj_vmap()` already maps imported BOs through `dma_buf_vmap()`, so try the vmap path
first and leave `drm_clflush_sg()` as the fallback for an exporter that cannot vmap. Index the
page-array fallbacks from the requested offset.

## Evidence

npu4, kernel 7.1.5, 64 MiB BO, sync size swept, pinned to one core, minimum of 50 runs. Both
modules built from this tree; they differ only by this commit.

| sync size | driver-owned, before | driver-owned, after | imported, before | imported, after |
|---|---|---|---|---|
| 4 KiB | 0.6 us | 0.6 us | 1056 us | 0.7 us |
| 32 KiB | 1.1 us | 1.1 us | 1056 us | 1.1 us |
| 1 MiB | 17.1 us | 17.1 us | 1055 us | 17.1 us |
| 16 MiB | 264 us | 264 us | 1056 us | 264 us |
| 64 MiB (whole) | 1056 us | 1055 us | 1056 us | 1056 us |

The imported column before the patch is flat: a 4 KiB sync costs what a 64 MiB sync costs. After, it
tracks the driver-owned column at every size, and the whole-BO case is unchanged. The driver-owned
column is the control and does not move.

Reproducer is below, ioctl-only against `/dev/accel/accel0`.

`checkpatch.pl --strict`, freshly downloaded from mainline: clean.

## Scope

An imported BO now holds a kernel mapping from its first sync until it is freed, as a shmem BO
already does; `amdxdna_gem_obj_free()` releases it through `amdxdna_gem_vunmap()`.

The `base.pages` and `mem.pages` arms are reached only when `amdxdna_gem_vmap()` fails.

Mainline carries the same function with three arms rather than four, so the fix applies there as a
smaller diff. I'll post that version to dri-devel myself, so it does not need forwarding from here.

<details>
<summary>Reproducer (ioctl only, no kernel run, no root)</summary>

```c
/*
 * Does DRM_IOCTL_AMDXDNA_SYNC_BO cost scale with the range it is given?
 *
 * Sweeps sync sizes over one BO, for a BO the driver owns and for the same BO
 * imported into a second DRM client. Reports the first sync separately, since
 * that is where any one-time mapping cost lands.
 */
#define _GNU_SOURCE
#include <errno.h>
#include <fcntl.h>
#include <sched.h>
#include <stdio.h>
#include <string.h>
#include <sys/ioctl.h>
#include <sys/mman.h>
#include <time.h>
#include <unistd.h>

#include <libdrm/drm.h>
#include "amdxdna_accel.h"

#define BO_SIZE   (64ull << 20)
#define REPS      50

static const unsigned long long sizes[] = {
	4096, 32ull << 10, 1ull << 20, 16ull << 20, BO_SIZE,
};

static int xioctl(int fd, unsigned long req, void *arg, const char *what)
{
	int ret = ioctl(fd, req, arg);

	if (ret)
		fprintf(stderr, "%s: %s\n", what, strerror(errno));
	return ret;
}

static double now_us(void)
{
	struct timespec ts;

	clock_gettime(CLOCK_MONOTONIC, &ts);
	return ts.tv_sec * 1e6 + ts.tv_nsec / 1e3;
}

static double sync_once(int fd, unsigned int handle, unsigned long long off,
			unsigned long long size)
{
	struct amdxdna_drm_sync_bo sync = {
		.handle = handle,
		.direction = SYNC_DIRECT_TO_DEVICE,
		.offset = off,
		.size = size,
	};
	double t0 = now_us();

	if (xioctl(fd, DRM_IOCTL_AMDXDNA_SYNC_BO, &sync, "SYNC_BO"))
		return -1;
	return now_us() - t0;
}

/* minimum over REPS, which is the figure least polluted by scheduling noise */
static double sync_min(int fd, unsigned int handle, unsigned long long off,
		       unsigned long long size)
{
	double best = 1e12;

	for (int i = 0; i < REPS; i++) {
		double dt = sync_once(fd, handle, off, size);

		if (dt < 0)
			return -1;
		if (dt < best)
			best = dt;
	}
	return best;
}

static void sweep(const char *label, int fd, unsigned int handle)
{
	printf("%s\n", label);
	printf("  first sync (4 KiB): %8.1f us\n", sync_once(fd, handle, 0, 4096));
	for (unsigned i = 0; i < sizeof(sizes) / sizeof(sizes[0]); i++)
		printf("  %8llu KiB %10.1f us\n", sizes[i] >> 10,
		       sync_min(fd, handle, 0, sizes[i]));
}

int main(void)
{
	cpu_set_t set;
	CPU_ZERO(&set);
	CPU_SET(2, &set);
	if (sched_setaffinity(0, sizeof(set), &set))
		perror("sched_setaffinity");

	int fd_a = open("/dev/accel/accel0", O_RDWR);
	int fd_b = open("/dev/accel/accel0", O_RDWR);

	if (fd_a < 0 || fd_b < 0) {
		perror("open /dev/accel/accel0");
		return 1;
	}

	struct amdxdna_drm_create_bo create = {
		.size = BO_SIZE,
		.type = AMDXDNA_BO_SHARE,
	};
	if (xioctl(fd_a, DRM_IOCTL_AMDXDNA_CREATE_BO, &create, "CREATE_BO"))
		return 1;

	struct amdxdna_drm_get_bo_info info = { .handle = create.handle };
	if (xioctl(fd_a, DRM_IOCTL_AMDXDNA_GET_BO_INFO, &info, "GET_BO_INFO"))
		return 1;

	void *va = mmap(NULL, BO_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED,
			fd_a, info.map_offset);
	if (va == MAP_FAILED) {
		perror("mmap");
		return 1;
	}
	memset(va, 0xa5, BO_SIZE);	/* fault the pages in and dirty them */

	struct drm_prime_handle exp = { .handle = create.handle, .flags = DRM_RDWR };
	if (xioctl(fd_a, DRM_IOCTL_PRIME_HANDLE_TO_FD, &exp, "PRIME_HANDLE_TO_FD"))
		return 1;

	struct drm_prime_handle imp = { .fd = exp.fd };
	if (xioctl(fd_b, DRM_IOCTL_PRIME_FD_TO_HANDLE, &imp, "PRIME_FD_TO_HANDLE"))
		return 1;

	printf("BO %llu MiB, min of %d\n", BO_SIZE >> 20, REPS);
	sweep("driver-owned BO", fd_a, create.handle);
	sweep("same BO, imported", fd_b, imp.handle);

	munmap(va, BO_SIZE);
	close(exp.fd);
	close(fd_b);
	close(fd_a);
	return 0;
}
```

</details>

### DIFF
--- a/drivers/accel/amdxdna/amdxdna_gem.c
+++ b/drivers/accel/amdxdna/amdxdna_gem.c
@@ -1349,6 +1349,7 @@ int amdxdna_drm_get_bo_info_ioctl(struct drm_device *dev, void *data, struct drm
 
 static int amdxdna_flush_bo(struct amdxdna_gem_obj *abo, u64 offset, u64 size)
 {
+	unsigned long first, nr_pages;
 	u64 end;
 
 	if (offset >= abo->mem.size)
@@ -1358,14 +1359,17 @@ static int amdxdna_flush_bo(struct amdxdna_gem_obj *abo, u64 offset, u64 size)
 		return -EINVAL;
 
 	size = min(abo->mem.size, end) - offset;
-	if (is_import_bo(abo))
-		drm_clflush_sg(abo->base.sgt);
-	else if (amdxdna_gem_vmap(abo))
+	first = offset >> PAGE_SHIFT;
+	nr_pages = (PAGE_ALIGN(offset + size) >> PAGE_SHIFT) - first;
+
+	if (amdxdna_gem_vmap(abo))
 		drm_clflush_virt_range(amdxdna_gem_vmap(abo) + offset, size);
+	else if (is_import_bo(abo))
+		drm_clflush_sg(abo->base.sgt);
 	else if (abo->base.pages)
-		drm_clflush_pages(abo->base.pages, abo->mem.size >> PAGE_SHIFT);
+		drm_clflush_pages(&abo->base.pages[first], nr_pages);
 	else if (abo->mem.pages)
-		drm_clflush_pages(abo->mem.pages, abo->mem.nr_pages);
+		drm_clflush_pages(&abo->mem.pages[first], nr_pages);
 	else
 		return -EINVAL;
 
--- a/drivers/accel/amdxdna/amdxdna_gem.c
+++ b/drivers/accel/amdxdna/amdxdna_gem.c
@@ -211,13 +211,11 @@ void amdxdna_gem_destroy_obj(struct amdxdna_gem_obj *abo)
 }
 
 /*
- * Obtains a kernel virtual address on the BO (usually of small size).
- * The mapping is established on the first call and stays valid until
- * amdxdna_gem_vunmap() is called.
+ * Returns the error rather than logging it, for a caller that has a fallback:
+ * an imported BO whose exporter implements no vmap op fails on every call.
  */
-void *amdxdna_gem_vmap(struct amdxdna_gem_obj *abo)
+static void *amdxdna_gem_vmap_try(struct amdxdna_gem_obj *abo)
 {
-	struct amdxdna_dev *xdna = to_xdna_dev(to_gobj(abo)->dev);
 	struct iosys_map map = IOSYS_MAP_INIT_VADDR(NULL);
 	int ret;
 
@@ -230,11 +228,27 @@ void *amdxdna_gem_vmap(struct amdxdna_gem_obj *abo)
 	if (!abo->mem.kva) {
 		ret = drm_gem_vmap(to_gobj(abo), &map);
 		if (ret)
-			XDNA_ERR(xdna, "Vmap bo failed, ret %d", ret);
-		else
-			abo->mem.kva = map.vaddr;
+			return ERR_PTR(ret);
+		abo->mem.kva = map.vaddr;
 	}
 	return abo->mem.kva;
+}
+
+/*
+ * Obtains a kernel virtual address on the BO (usually of small size).
+ * The mapping is established on the first call and stays valid until
+ * amdxdna_gem_vunmap() is called.
+ */
+void *amdxdna_gem_vmap(struct amdxdna_gem_obj *abo)
+{
+	struct amdxdna_dev *xdna = to_xdna_dev(to_gobj(abo)->dev);
+	void *kva = amdxdna_gem_vmap_try(abo);
+
+	if (IS_ERR(kva)) {
+		XDNA_ERR(xdna, "Vmap bo failed, ret %ld", PTR_ERR(kva));
+		return NULL;
+	}
+	return kva;
 }
 
 /*
@@ -1350,6 +1364,7 @@ int amdxdna_drm_get_bo_info_ioctl(struct drm_device *dev, void *data, struct drm
 static int amdxdna_flush_bo(struct amdxdna_gem_obj *abo, u64 offset, u64 size)
 {
 	unsigned long first, nr_pages;
+	void *kva;
 	u64 end;
 
 	if (offset >= abo->mem.size)
@@ -1362,8 +1377,9 @@ static int amdxdna_flush_bo(struct amdxdna_gem_obj *abo, u64 offset, u64 size)
 	first = offset >> PAGE_SHIFT;
 	nr_pages = (PAGE_ALIGN(offset + size) >> PAGE_SHIFT) - first;
 
-	if (amdxdna_gem_vmap(abo))
-		drm_clflush_virt_range(amdxdna_gem_vmap(abo) + offset, size);
+	kva = amdxdna_gem_vmap_try(abo);
+	if (!IS_ERR(kva))
+		drm_clflush_virt_range(kva + offset, size);
 	else if (is_import_bo(abo))
 		drm_clflush_sg(abo->base.sgt);
 	else if (abo->base.pages)


drivers/accel/amdxdna/amdxdna_gem.c
```diff
@@ -1358,14 +1359,17 @@ static int amdxdna_flush_bo(struct amdxdna_gem_obj *abo, u64 offset, u64 size)
 		return -EINVAL;
 
 	size = min(abo->mem.size, end) - offset;
-	if (is_import_bo(abo))
-		drm_clflush_sg(abo->base.sgt);
-	else if (amdxdna_gem_vmap(abo))
+	first = offset >> PAGE_SHIFT;
+	nr_pages = (PAGE_ALIGN(offset + size) >> PAGE_SHIFT) - first;
+
+	if (amdxdna_gem_vmap(abo))
```
houlz0507: What will happen if the BO is imported and exporter does not implement vmap? Will amdxdna_gem_vmap() generate unwanted error message? And the next line calls amdxdna_gem_vmap() and generate the error message again.

maxzhen:
I believe the BO flush is now implemented in SHIM, not going through driver any more. @atassis , are you using driver's IOCTL directly without going through SHIM?

drivers/accel/amdxdna/amdxdna_gem.c
atassis: Correctness is unaffected: `amdxdna_gem_vmap()` returns NULL on failure, so it falls through to `is_import_bo()` and still gets `drm_clflush_sg()`.

I measured that rather than arguing it, since the arm had never run. No exporter I can reach lacks `.vmap` (udmabuf, system_heap and vgem all implement it), so I forced `dma_buf_vmap()` to return -EINVAL, which is what it returns when the op is NULL. 64 MiB BO, sync size swept, imported column:

| sync size | vmap ok | vmap forced to fail |
|---|---|---|
| 4 KiB | 0.6 us | 1058.5 us |
| 1 MiB | 17.1 us | 1058.8 us |
| 64 MiB | 1056.5 us | 1060.0 us |

Flat at every size is the whole-sgt flush, so the fallback is reached and behaves as before. No ioctl errors, and the driver-owned control column does not move.

On the log: 251 lines for 251 syncs, one per `SYNC_BO`, because nothing caches the failure. Not twice, though. The second call sits on the line that only runs when the first returned non-NULL, so it never logs. The double call was real; the double log is not reachable.

Fixed in 538fbb7. `amdxdna_gem_vmap_try()` returns the error, `amdxdna_gem_vmap()` is that plus the log, and the flush probes with the quiet one. Same injection on top of it: 0 log lines, imported column still flat. I left the failure uncached, since the retry is a `dma_resv` lock and a NULL-op test in front of a whole-scatterlist flush.

atassis:
No. I go through `xrt::bo::sync`, and you are right about where that lands: `buffer::sync()` does the userspace `clflush_data()` unless `Debug.force_driver_sync` is set, default false (`src/shim/buffer.cpp:167`). The reproducer is ioctl-only to isolate the driver path.

What keeps it worth fixing is that SYNC_BO is uAPI on the accel node, and `drivers/accel/amdxdna` carries this function into mainline, where any DRM client can reach it rather than only XRT.

If SYNC_BO is deprecated and userspace flushing is the only supported path, say so and I will drop this and stop sending SYNC_BO patches. Otherwise mainline keeps a function that costs 1056 us to flush 4 KiB of an imported BO.

houlz0507 (approved):

maxzhen:
> No. I go through `xrt::bo::sync`, and you are right about where that lands: `buffer::sync()` does the userspace `clflush_data()` unless `Debug.force_driver_sync` is set, default false (`src/shim/buffer.cpp:167`). The reproducer is ioctl-only to isolate the driver path.
> 
> What keeps it worth fixing is that SYNC_BO is uAPI on the accel node, and `drivers/accel/amdxdna` carries this function into mainline, where any DRM client can reach it rather than only XRT.
> 
> If SYNC_BO is deprecated and userspace flushing is the only supported path, say so and I will drop this and stop sending SYNC_BO patches. Otherwise mainline keeps a function that costs 1056 us to flush 4 KiB of an imported BO.

No, we can't remove it since it's part of user/kernel interface. I was just curious about how you discovered it. Thanks for the fix.

maxzhen:
ok to test

atassis:
In this exact case-  zero-copy video work, not code reading. I was checking whether an amdgpu GTT buffer could be exported and imported into amdxdna and written directly by the array - it can, which is the primitive for decode → NPU → display without host round-trips. That put imported BOs on a hot path, so I swept what a per-frame sync costs. The imported column came back flat at 1056 us for a 64 MiB BO at every size from 4 KiB up, and a flat curve meant the range was not reaching anything. Reading `amdxdna_flush_bo()` then showed `is_import_bo()` being tested ahead of the vmap arm.

In general- I am trying to implement my own xdna backend for general purpose, have installed whole stack of related repos (from IRON/mlir-aie and down to xdna-driver), and then if smth is not allowing me to build or do things (it hangs, throws and else), I am looking for an issue on another level. 
I accumulate in local KB all the knowledge, measurements and else and then try to achieve a new result each time plus to learn smth new for myself.

PR #1554, opened 2026-08-04
accel/amdxdna: ve2: Align AIE partition reinit logic with legacy driver

rajkumar-xilinx:
-Wake UP only lead column in ve2_mgmt_handshake_init() 
-Clear is_idle_due_to_context bit once consumed matching with legacy driver.
-In ve2_mgmt_schedule_cmd(): notify fw in the steady-state case (context already active, no switch pending), matching legacy's behavior. 
-The scheduler workqueue is created with a CPU-bound, single max_active, high-priority workqueue instead of create_singlethread_workqueue(). Matching legacy's create_workqueue() semantics.

Verified on HW board with an 81-sample BevF (ResNet50 + encoder + decoder) inference benchmark: default mode improves from 26.34 FPS to 28.24 FPS and ve2_perf_optimization=1 mode improves from 28.70 FPS to 30.03 FPS, now matching the legacy driver's 28.22/30.01 FPS.


rajkumar-xilinx:
All hw tests passed.

chvamshi-xilinx (approved):

PR #1556, opened 2026-08-04
Update VTD archive

hlaccabu:


NishadSaraf (changes requested):
Hi @hlaccabu, Thanks for the PR. Better to be in lockstep with the Windows driver. Can you please pin this to c579b12aba708d81b599bb2c9105ba68f3e721e6 instead?

NishadSaraf (approved):

PR #1559, opened 2026-08-05
build: use cmake on fedora, fixes #1558

### LINKED ISSUE #1558: Fedora 44 uses `cmake` instead of `cmake3` in build/build.sh
### Summary

On Fedora 44, `build/build.sh` assumes that Fedora uses `cmake3`:

```sh
if [[ $OSDIST == "centos" ]] || [[ $OSDIST == "amzn" ]] || [[ $OSDIST == "rhel" ]] || [[ $OSDIST == "fedora" ]]; then
    CMAKE=cmake3
fi
```

However, Fedora 44 ships CMake 4 as `cmake` and no longer provides a `cmake3` executable.

As a result, the build fails with:

```text
cmake3 is not installed
```

Using the existing `cmake` executable on Fedora 44 builds successfully.

### Related

I also noticed a similar report in #1323.

Would you be open to removing the Fedora-specific `cmake3` override while keeping the existing behavior for CentOS, RHEL, and Amazon Linux?

Ill make a PR with the change i made that removes this check

ahhyoushh:
Fixes #1558

Fedora no longer provides a `cmake3` executable and ships CMake 4 as `cmake`.

This change removes the Fedora-specific `cmake3` override while preserving the existing behavior for CentOS, RHEL, and Amazon Linux.

Verified on Fedora 44.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -330,7 +330,7 @@ XBUTIL_VALIDATE_BINS_DIR=$DOWNLOAD_BINS_DIR/download_raw/xbutil_validate/bins
 
 # Sanity check
 if [[ $CMAKE_MAJOR_VERSION != 3 ]]; then
-    if [[ $OSDIST == "centos" ]] || [[ $OSDIST == "amzn" ]] || [[ $OSDIST == "rhel" ]] || [[ $OSDIST == "fedora" ]]; then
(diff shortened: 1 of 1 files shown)


ahhyoushh:
Superseded by #1560, which implements executable detection (`cmake`/`cmake3`) as suggested in #1558. Closing this PR.

PR #1560, opened 2026-08-05
build: detect available cmake executable

### LINKED ISSUE #1558: Fedora 44 uses `cmake` instead of `cmake3` in build/build.sh

ahhyoushh:
Replace distribution-specific CMake selection with executable detection.

The build script now:

- uses `cmake` when it provides CMake 3 or newer,
- falls back to `cmake3` if needed,
- reports an error if neither is available.

This fixes Fedora 44, where CMake 4 is provided as `cmake` and no `cmake3` executable exists.

Fixes #1558

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -329,14 +329,22 @@ DOWNLOAD_BINS_DIR=./amdxdna_bins
 XBUTIL_VALIDATE_BINS_DIR=$DOWNLOAD_BINS_DIR/download_raw/xbutil_validate/bins
 
 # Sanity check
-if [[ $CMAKE_MAJOR_VERSION != 3 ]]; then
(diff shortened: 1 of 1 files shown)


maxzhen:
ok to test

PR #1564, opened 2026-08-05
accel/amdxdna: clear warnings of configure_kernel.sh probes

donwalkarsoham:
make configure_kernel.sh probes warning-clean for CONFIG_WERROR kernels

The kernel 7.2-rc6 has CONFIG_WERROR=y due to which the config files warnings are converted to errors leading to the failure of probes in the configure_kernel.sh. Due to which certain macros are not defined while they should have been defined leading to build errors.

Solution -
Cleaning the probes of any warnings.

Test -
Successfully built the xdna driver with kernel 7.2.0-rc6.

### DIFF
--- a/drivers/accel/tools/configure_kernel.sh
+++ b/drivers/accel/tools/configure_kernel.sh
@@ -181,7 +181,7 @@ int main(void)
 		.num_rqs = DRM_SCHED_PRIORITY_COUNT,
 	};
 
-	return 0;
+	return b.num_rqs;
 }
 EOF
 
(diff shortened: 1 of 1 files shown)


drivers/accel/tools/configure_kernel.sh
```diff
@@ -313,8 +314,7 @@ try_compile HAVE_xen_phy_dma_ops << 'EOF'
 #include <xen/phy-dma-ops.h>
 int main(void)
 {
-	const struct dma_map_ops *a = &xen_phy_dma_ops;
-	return 0;
+	return sizeof(xen_phy_dma_ops) != 0;
```
xdavidz, line 317: No, this is not what we want still!
we want to use configure_kernel.sh to work with all possible kernel versions by compile it and find erros. You bascially changed the code to not report errors, this is wrong.

Please close this PR.

xdavidz (changes requested):

PR #1567, opened 2026-08-06
accel/amdxdna: hold CU BO references for the life of the context

### LINKED ISSUE #1566: amdxdna: PM resume fails when a hwctx restart looks up an already-freed CU BO
## Summary

With more than one client dispatching, a test intermittently dies with an uncaught XRT exception:

```
terminate called after throwing an instance of 'xrt_core::system_error'
  what():  DRM_IOCTL_AMDXDNA_GET_INFO IOCTL failed (err=-22): Invalid argument
```

`GET_INFO` is incidental. `aie2_get_info()` calls `amdxdna_pm_resume_get_locked()` before it reaches
its `switch (args->param)`, so a failed runtime-PM resume is reported against whichever ioctl
happened to trigger the resume. dmesg gives the real chain:

```
amdxdna 0000:c5:00.1: [drm] *ERROR* aie2_config_cu: Lookup GEM object failed
amdxdna 0000:c5:00.1: [drm] *ERROR* aie2_hwctx_restart: Config cu failed, ret -22
amdxdna 0000:c5:00.1: [drm] *ERROR* amdxdna_pm_resume_get: Resume failed: -22
amdxdna 0000:c5:00.1: [drm] device is already started
```

On resume the driver restarts hardware contexts and re-configures their CUs.
`aie2_config_cu()` does

```c
gobj = drm_gem_object_lookup(hwctx->client->filp, cu->cu_bo);
if (!gobj) {
        XDNA_ERR(xdna, "Lookup GEM object failed");
        return -EINVAL;
}
```

and gets NULL -- the CU BO backing a context being restarted is already gone. So a context is
outliving the buffer object that `hwctx_restart` needs, and the resume path has no way to skip it.

## Reproduction

Run N copies of an already-built example concurrently, repeatedly. Each is an ordinary
xclbin + insts run; nothing exotic.

Rates measured on this box, counting a missing `PASS!` from the example's own self-check:

| concurrent clients | runs | failures |
|---|---|---|
| 1 | 150 | 0 |
| 2 | 190 | 2 |
| 4 | 150 | 0 |
| 8 | 135 | 0 |
| 12 | 135 | 0 |

Both failures are dmesg-correlated to the second with the aborting process.

## Observations

- **At least two clients are required.** 0 in 150 single-client runs.
- **It is not monotonic in concurrency.** 0 in 420 runs at 4, 8 and 12 clients, where the device
  stays continuously busy and so does not runtime-suspend.
- The device does autosuspend: `power/control=auto`, `power/autosuspend_delay_ms=5000`, and
  `runtime_status` reads `suspended` throughout an idle window.

That pattern suggests the window needs an idle period (so a resume actually happens) plus a second
client (so a context can be going away while resume restarts it). I could not confirm it: a
deliberately idled arm produced 0 failures in 50 runs, but at the observed ~1% rate that arm expects
~0.5 events, so it is underpowered rather than evidence against.

## Environment

```
NPU     c5:00.1 [1022:17f0] rev 10  Strix/Krackan/Strix Halo NPU
driver  in-kernel amdxdna (accel driver 0.8.0), /lib/modules/7.1.5-1-cachyos/kernel/drivers/accel/amdxdna/
kernel  7.1.5-1-cachyos
XRT     2.21.75
fw      amdnpu/17f0_10/npu_7.sbin
```

The code path is unchanged at `amd/xdna-driver` `main` as of 2026-08-05
(`drivers/accel/amdxdna/aie2_message.c`), so this is not something already fixed there.

Closest existing reports I found are #629 (create_ctx slower with auto suspend/resume) and #18
(suspend and hibernate), neither of which is this race.

I have a reproduction and the trace but no fix, so this is an issue rather than a PR. Happy to test
a patch or gather more data on request -- in particular I can run a higher-powered idle-window arm
if that would help confirm the trigger.

atassis:
Fixes #1566.

## Problem

`hwctx->cus` stores userspace GEM *handles*, and `aie2_config_cu()` re-resolved them against
`hwctx->client->filp` on every call. It is called a second time from `aie2_hwctx_restart()` on
runtime-PM resume, long after the config ioctl returned. A client that has since dropped the handle
makes that lookup fail, and the failure takes the whole resume down:

```
amdxdna 0000:c5:00.1: [drm] *ERROR* aie2_config_cu: Lookup GEM object failed
amdxdna 0000:c5:00.1: [drm] *ERROR* aie2_hwctx_restart: Config cu failed, ret -22
amdxdna 0000:c5:00.1: [drm] *ERROR* amdxdna_pm_resume_get: Resume failed: -22
```

`aie2_get_info()` calls `amdxdna_pm_resume_get_locked()` before it reaches its
`switch (args->param)`, so the -EINVAL is reported against whichever ioctl happened to trigger the
resume. From userspace it looks like an unrelated GET_INFO failure, and XRT raises it as an uncaught
`xrt_core::system_error`, so the application aborts:

```
terminate called after throwing an instance of 'xrt_core::system_error'
  what():  DRM_IOCTL_AMDXDNA_GET_INFO IOCTL failed (err=-22): Invalid argument
```

## Fix

Resolve and validate each CU BO once in `aie2_hwctx_cu_config()` and hold a reference until the
context is torn down. `aie2_config_cu()` then uses the held objects, so it no longer depends on a
handle still being live in the client's file table.

This matches how the same context already treats its other buffers: `priv->heap` and
`priv->cmd_buf[]` are both referenced for the life of the hwctx and released in `aie2_hwctx_fini()`.

The `num_cus != num_cu_bos` check in `aie2_config_cu()` is a guard against a future path reaching it
without the config step having run; it is not expected to fire.

## Test

Built clean (no new warnings) as an external module against 7.1.5 with `LLVM=1`, on the hardware the
bug was found on:

```
NPU     c5:00.1 [1022:17f0] rev 10  Strix/Krackan/Strix Halo NPU
XRT     2.21.75
fw      amdnpu/17f0_10/npu_7.sbin
```

The reproduction from #1566 is two clients dispatching with an idle window between batches, which
lets the device autosuspend (`autosuspend_delay_ms=5000`) so a resume actually runs. Rates measured
before this patch, counting a missing self-check PASS:

| concurrent clients | runs | failures |
|---|---|---|
| 1 | 150 | 0 |
| 2 | 190 | 2 |
| 4 | 150 | 0 |
| 8 | 135 | 0 |
| 12 | 135 | 0 |

Both failures were dmesg-correlated to the second with the aborting process.

Built as an external module against 7.1.5 and loaded in place of the in-kernel driver, both with and
without this patch. Same tree, same toolchain, same kernel, same workload, so the patch is the only
variable:

| module | GET_INFO occurrences | 2-client jobs |
|---|---|---|
| `main` unpatched | **2** | 300 |
| `main` + this patch | **0** | 870 |

Fisher one-sided p = 0.066. I would not lean on that number -- two events is a thin baseline, and the
argument for the patch is that the failing lookup is gone from the resume path, not that the rate
moved. The useful part of the control is simply that the bug reproduces on current `main`, so this is
not an artifact of the older in-kernel driver the failures were first seen on.

Counting occurrences rather than test failures is deliberate: my harness retries this signature, so
both control hits recovered on the second attempt and the test still reported PASS. The driver-level
event is what the table counts.

Throughput is unchanged with the references held: 15 concurrent runs over 5 designs take 14.9s at 8
clients patched against 16.2s before, and 66.7s against 68.2s at one client.

Unrelated, but seen at probe on 7.1.5 with both builds, so it is not from this change:

```
amdxdna_dpt_publish: fw_log: Failed to allocate buffer: -12
amdxdna_dpt_init: Failed to enable FW logging: -12
```

with a stack trace through `amdxdna_alloc_msg_buff` / `amdxdna_fw_log_init`. Happy to open a separate
issue if it is not already known.

Happy to change where the reference is held if `hwctx->priv` is the wrong home for it.

### DIFF
--- a/drivers/accel/amdxdna/aie2_ctx.c
+++ b/drivers/accel/amdxdna/aie2_ctx.c
@@ -858,12 +858,18 @@ static void aie2_hwctx_put_cu_bos(struct amdxdna_hwctx *hwctx)
 {
 	u32 i;
 
-	for (i = 0; i < hwctx->priv->num_cu_bos; i++)
+	if (!hwctx->priv->cu_bos)
+		return;
+
+	/* A partially filled array is left zeroed past the failed lookup. */
+	for (i = 0; i < hwctx->cus->num_cus; i++) {
+		if (!hwctx->priv->cu_bos[i])
+			break;
 		drm_gem_object_put(to_gobj(hwctx->priv->cu_bos[i]));
+	}
 
 	kfree(hwctx->priv->cu_bos);
 	hwctx->priv->cu_bos = NULL;
-	hwctx->priv->num_cu_bos = 0;
 }
 
 void aie2_hwctx_fini(struct amdxdna_hwctx *hwctx)
@@ -922,9 +928,9 @@ static int aie2_config_cu_resp_handler(void *handle, void __iomem *data, size_t 
 /*
  * Resolve each configured CU BO once and keep a reference for the life of the
  * context. hwctx->cus holds userspace GEM handles, which stop resolving as soon
- * as the client drops them -- but aie2_config_cu() runs again on every resume,
- * so a dropped handle turned into a failed hwctx restart and left the device
- * unable to resume at all.
+ * as the client drops them, while aie2_config_cu() runs again on every resume.
+ * A dropped handle would fail the hwctx restart and leave the device unable to
+ * resume at all.
  */
 static int aie2_hwctx_hold_cu_bos(struct amdxdna_hwctx *hwctx)
 {
@@ -944,8 +950,7 @@ static int aie2_hwctx_hold_cu_bos(struct amdxdna_hwctx *hwctx)
 		return -EINVAL;
 	}
 
-	hwctx->priv->cu_bos = kcalloc(num_cus, sizeof(*hwctx->priv->cu_bos),
-				      GFP_KERNEL);
+	hwctx->priv->cu_bos = kzalloc_objs(*hwctx->priv->cu_bos, num_cus);
 	if (!hwctx->priv->cu_bos)
 		return -ENOMEM;
 
@@ -967,7 +972,6 @@ static int aie2_hwctx_hold_cu_bos(struct amdxdna_hwctx *hwctx)
 
 		/* Reference retained; released by aie2_hwctx_put_cu_bos(). */
 		hwctx->priv->cu_bos[i] = abo;
-		hwctx->priv->num_cu_bos = i + 1;
 	}
 
 	return 0;
--- a/drivers/accel/amdxdna/aie2_ctx.c
+++ b/drivers/accel/amdxdna/aie2_ctx.c
@@ -854,6 +854,18 @@ free_priv:
 	return ret;
 }
 
+static void aie2_hwctx_put_cu_bos(struct amdxdna_hwctx *hwctx)
+{
+	u32 i;
+
+	for (i = 0; i < hwctx->priv->num_cu_bos; i++)
+		drm_gem_object_put(to_gobj(hwctx->priv->cu_bos[i]));
+
+	kfree(hwctx->priv->cu_bos);
+	hwctx->priv->cu_bos = NULL;
+	hwctx->priv->num_cu_bos = 0;
+}
+
 void aie2_hwctx_fini(struct amdxdna_hwctx *hwctx)
 {
 	struct amdxdna_dev *xdna;
@@ -893,6 +905,7 @@ void aie2_hwctx_fini(struct amdxdna_hwctx *hwctx)
 	drm_gem_object_put(to_gobj(hwctx->priv->heap));
 
 	mutex_destroy(&hwctx->priv->io_lock);
+	aie2_hwctx_put_cu_bos(hwctx);
 	kfree(hwctx->col_list);
 	kfree(hwctx->priv);
 	kfree(hwctx->cus);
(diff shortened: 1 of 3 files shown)


maxzhen:
ok to test

houlz0507:
Just try to understand more the motivation. Does user intentionally remove cu bo handle?  Or this is just for a negative test?

atassis:
@houlz0507 Neither.
I have met this bug while running the mlir-aie lit suite with concurrent NPU dispatch (https://github.com/Xilinx/mlir-aie/pull/3506). Ordinary XRT usage, nothing in the tests closes a handle on purpose.
The handle dies anyway, on exit. `drm_file_free()` calls `drm_gem_release()` before `postclose`, and `postclose` is where `amdxdna_client_cleanup()` does the `list_del(&client->node)` and destroys the hwctxs. Between those two the client's handle table is empty while its contexts are still on `xdna->client_list`. If another client's ioctl resumes the device in that window, `aie2_hwctx_resume()` restarts a context whose CU BOs are already gone. Both sides serialize on `dev_lock`, which is why it takes two clients plus an idle window, and why it never fired at 8 or 12 clients where the device never autosuspends.
So no userspace can avoid it by holding the handle longer.

drivers/accel/amdxdna/aie2_ctx.c
```diff
@@ -906,6 +919,64 @@ static int aie2_config_cu_resp_handler(void *handle, void __iomem *data, size_t
...
+	 * num_cus is only limited by the page-sized config buffer, so an
+	 * over-limit request would otherwise take a reference per CU before
+	 * anything rejected it.
+	 */
+	if (num_cus > MAX_NUM_CUS) {
+		XDNA_DBG(xdna, "Exceed maximum CU %d", MAX_NUM_CUS);
+		return -EINVAL;
+	}
+
+	hwctx->priv->cu_bos = kcalloc(num_cus, sizeof(*hwctx->priv->cu_bos),
```
houlz0507: kzalloc_objs should be used.

drivers/accel/amdxdna/aie2_ctx.c
```diff
@@ -854,6 +854,18 @@ int aie2_hwctx_init(struct amdxdna_hwctx *hwctx)
 	return ret;
 }
 
+static void aie2_hwctx_put_cu_bos(struct amdxdna_hwctx *hwctx)
+{
+	u32 i;
+
+	for (i = 0; i < hwctx->priv->num_cu_bos; i++)
```
houlz0507: Maybe if (!hwctx->priv->cu_bos[i]) break?

drivers/accel/amdxdna/aie2_ctx.c
```diff
@@ -854,6 +854,18 @@ int aie2_hwctx_init(struct amdxdna_hwctx *hwctx)
...
+		drm_gem_object_put(to_gobj(hwctx->priv->cu_bos[i]));
+
+	kfree(hwctx->priv->cu_bos);
+	hwctx->priv->cu_bos = NULL;
+	hwctx->priv->num_cu_bos = 0;
```
houlz0507: Do we really need num_cu_bos considering we have hwctx->cus->num_cus already?

drivers/accel/amdxdna/aie2_message.c
```diff
@@ -618,31 +617,29 @@ int aie2_config_cu(struct amdxdna_hwctx *hwctx,
 		return -EINVAL;
 	}
 
+	if (hwctx->cus->num_cus != hwctx->priv->num_cu_bos) {
```
houlz0507: Do we need this? aie2_hwctx_hold_cu_bos() would fail and return.

houlz0507 (changes requested):

houlz0507:
> @houlz0507 Neither. I have met this bug while running the mlir-aie lit suite with concurrent NPU dispatch (https://github.com/Xilinx/mlir-aie/pull/3506). Ordinary XRT usage, nothing in the tests closes a handle on purpose. The handle dies anyway, on exit. `drm_file_free()` calls `drm_gem_release()` before `postclose`, and `postclose` is where `amdxdna_client_cleanup()` does the `list_del(&client->node)` and destroys the hwctxs. Between those two the client's handle table is empty while its contexts are still on `xdna->client_list`. If another client's ioctl resumes the device in that window, `aie2_hwctx_resume()` restarts a context whose CU BOs are already gone. Both sides serialize on `dev_lock`, which is why it takes two clients plus an idle window, and why it never fired at 8 or 12 clients where the device never autosuspends. So no userspace can avoid it by holding the handle longer.

I see. Thanks a lot for your information and fix.

atassis:
Thanks for a review, all four applied. Happy to help!

houlz0507 (approved):

maxzhen (approved):

PR #1568, opened 2026-08-06
accel/amdxdna: make the FW log ring size configurable

atassis:
## Problem

`AMDXDNA_DPT_FW_LOG_SIZE` is `SZ_4M`, an order-10 allocation made from probe. Once memory is
fragmented that block may not be formable, and FW logging stays off for the lifetime of the module
with no way to ask for a size that would fit. Probe itself is unaffected: `amdxdna_dpt_init()`
warns and returns 0.

## Fix

Two commits:

1. Keep the FW log ring size on `struct amdxdna_dev` and expose it as a `fw_log_size` debugfs node.
   A read prints the ring in use while logging is active, otherwise the size the next enable will
   ask for. A write takes effect at the next enable, or right away by republishing at the level
   already in use. `amdxdna_alloc_msg_buff()` is untouched and stays exact-or-fail.
2. Print the requested size in the allocation failure, and point at the node in the FW-log init
   warning. `Failed to allocate buffer: -12` said neither how much was asked for nor what to do.

The automatic shrink to a fixed `SZ_1M` floor from the first revision is dropped.

## Test

Built with `-Werror` per commit; `checkpatch.pl --strict` clean.

On npu4 (0000:c5:00.1), kernel 7.1.8, module built from this branch:

```
== defaults (expect fw_log_size 4194304, fw_log_level 2)
fw_log_size                  -> 4194304
fw_log_level                 -> 2

== shrink to 1M while logging is active
write fw_log_size=1048576 -> accepted
fw_log_size                  -> 1048576
fw_log_level                 -> 2

== off/on cycle keeps the configured size
write fw_log_level=0 -> accepted
write fw_log_level=2 -> accepted
fw_log_size                  -> 1048576
fw_log_level                 -> 2

== rejects (expect both rejected)
write fw_log_size=8388608 -> rejected
write fw_log_size=0 -> rejected
fw_log_size                  -> 1048576

== back to 4M
write fw_log_size=4194304 -> accepted
fw_log_size                  -> 4194304
```

The write at 1M republishes: the read-back is the new ring's size, so the allocation and the
firmware config message both went through, and the level survives it. No new dmesg output beyond
the usual firmware load and init lines.

The allocation-failure branch is not exercised here, since I cannot fragment memory on demand;
only its message changed.

## Scope

FW log only. FW trace publishes from an explicit user request rather than from probe, so it can
simply be retried later; say the word if you want the same node for it.

The node is manual by construction, so a box that loses the allocation at probe has no logging
until someone writes it. That is the case I hit (module reload in CI), and it is not covered here.

### DIFF
--- a/drivers/accel/amdxdna/amdxdna_debugfs.c
+++ b/drivers/accel/amdxdna/amdxdna_debugfs.c
@@ -11,6 +11,7 @@
 #include <drm/drm_file.h>
 #include <linux/cleanup.h>
 #include <linux/debugfs.h>
+#include <linux/mmzone.h>
 #include <linux/pm_runtime.h>
 #include <linux/seq_file.h>
 #include <linux/string.h>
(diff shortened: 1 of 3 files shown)


maxzhen:
ok to test

maxzhen:
retest this please

drivers/accel/amdxdna/aie.c
```diff
@@ -601,41 +601,77 @@ int amdxdna_get_frame_boundary_preempt_state(struct aie_device *aie,
...
+			  hdl->size, size);
+
 	return hdl;
 
 free_hdl:
 	kfree(hdl);
 	return ERR_PTR(-ENOMEM);
 }
 
+struct amdxdna_msg_buf_hdl *amdxdna_alloc_msg_buff(struct amdxdna_dev *xdna, u32 size)
```
houlz0507: Because only dpt might require this,  I would prefer to revert amdxdna_alloc_msg_buff() and put the retry loop to _atmost. Also,
     1. should dpt use smaller (e.g. 1M) as default? And make the size configurable through debugfs?
     2. maybe dpt_init failure should not fail probe?
 @NishadSaraf @maxzhen , what do you think?

maxzhen: > Because only dpt might require this, I would prefer to revert amdxdna_alloc_msg_buff() and put the retry loop to _atmost. Also, 1. should dpt use smaller (e.g. 1M) as default? And make the size configurable through debugfs? 2. maybe dpt_init failure should not fail probe? @NishadSaraf @maxzhen , what do you think?

I agree that dpt should not fail probe. It is a debug feature. When it fails to allocate memory, we can print a clear error message and reject xrt-smi's request to enable fw log/trace later.

As for the size, I don't see a difference b/w 4M and 1M or any other size. Any contig mem allocation may fail and the failure needs to be handled no matter what. 99% of cases, driver is loaded during boot. If you can't get 4M contig buffer, you may also fail to get a 1M one.

So, I'm less concerned about the size, but more about how we handle the failure when it occurs.

We may also add a debugfs to allow user to adjust the size, in the future, to allow user to retry some smaller size when it fails to obtain a 4M buffer during probe.

NishadSaraf (changes requested):
Hi @atassis, @houlz0507, @maxzhen,

One clarification on the "dpt_init should not fail probe" point: this is already the case on main. amdxdna_dpt_init() logs an XDNA_WARN on failure and unconditionally returns 0, and the probe path (aie2_pci.c) doesn't check its return value anyway — so a failed FW log allocation only warns, it never aborts device bring-up. FW_LOG stays off for the module lifetime, but the device comes up fine.

I agree with @maxzhen that the more important thing here is how we handle the allocation failure, not the buffer size. A contiguous allocation of any size can fail under fragmentation, and if you can't get 4M during boot you may well not get 1M either — so shrinking the ring doesn't fundamentally change the failure mode, it just moves the threshold.

On the size itself: while the firmware and driver design do employ corruption detection and recovery mechanisms, those are not immune to shrinking the buffer beyond a certain point. 4M is a good default for most workloads at the INFO log level, but for the scenario described here how small the buffer can safely go really depends on the workload being debugged and the log level. Because of that, I don't think an automatic shrink-on-retry to a fixed SZ_1M floor is the right knob — the "right" size isn't a single constant.

So IMO the better direction is to let the user calibrate the buffer size for their given workload and log level via a debugfs node, rather than having the driver silently pick a smaller size. That's a clean, opt-in path for a user who hit the allocation failure to retry with a size that suits their case, and it keeps amdxdna_alloc_msg_buff() untouched (addressing @houlz0507's revert request) and the exact-or-fail contract intact. If you want to take that direction, I'd suggest dropping the _atmost shrink change in favor of the debugfs-configurable size.

atassis:
Dropped the automatic shrink and the `amdxdna_alloc_msg_buff()` change, and pushed
@NishadSaraf's direction instead: a `fw_log_size` debugfs node, plus a failure message that names
the size that could not be allocated. @houlz0507's revert request is satisfied by there being
nothing left in the shared allocator to revert.

One correction, because it decides whether the node helps anyone: "if you can't get a 4M contig
buffer you may also fail to get a 1M one" holds under memory exhaustion, not under fragmentation.
An order-10 block needs two buddy-aligned order-9 blocks, recursively down; pages scattered at
orders 0-9 satisfy 1M and cannot satisfy 4M. On the box that failed, `/proc/buddyinfo` had orders
0-9 populated and order-10 empty, and 1M allocated on the spot. If smaller sizes failed too, a size
knob would be decoration.

Smaller point, same direction: the allocation that fails at probe is for
`AMDXDNA_DPT_FW_LOG_LEVEL_DEFAULT`, which is WARN rather than INFO, so a 1M ring holds a lot of
history at the level this path actually uses.

On the default size: keeping 4M. At boot the block is normally there and the depth is worth having;
the node is for the boxes where it is not.

The one case this does not cover is the one I hit, an unattended module reload with nobody at a
shell. If you want that covered later, the same node can carry a floor rather than an absolute
size, defaulting to the requested size so nothing changes unless someone sets it. Out of scope
here.

And agreed on `amdxdna_dpt_init()`: it already warns and returns 0, so probe was never at risk.

atassis:
@NishadSaraf this took your suggested direction: the `_atmost` shrink is gone, `amdxdna_alloc_msg_buff()` is untouched per @houlz0507's revert request, and the size is now calibrated through a `fw_log_size` debugfs node instead. Pushed 26 Aug - the change request predates it, so this may not have resurfaced for you. Mind taking another look?

NishadSaraf:
> @NishadSaraf this took your suggested direction: the `_atmost` shrink is gone, `amdxdna_alloc_msg_buff()` is untouched per @houlz0507's revert request, and the size is now calibrated through a `fw_log_size` debugfs node instead. Pushed 26 Aug - the change request predates it, so this may not have resurfaced for you. Mind taking another look?

@atassis, thanks for the reminder. I'll get back to you on this by Friday EOD.

NishadSaraf (approved):
Please rebase and squash the commits.

atassis:
@NishadSaraf done!

NishadSaraf (approved):

PR #1570, opened 2026-08-06
add try_compile sanity test to check "feature absent" (expected failure)

xdavidz:
and "toolchain broken" (unexpected failure).

@maxzhen I noticed on some "customized recent kernel versions", the config_kernel.h did not work well.
This ends up with that on those kernels, some toolchain are missing which causes the try_compile in configure_kernel.sh fail.
However, the current try_compile will fail through those toochain broken to still generate a config_kernel.h.

Thus, I added a sanity test to give try_compile a must pass test, if it fails, no need to continue. Thus, user should fix their building env first.

Tested the official linux kernels are not borken, only some customized kernel built by devloper may have this special issue.

### DIFF
--- a/drivers/accel/tools/configure_kernel.sh
+++ b/drivers/accel/tools/configure_kernel.sh
@@ -107,6 +107,40 @@ EOF
     rm -rf "$tmpdir"
 }
 
+# ---- Sanity-check the build toolchain ----------------------------------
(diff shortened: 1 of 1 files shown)


maxzhen (approved):
Makes a lot of sense. We do need to tell if the failure is caused by feature not present or tool chain issue.

PR #1572, opened 2026-08-07
accel/amdxdna/ve2: Capture host_time_high/low before aie_partition_uc_wakeup()

rajkumar-xilinx:
-Stamp host_time data directly into each column's handshake struct in partition-local memory. 
-The value CERT reads on boot reflects latest host_time rather than the time at which cert_setup_partition() ran, several hundred microseconds earlier, before the (comparatively slow) aie_partition_initialize() call.

### DIFF
--- a/drivers/accel/amdxdna/ve2_mgmt.c
+++ b/drivers/accel/amdxdna/ve2_mgmt.c
@@ -92,7 +92,6 @@ static void cert_setup_partition(struct amdxdna_mgmtctx *mgmtctx,
 	struct amdxdna_ctx_priv *vp = ve2_hw_priv(hwctx);
 	struct ve2_config_hwctx *cfg = NULL;
 	u64 hsa_addr = U64_MAX;
-	u64 host_time;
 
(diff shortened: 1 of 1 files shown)


rajkumar-xilinx:
Trace point timestamp details
        host.exe-53912   [002] ..... 18230.703967: __amdxdna_trace_point: XRT_PROFILING_TRACE_PARTITIO
N_DONE: ve2_mgmt_handshake_init() pid=53912 [0,1,4]
        host.exe-53912   [002] ..... 18230.703971: __amdxdna_trace_point: XRT_PROFILING_TRACE_HOST_TIM
E_SETUP: ve2_mgmt_handshake_init() pid=53912 [0,1,4]

CERT log timestamp start at:
[18230.703967522] [CERT] mpnpu base time: high 0x1094 low 0xaaa24a50

Without host_time config changes, CERT log looks like below
[0.000000423] [CERT] mpnpu base time: high 0x0 low 0x0

chvamshi-xilinx (approved):

PR #1576, opened 2026-08-08
 Enable AMD NPU (amdxdna) from a Xen domU guest over virtio

wendyliang25:
Summary

  This series brings up AMD NPU usage inside a Xen domU guest via the virtio-gpu / vxdna path. It covers the kernel dom0 enablement for guest-foreign memory, the vxdna host-side BO mapping and cross-context sharing model, guest user-pointer buffers, guest-appropriate permission
  semantics, a test-coverage relaxation, and a teardown-deadlock fix.

  What's included

  Kernel (dom0 driver)
  - accel/amdxdna: support Xen foreign guest memory for ubuf BOs — Xen foreign guest memory mapped via privcmd (MMAPBATCH) is VM_IO|VM_PFNMAP and can't be pinned with pin_user_pages(). Detect it from VMA state and borrow privcmd's local page array instead of pinning,
  tracked via amdxdna_mem.foreign so those pages are never unpinned on free. gntdev (grant) pages keep the normal pin path; the grant-revocation-before-BO-destroy hazard is documented.

  vxdna (host-side VMM component)
  -  vxdna: map host-backed BOs via map_offset and share them across contexts — mmap host-backed blob BOs through GET_BO_INFO map_offset (mirroring the native shim); make AMDXDNA_BO_SHARE device-owned on the long-lived device fd and share it across guest contexts via
  dma-buf export/import (each context imports into its own fd, getting a per-fd handle + reference); and map multi-chunk DEV_HEAP contiguously in the per-context heap arena. DEV_HEAP stays context-scoped; only SHARE is cross-context.
  - vxdna: destroy hwctx before joining the fence polling thread — fixes a teardown deadlock: the polling thread blocks in SYNCOBJ_TIMELINE_WAIT, whose fence is only signalled when the kernel tears down the hwctx — which was gated behind the join. Issue DESTROY_HWCTX
  first, then join, then destroy the syncobj; also set m_stop_polling under m_fences_lock to close a lost-wakeup window.

  shim (guest-side)
  - shim/virtio: return EACCES for guest-forbidden get_info/set_state — set_state and the privileged AIE-debug get_info queries aren't forwarded from the guest; return EACCES (was ENOTSUP) so callers degrade gracefully like on bare metal (e.g. xrt-smi validate keeps
  running in default power mode). Also fixes copy-pasted query labels.
  - shim/virtio: create userptr BOs via RESOURCE_CREATE_BLOB userptr — back a guest AMDXDNA_BO_SHARE userptr buffer with the caller's own pages by pinning the range as a virtio-gpu guest resource (VIRTGPU_BLOB_FLAG_USE_USERPTR) and having the host allocate a shared BO
  over its iovecs. The flag/struct/ioctl are defined locally (with a size + offset static_assert) so it builds against headers predating userptr support and fails cleanly on kernels/VMMs without it.

### DIFF
--- a/drivers/accel/amdxdna/amdxdna_ubuf.c
+++ b/drivers/accel/amdxdna/amdxdna_ubuf.c
@@ -11,6 +11,7 @@
 #include <linux/overflow.h>
 #include <linux/pagemap.h>
 #include <linux/vmalloc.h>
+#include <xen/xen.h>
 
 #include "amdxdna_gem.h"
 #include "amdxdna_pci_drv.h"
@@ -38,7 +39,8 @@ static void amdxdna_gem_ubuf_obj_free(struct drm_gem_object *gobj)
 
 	amdxdna_dma_unmap_bo(xdna, abo);
 	amdxdna_ubuf_unmap_dma(abo);
-	if (abo->mem.nr_pages)
+	/* Foreign pages are borrowed from the Xen mapping; never unpin them. */
+	if (abo->mem.nr_pages && !abo->mem.foreign)
 		unpin_user_pages(abo->mem.pages, abo->mem.nr_pages);
 	atomic64_sub(abo->mem.size >> PAGE_SHIFT, &abo->mem.mm->pinned_vm);
 	kvfree(abo->mem.pages);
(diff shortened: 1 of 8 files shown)


NishadSaraf:
retest this please

test/shim_test/shim_test.cpp
maxzhen, line 1896: Can we also enable this test case?

maxzhen (dismissed):

test/shim_test/shim_test.cpp
wendyliang25: Thanks @maxzhen, however, in case of XEN DomU, there is implementation to directly map domU pages as foreign page in dom0. In this case, the qemu xen implementation hardcoded to foreign pages as WRITE|READ:
https://github.com/qemu/qemu/blob/master/hw/xen/xen-mapcache.c#L277-L289. In this case, the negative dev_filter_is_aie_and_amdxdna_drv test will fail.

maxzhen: OK, so what needs to happen is that:

- virt-gpu detect a read only mapping in guest
- pass readonly flag to vxdna
- vxdna pass readonly flag to xdna driver who should honor it and set up readonly iova mapping (in pasid mode, vxdna should mremap it as readonly)

Let's discuss it more offline.

houlz0507 (dismissed):

src/shim/virtio/platform_virtio.cpp
```diff
@@ -30,12 +30,64 @@ roundup_64bit(size_t size)
...
+// field) and the offsets of the two fields that define this request (guards a
+// reorder that happens to preserve the total size). If the layout above is ever
+// edited, the build fails loudly rather than silently shifting the ioctl number
+// or field positions (which would surface only at runtime as ENOTTY or garbage).
+static_assert(sizeof(virtgpu_resource_create_blob_userptr) == 72,
+              "virtgpu userptr blob request size must match the guest kernel UAPI");
+static_assert(offsetof(virtgpu_resource_create_blob_userptr, size) == 16,
+              "virtgpu userptr blob 'size' offset must match the guest kernel UAPI");
+static_assert(offsetof(virtgpu_resource_create_blob_userptr, userptr) == 48,
+              "virtgpu userptr blob 'userptr' offset must match the guest kernel UAPI");
```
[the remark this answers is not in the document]
wendyliang25: added `#include <cstddef>`

drivers/accel/amdxdna/amdxdna_ubuf.c
```diff
@@ -38,7 +39,8 @@ static void amdxdna_gem_ubuf_obj_free(struct drm_gem_object *gobj)
 
 	amdxdna_dma_unmap_bo(xdna, abo);
 	amdxdna_ubuf_unmap_dma(abo);
-	if (abo->mem.nr_pages)
+	/* Foreign pages are borrowed from the Xen mapping; never unpin them. */
+	if (abo->mem.nr_pages && !abo->mem.foreign)
 		unpin_user_pages(abo->mem.pages, abo->mem.nr_pages);
 	atomic64_sub(abo->mem.size >> PAGE_SHIFT, &abo->mem.mm->pinned_vm);
 	kvfree(abo->mem.pages);
```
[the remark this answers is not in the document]
wendyliang25: this is false alarm, for XEN direct map foreign pages, we didn't pin or get_page() within this function, and thus we don't need to put_page() or unpin

PR #1583, opened 2026-08-10
accel/amdxdna/ve2: fix context-switch handoff and HMM invalidate on AUX

rajkumar-xilinx:
- ve2_mgmt.c: restore is_partition_idle as a valid idle signal inside ve2_response_ctx_switch_req() instead of clearing it at the call site before the function ever saw it, and restore the legacy driver's arm-and-notify fallback (ve2_request_context_switch() + notify_fw_cmd_ready()) for when the FIFO head belongs to a different context but neither idle flag is set yet. Without this, an already-idle firmware had nothing to prompt it to ack, so a pending context switch could hang until the client-side command timeout.
- amdxdna_gem.c: skip the dma_resv_wait_timeout() call in amdxdna_hmm_invalidate() on VE2/AUX, since AUX has no cmd_submit-side BO lockdown that queues fences on the reservation object, so there is nothing to wait on there.


rajkumar-xilinx:
Aligned ve2_response_ctx_switch_req() logic with legacy driver. Verified all hw tests. Soft hang issue has not seen for start_col test.

PR #1585, opened 2026-08-10
fix(npu): recover wedged AIE2 firmware with a full NPU power-cycle

bong-water-water-bong:
# NPU firmware wedge recovery — PR to amd/xdna-driver

Branch: `feat/npu-wedge-recovery` (commit 424edeb)
Fork: `bong-water-water-bong/xdna-driver-upstream`
Upstream: `amd/xdna-driver` (current HEAD 5b875dc)

## Problem

On Strix Halo (PCI 0x17f0), the NPU firmware can wedge after ~1h of sustained
MoE decode (the 35B MoE NPU workstream; 1bit-systems issue #1568: "decode
degrades to constant '/' after ~1h"). Symptoms:

- Job scheduler fires a job timeout; the timed-out job never completes.
- Per-context recovery keeps failing: the management mailbox path
  (CREATE_CONTEXT / MAP_HOST_BUF / CONFIG_CU) cannot fix the device because
  the firmware is unresponsive, or its ERT rejects new contexts.
- Firmware ERT returns `AIE2_STATUS_MGMT_ERT_NOAVAIL (0x2000003)` for
  CREATE_CONTEXT once the device has degraded (observed in dmesg on
  7.0.0-29, in-tree amdxdna):

```
amdxdna 0000:c6:00.1: [drm] *ERROR* aie2_send_mgmt_msg_wait: command opcode 0x2 failed, status 0x2000003
amdxdna 0000:c6:00.1: [drm] *ERROR* aie2_hwctx_init: Alloc hw resource failed, ret -22
amdxdna 0000:c6:00.1: [drm] *ERROR* amdxdna_drm_create_hwctx_ioctl: Init hwctx failed, ret -22
```

A per-context restart re-submits to the same dead silicon. Only a full SMU
power-cycle (psp_stop + smu_fini, then smu_init + psp_start + mgmt fw init)
reloads the firmware and resets the ERT.

## Fix

`aie2_hw_reset()` — recover a wedged NPU with a full hardware power-cycle:

1. Suspend every context (existing `aie2_hw_suspend`: wait-for-idle,
   destroy context, stop scheduler).
2. `aie2_hw_stop()` + `aie2_hw_start()` — powers the NPU off/on through the
   SMU, reloading the firmware.
3. Resume every context (existing `aie2_hwctx_resume`: re-create context,
   remap heaps, config CU). Continues past a failed context so one broken
   context does not leave the rest of the device suspended.

Triggered from `aie2_sched_job_timedout()` when any of:

- the app-health query fails with a real error (`ret != -EOPNOTSUPP`), or
- the firmware reports `fatal_info.fatal_type`, or
- the per-context restart fails (covers the ERT-NOAVAIL case, where the
  firmware is alive but rejects new contexts).

`dev_lock` is already held in the timeout path (`guard(mutex)`), which
`aie2_hwctx_suspend` requires.

## Design decisions

- **`dpu_pc == UINT_MAX` is NOT a wedge signal.** `struct app_health_report`
  documents `dpu_pc` as UINT_MAX "before execution begins or after
  successful completion" (aie2_msg_priv.h) — i.e. the idle value. Using it
  as a death signal would false-positive on a healthy device whose DPU
  simply never latched a PC. The health-query failure + fatal_type +
  restart-failure signals are unambiguous.
- **Rate-limited**: one reset per `AIE2_HW_RESET_MIN_INTERVAL_MS` (60s).
  A firmware that re-wedges immediately must not produce a reset storm
  (each reset is a full SMU power-cycle). Repeated timeouts past the window
  indicate a deeper firmware problem and are surfaced in the log instead.
- **Bounded stall**: with a dead firmware, the suspend path's mailbox
  destroy is bounded by the existing TX/RX mailbox timeouts (2s/5s) — the
  same bound the existing TDR per-context recovery already accepts.

## Test status

- Compile-tested against kernel 7.0.0-29 (in-tree amdxdna) — clean build.
- Live wedge reproduction (~1h of MoE decode) and module hot-swap test are
  pending; happy to run them and report dmesg before merge.

## Repro notes

- Hardware: Strix Halo NPU 0x17f0 rev 0x11, firmware amdnpu/17f0_11/npu_7.sbin
- Workload: sustained MoE decode (1bit 35B MoE workstream) until the
  scheduler timeout fires and per-context recovery fails.

### DIFF
--- a/drivers/accel/amdxdna/aie2_ctx.c
+++ b/drivers/accel/amdxdna/aie2_ctx.c
@@ -489,6 +489,7 @@ aie2_sched_job_timedout(struct drm_sched_job *sched_job)
 	struct amdxdna_dev *xdna;
 	struct aie_device *aie;
 	bool fw_dead = false;
+	bool fw_fatal = false;
 	int ret;
 
 	xdna = hwctx->client->xdna;
@@ -511,12 +512,11 @@ aie2_sched_job_timedout(struct drm_sched_job *sched_job)
 				fw_dead = true;
 			kfree(report);
 		} else {
-			/* A fatal_type report is NOT a wedge signal by itself: the
-			 * firmware answered the query, so it is alive. dpu_pc ==
-			 * UINT_MAX is the documented idle value (see struct
-			 * app_health_report, aie2_msg_priv.h) and is likewise
-			 * deliberately not used as a wedge signal to avoid false
-			 * positives. */
+			/* fatal_type means the ERT hit a fatal exception. The firmware
+			 * still answers queries, so it is NOT a wedge signal by itself:
+			 * it only escalates when the per-context restart also fails. */
+			if (report->fatal_info.fatal_type)
+				fw_fatal = true;
 			job->aie2_job_health = report;
 		}
 	}
@@ -535,17 +535,22 @@ aie2_sched_job_timedout(struct drm_sched_job *sched_job)
 	aie2_hwctx_stop(xdna, hwctx, sched_job);
 
 	ret = aie2_hwctx_restart(xdna, hwctx);
-	if (ret || fw_dead) {
+	if (fw_dead || (fw_fatal && ret)) {
 		/*
 		 * Per-context recovery goes through the mailbox; it cannot fix a
-		 * dead firmware (an unresponsive health query or a restart the
-		 * firmware rejects). Power-cycle the NPU via SMU to reload the
-		 * firmware. aie2_hw_reset() rate-limits itself to one reset per
-		 * AIE2_HW_RESET_MIN_INTERVAL_MS.
+		 * dead firmware (an unresponsive health query, or a restart the
+		 * firmware rejects after reporting a fatal error). Power-cycle the
+		 * NPU via SMU to reload the firmware. aie2_hw_reset() rate-limits
+		 * itself to one reset per AIE2_HW_RESET_MIN_INTERVAL_MS.
 		 */
 		XDNA_WARN(xdna, "NPU firmware unhealthy (ctx restart ret %d, health %s) - power-cycling NPU",
-			  ret, fw_dead ? "unresponsive" : "ok");
+			  ret, fw_dead ? "unresponsive" : "fatal+rejected");
 		aie2_hw_reset(xdna);
+	} else if (ret) {
+		/* Restart rejected but no fatal report: the firmware is responsive,
+		 * so do not escalate to a device-wide power-cycle; log and let the
+		 * next timeout retry. */
+		XDNA_WARN(xdna, "NPU context restart rejected (ret %d) without fatal report - not power-cycling", ret);
 	}
 
 #ifdef HAVE_drm_gpu_sched_stat_reset
--- a/drivers/accel/amdxdna/aie2_ctx.c
+++ b/drivers/accel/amdxdna/aie2_ctx.c
@@ -512,9 +512,11 @@ aie2_sched_job_timedout(struct drm_sched_job *sched_job)
 				fw_dead = true;
 			kfree(report);
 		} else {
-			/* fatal_type means the ERT hit a fatal exception. The firmware
+			/*
+			 * fatal_type means the ERT hit a fatal exception. The firmware
 			 * still answers queries, so it is NOT a wedge signal by itself:
-			 * it only escalates when the per-context restart also fails. */
+			 * it only escalates when the per-context restart also fails.
+			 */
 			if (report->fatal_info.fatal_type)
 				fw_fatal = true;
 			job->aie2_job_health = report;
@@ -543,14 +545,21 @@ aie2_sched_job_timedout(struct drm_sched_job *sched_job)
 		 * NPU via SMU to reload the firmware. aie2_hw_reset() rate-limits
 		 * itself to one reset per AIE2_HW_RESET_MIN_INTERVAL_MS.
 		 */
-		XDNA_WARN(xdna, "NPU firmware unhealthy (ctx restart ret %d, health %s) - power-cycling NPU",
+		XDNA_WARN(xdna,
+			  "NPU firmware unhealthy (ctx restart ret %d, health %s) "
+			  "- power-cycling NPU",
 			  ret, fw_dead ? "unresponsive" : "fatal+rejected");
 		aie2_hw_reset(xdna);
 	} else if (ret) {
-		/* Restart rejected but no fatal report: the firmware is responsive,
+		/*
+		 * Restart rejected but no fatal report: the firmware is responsive,
 		 * so do not escalate to a device-wide power-cycle; log and let the
-		 * next timeout retry. */
-		XDNA_WARN(xdna, "NPU context restart rejected (ret %d) without fatal report - not power-cycling", ret);
+		 * next timeout retry.
+		 */
+		XDNA_WARN(xdna,
+			  "NPU context restart rejected (ret %d) without fatal report "
+			  "- not power-cycling",
+			  ret);
 	}
 
 #ifdef HAVE_drm_gpu_sched_stat_reset
--- a/drivers/accel/amdxdna/aie2_ctx.c
+++ b/drivers/accel/amdxdna/aie2_ctx.c
@@ -511,13 +511,12 @@ aie2_sched_job_timedout(struct drm_sched_job *sched_job)
 				fw_dead = true;
 			kfree(report);
 		} else {
-			/* fatal_type is set by the firmware when it reports an
-			 * unrecoverable error. Note: dpu_pc == UINT_MAX is the
-			 * documented idle value (see struct app_health_report,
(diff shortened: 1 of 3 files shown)


drivers/accel/amdxdna/aie2_pci.c
```diff
@@ -488,6 +488,62 @@ static int aie2_hw_resume(struct amdxdna_dev *xdna)
...
+		return 0;
+	}
+
+	if (time_before(jiffies, ndev->last_reset_jiffies +
+				msecs_to_jiffies(AIE2_HW_RESET_MIN_INTERVAL_MS))) {
+		XDNA_WARN(xdna, "NPU reset skipped: last reset %d ms ago",
+			  jiffies_to_msecs(jiffies - ndev->last_reset_jiffies));
+		return -EAGAIN;
+	}
+	ndev->last_reset_jiffies = jiffies;
```
houlz0507, line 529: Should this be moved after reset success?

drivers/accel/amdxdna/aie2_ctx.c
```diff
@@ -525,7 +535,19 @@ aie2_sched_job_timedout(struct drm_sched_job *sched_job)
 	job->job_timeout = true;
 	aie2_hwctx_stop(xdna, hwctx, sched_job);
 
-	aie2_hwctx_restart(xdna, hwctx);
+	ret = aie2_hwctx_restart(xdna, hwctx);
+	if (ret || fw_dead) {
```
houlz0507: Is it possible that fw_dead is set and aie2_hwctx_restart() success?

drivers/accel/amdxdna/aie2_ctx.c
```diff
@@ -506,10 +505,21 @@ aie2_sched_job_timedout(struct drm_sched_job *sched_job)
...
 			kfree(report);
-		else
+		} else {
+			/* fatal_type is set by the firmware when it reports an
+			 * unrecoverable error. Note: dpu_pc == UINT_MAX is the
+			 * documented idle value (see struct app_health_report,
+			 * aie2_msg_priv.h) and is deliberately NOT used as a
+			 * wedge signal to avoid false positives. */
+			if (report->fatal_info.fatal_type)
+				fw_dead = true;
```
houlz0507: Not sure if this could be marked to fw_dead. Firmware still responses.

drivers/accel/amdxdna/aie2_pci.c
bong-water-water-bong: Stamping before the reset is deliberate: if the power-cycle itself fails (NPU stays wedged), retries stay bounded to one per AIE2_HW_RESET_MIN_INTERVAL_MS window instead of storming the SMU on every job timeout. Retrying a failed power-cycle sooner doesn't un-wedge anything. Added a comment in b285390 documenting this.

drivers/accel/amdxdna/aie2_ctx.c
bong-water-water-bong: With b285390 this is no longer possible: fw_dead now means only that the health query failed, so the firmware is unresponsive and the mailbox restart will fail too. fatal_type no longer sets fw_dead — a responsive firmware that reports fatal gets the per-context restart instead of a device-wide power-cycle.

drivers/accel/amdxdna/aie2_ctx.c
bong-water-water-bong: Agreed — the health query succeeded, so the firmware is alive. Removed the fatal_type -> fw_dead mapping in b285390: a fatal report alone no longer triggers the NPU power-cycle; only an unresponsive health query or a rejected per-context restart does. The fatal report is still surfaced to userspace via the context health report for diagnostics.

bong-water-water-bong:
Refined the wedge logic per the review (387d7eb):

- `fw_dead` (health query unanswered) -> power-cycle, regardless of restart result.
- `fatal_type` alone -> NOT a wedge: firmware answered the query, so it is responsive.
- `fatal_type` + rejected restart -> power-cycle (firmware reported fatal AND refused recovery).
- Plain restart rejection without a fatal report -> logged, no device-wide power-cycle; next timeout retries.

The fatal report is still surfaced to userspace via the context health report. If fatal_type semantics differ (e.g. latched until power-cycle), happy to adjust.

houlz0507 (dismissed):

maxzhen:
ok to test

maxzhen:
@bong-water-water-bong , the checkpatch in CI is complaining as below:
```
  cmd: |-
    cd "/scratch/z1aiebuild-ws"/xdna-driver
    changed_files=$(git diff --name-only origin/main...HEAD --  'include/**/*.h' 'drivers/accel/amdxdna/*.c' 'drivers/accel/amdxdna/*.h')
    if [ -n "$changed_files" ]; then
      "/scratch/z1aiebuild-ws/xdna-driver/tools/codingsty_check.sh" $changed_files
    fi
  delta: '0:00:03.790166'
  end: '2026-08-11 13:46:48.673757'
  msg: non-zero return code
  rc: 1
  start: '2026-08-11 13:46:44.883591'
  stderr: ''
  stderr_lines: <omitted>
  stdout: |-
    checkpatch: blob 2b7a42bbdd94 (freshly downloaded from mainline)
    source:     https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/plain/scripts/checkpatch.pl
    mainline:   last touched in commit 2e05544060b9 (https://git.kernel.org/torvalds/c/2e05544060b9)
  
    WARNING: Block comments use a trailing */ on a separate line
    #519: FILE: drivers/accel/amdxdna/aie2_ctx.c:519:
    +                        * it only escalates when the per-context restart also fails. */
  
    CHECK: line length of 109 exceeds 100 columns
    #548: FILE: drivers/accel/amdxdna/aie2_ctx.c:548:
    +               XDNA_WARN(xdna, "NPU firmware unhealthy (ctx restart ret %d, health %s) - power-cycling NPU",
  
    WARNING: Block comments use a trailing */ on a separate line
    #554: FILE: drivers/accel/amdxdna/aie2_ctx.c:554:
    +                * next timeout retry. */
  
    CHECK: line length of 119 exceeds 100 columns
    #555: FILE: drivers/accel/amdxdna/aie2_ctx.c:555:
    +               XDNA_WARN(xdna, "NPU context restart rejected (ret %d) without fatal report - not power-cycling", ret);
  
    total: 0 errors, 2 warnings, 2 checks, 1321 lines checked
    WARNING: Block comments use a trailing */ on a separate line
    #526: FILE: drivers/accel/amdxdna/aie2_pci.c:526:
    +        * the SMU on every job timeout. */
  
    WARNING: Block comments use a trailing */ on a separate line
    #539: FILE: drivers/accel/amdxdna/aie2_pci.c:539:
    +        * does not leave the rest of the device suspended. */
  
    total: 0 errors, 2 warnings, 0 checks, 1479 lines checked
  stdout_lines: <omitted>

```
Could you address these issues?

bong-water-water-bong:
Checkpatch nits addressed (61980ac): block comments now use kernel style (closing */ on its own line) and the two long XDNA_WARN lines are split. Verified locally with tools/codingsty_check.sh — clean (0 warnings), and both files still compile with the original kernel flags. Thanks @maxzhen.

maxzhen (approved):

bong-water-water-bong:
Tested on Strix Halo (the box from the PR description) with this branch (61980ac) built and loaded.

After re-enrolling the MOK key (Secure Boot module signature) and rebooting, the NPU comes up clean with the merged driver:
- amdxdna loads (out-of-tree, no signature errors), firmware 17f0_11 loads, /dev/accel0 present
- device enumeration + xrt smoke test pass
- end-to-end NPU inference works (llama3.2:1b via FastFlowLM, ~57 tok/s) with no mmap EAGAIN / "flush memory" issues
- no CREATE_CONTEXT NOAVAIL or job-timeout recurrence since reboot

The power-cycle recovery path itself has not needed to fire since reboot (no wedge in this window). The previously-wedging scenario was ~1h of sustained MoE decode, so the long-running decode test is still the definitive check — I will report back here if the wedge reproduces and the reset path gets exercised.

Thanks again for the review and merge.

bong-water-water-bong:
Follow-up on the definitive check: the 1-hour sustained MoE decode run completed clean with this branch (61980ac) built and loaded on the Strix Halo box.

- 3600s continuous decode (Qwen3.6-35B-A3B MoE, fresh context per request), 399 requests
- zero degraded outputs (no constant-'/', no token soup, no empty responses)
- all 39 known-answer probes correct (arithmetic, geography, etc.)
- zero amdxdna CREATE_CONTEXT/job-timeout errors in the kernel log across the entire hour
- final manual probes after the run: 2+2=4, Paris, 9*8=72 — all correct

The power-cycle recovery path was not exercised: the firmware never wedged in the window, so `aie2_hw_reset()` stayed dormant — which is the expected outcome for a healthy sustained-load run. If the wedge reproduces on a longer or heavier run, I'll report the reset path firing.

Thanks again for the review and merge.

PR #1586, opened 2026-08-11
accel/amdxdna/ve2: Add named tracepoints for hwctx/partition/cmd latency

rajkumar-xilinx:
Replace the generic __amdxdna_trace_point probe (three anonymous u64 args whose meaning varied per call site) with 17 named, structured TRACE_EVENTs covering the VE2 command/context lifecycle:
 - hwctx init/fini:        xdna_hwctx_init_{start,done}, xdna_hwctx_fini_{start,done}
 - partition init:         xdna_partition_init_{start,done}
 - IRQ handling:           xdna_irq_{enter,exit}
 - command submit/wait:    xdna_cmd_submit_start, xdna_cmd_submit, xdna_cmd_wait_{start,done}, xdna_cmd_complete
 - queue indices:          xdna_queue_{write,read}_index
 - scheduler work:         xdna_mgmt_schedule_cmd_{start,done}, xdna_sched_work_{start,done}

Each event carries typed, self-documenting fields (hwctx_id, seq, start_col, state, etc.) instead of positional arguments, so per-stage latency (init, partition bring-up, submit, wait, fw exec, scheduler) can be derived directly from ftrace/perf timestamps without decoding call-site-specific argument layouts.

accel/amdxdna/tools: 
-Fix stale trace group, add -h/--help and READMEs npu_perf_trace.sh recorded perf events with "-e amdxdna_trace:*", but the driver's TRACE_SYSTEM was renamed to "amdxdna" a while back (see xrt_profile's matching fix in #781), so it was silently capturing zero amdxdna tracepoints. 
-Update it to "-e amdxdna:*", and fix the same stale "amdxdna_trace:" prefix in npu_perf_analyze.sh's usage examples. This group is shared by aie2/aie4/ve2 alike, so the fix benefits all backends, not just VE2. 
-Add -h/--help to both scripts (npu_perf_trace.sh had no usage() at all; npu_perf_analyze.sh only printed usage on zero args). --help works without requiring root. 
-Add README.md documenting the capture -> analyze pipeline, all options (including the new -key/-k correlation filter), a VE2 tracepoint reference table, and troubleshooting tips.

### DIFF
--- a/drivers/accel/amdxdna/ve2_hwctx.c
+++ b/drivers/accel/amdxdna/ve2_hwctx.c
@@ -241,8 +241,7 @@ static void hsa_queue_commit_slot(struct amdxdna_hwctx *hwctx, u64 seq)
 	}
 	hsa_queue_sync_write_index_for_write(queue);
 
-	trace_amdxdna_trace_point("XRT_PROFILING_TRACE_UPDATE_WRITE_INDEX",
-				  hwctx->client->pid, hwctx->id, seq, header->write_index);
+	trace_xdna_queue_write_index(hwctx->name, hwctx->id, seq, header->write_index);
 	mutex_unlock(&queue->hq_lock);
 }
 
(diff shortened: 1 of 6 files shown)


drivers/accel/amdxdna/ve2_hwctx.c
```diff
@@ -1010,6 +1015,13 @@ int ve2_cmd_submit(struct amdxdna_hwctx *hwctx, struct amdxdna_sched_job *job, u
 		return ret;
 	}
 
+	/*
+	 * Command(s) are now committed into the host queue (write_index
+	 * advanced); *seq is the last sub-command's slot, i.e. the one
+	 * ve2_cmd_wait()/xdna_cmd_complete will report completion for.
+	 */
+	trace_xdna_cmd_submit(hwctx->name, hwctx->id, *seq, op);
```
chvamshi-xilinx, line 1023: do we need to call this trace_xdna_cmd_submit_done?

chvamshi-xilinx (approved):
one minor comment. You can fix it in next PR

PR #1610, opened 2026-08-18
Report AIE4 per-context preemption telemetry

NishadSaraf:
On AIE4 preemption counters could not be attributed to a hardware
context: FW TID showed a plain row counter and Ctx ID was always zero.
Unlike AIE2p, AIE4 published no mapping between the context id firmware
assigns, which is what indexes the counters, and the id the driver reports
to userspace.


NishadSaraf:
retest this please

maxzhen (approved):

PR #1631, opened 2026-08-26
shim: classify terminal command states instead of comparing values

atassis:
`poll_command()` decides a command is done with `state >= ERT_CMD_STATE_COMPLETED`, which assumes every in-flight state sorts below it.

`ert.h` documents `ERT_CMD_STATE_QUEUED`, `ERT_CMD_STATE_SUBMITTED` and `ERT_CMD_STATE_RUNNING` as internal scheduler states, i.e. in flight. But `SUBMITTED = 7` was appended after `ABORT = 6` to keep the ABI, so it sorts on the terminal side. A producer for it would make `poll_command()` report a running command as finished.

Nothing writes `SUBMITTED` today, so this is a trap rather than a live bug. Replace the comparison with an explicit switch, so `-Wswitch` flags the next state nobody has classified.

Verified: the helper compiles clean against `ert.h` with `-Wswitch -Wswitch-enum -Werror`, and deleting one `case` fails the build with `enumeration value 'ERT_CMD_STATE_NORESPONSE' not handled in switch`.

### DIFF
--- a/src/shim/hwq.cpp
+++ b/src/shim/hwq.cpp
@@ -99,7 +99,7 @@ poll_command(xrt_core::buffer_handle *cmd) const
 {
   auto boh = static_cast<cmd_buffer*>(cmd);
   auto cmdpkt = reinterpret_cast<volatile ert_packet *>(boh->vaddr());
-  if (cmdpkt->state >= ERT_CMD_STATE_COMPLETED) {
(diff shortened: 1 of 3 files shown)


maxzhen:
ok to test

maxzhen (approved):

PR #1635, opened 2026-08-27
accel/amdxdna: support graceful hw context destroy and restore on aie4

donwalkarsoham:
Suspend destroys hw contexts non-gracefully, so firmware discards their state and resume rebuilds from scratch. Firmware can instead park a context and return a restore_id for a later create to replay.

Destroy gracefully on suspend, keep the restore_id and pass it to the next create. The parked state lives in a firmware context restore pool that does not survive a firmware suspend, so copy it into a driver buffer once all contexts are destroyed and hand it back after the partition is recreated. Reset and TDR stay non-graceful, and an FLR drops both the ids and the pool. If firmware returns restore_id 0, or the pool cannot be saved, the ids are cleared and parked jobs aborted so resume builds fresh contexts.


xdavidz:
> Suspend destroys hw contexts non-gracefully, so firmware discards their state and resume rebuilds from scratch. Firmware can instead park a context and return a restore_id for a later create to replay.
> 
> Destroy gracefully on suspend, keep the restore_id and pass it to the next create. The parked state lives in a firmware context restore pool that does not survive a firmware suspend, so copy it into a driver buffer once all contexts are destroyed and hand it back after the partition is recreated. Reset and TDR stay non-graceful, and an FLR drops both the ids and the pool. If firmware returns restore_id 0, or the pool cannot be saved, the ids are cleared and parked jobs aborted so resume builds fresh contexts.

Will you be able to develope a shim_test for this too?

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -352,6 +369,17 @@ void aie4_hwctx_destroy(struct amdxdna_hwctx *hwctx, enum aie4_hwctx_flags flags
...
+	/*
+	 * A graceful destroy that returns restore_id 0 means firmware could not
+	 * reach a safe point and gave up on preserving the ctx, so its queue is
+	 * left in the same indeterminate state a TDR leaves behind. Abort the
+	 * parked jobs rather than let resume re-drive them, and reset the HSA
+	 * indices so the recreated ctx starts clean. Done here, once the ctx is
+	 * disconnected, as aie4_hwctx_cleanup_running_jobs() requires.
+	 */
+	if (graceful && !priv->restore_id)
+		aie4_hwctx_cleanup_running_jobs(hwctx, false);
```
wendyliang25: please rebase, function has renamed, it is wait_for_idle.
why graceful destroy failure is as TDR? but not the same normal destroy? do you know what firmware will do?
i don't see when `graceful && !priv->restore_id`, this function returns any failure,
which suggests the suspend flow assumes it succeeded.
i think we should move the cleanup_running_jobs()/wait_for_idle out from destroy hwctx, instead, it should be called by the caller.
and today, without graceful destroy, we don't do wait_for_idle() in suspend.

donwalkarsoham:
> > Suspend destroys hw contexts non-gracefully, so firmware discards their state and resume rebuilds from scratch. Firmware can instead park a context and return a restore_id for a later create to replay.
> > Destroy gracefully on suspend, keep the restore_id and pass it to the next create. The parked state lives in a firmware context restore pool that does not survive a firmware suspend, so copy it into a driver buffer once all contexts are destroyed and hand it back after the partition is recreated. Reset and TDR stay non-graceful, and an FLR drops both the ids and the pool. If firmware returns restore_id 0, or the pool cannot be saved, the ids are cleared and parked jobs aborted so resume builds fresh contexts.
> 
> Will you be able to develope a shim_test for this too?

Yes, a shim test is added in the PR.

drivers/accel/amdxdna/aie4_ctx.c
donwalkarsoham: Rebased, dropped the TDR handling, you're right. restore_id 0 means nothing was preserved, so the context is left exactly as a plain destroy leaves it and resume re-drives the jobs on a fresh context, same as today. Aborting would only lose commands that currently complete.
No drain left in aie4_hwctx_destroy() — it stays with the caller, and suspend still doesn't wait for idle. The suspend change is now just AIE4_HWCTX_NORMAL → AIE4_HWCTX_GRACEFUL.
On the missing failure: a failed destroy also returns 0, so both it and "firmware declined" fall back to a fresh context, which is safe either way

donwalkarsoham: Hi Wendy,
I checked with the FW team restore_id = 0 with the graceful flag set means the context hit an error during the destroy and can't be restored. They also confirmed a non-zero id is only returned once preemption completes successfully, so a zero means preemption did not complete.
Your other point holds up as well: FW already requests preemption and waits for it to complete as part of the graceful destroy, so the driver-side drain was duplicating work it does anyway. It's out of aie4_hwctx_destroy() and stays with the caller, and suspend still doesn't wait for idle — which also makes the wait_for_idle rename moot here.
On the missing failure signal — agreed, it isn't surfaced today. I've gone back to FW with one more question before changing the behaviour: whether a zero leaves the parked jobs safe to re-drive, or whether they have to be aborted. I'd rather have that confirmed than guess. Will update the patch once they answer.

One thing I'll fix regardless: a destroy message that fails and a successful destroy returning 0 currently collapse into the same value, so the driver can't tell "no answer from FW" from "FW says the context errored."

Rebased on latest main.

xdavidz: > Hi Wendy, I checked with the FW team restore_id = 0 with the graceful flag set means the context hit an error during the destroy and can't be restored. They also confirmed a non-zero id is only returned once preemption completes successfully, so a zero means preemption did not complete. Your other point holds up as well: FW already requests preemption and waits for it to complete as part of the graceful destroy, so the driver-side drain was duplicating work it does anyway. It's out of aie4_hwctx_destroy() and stays with the caller, and suspend still doesn't wait for idle — which also makes the wait_for_idle rename moot here. On the missing failure signal — agreed, it isn't surfaced today. I've gone back to FW with one more question before changing the behaviour: whether a zero leaves the parked jobs safe to re-drive, or whether they have to be aborted. I'd rather have that confirmed than guess. Will update the patch once they answer.
> 
> One thing I'll fix regardless: a destroy message that fails and a successful destroy returning 0 currently collapse into the same value, so the driver can't tell "no answer from FW" from "FW says the context errored."
> 
> Rebased on latest main.

All of these should be designed well, thus I recommend you document all of these ontop a page, we will review this.
Document here won't help us to chase down every design corner cases.

@donwalkarsoham, I will put a do-no-merge till we review the full design of the graceful shutdown. Especially, all failure cases handling design.

donwalkarsoham: ok sure, will document the cases.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -322,8 +341,26 @@ void aie4_hwctx_destroy(struct amdxdna_hwctx *hwctx, enum aie4_hwctx_flags flags
 	if (priv->kernel_submit && has_reset)
 		wake_up_all(&priv->job_list_wq);
 
-	if (flags != AIE4_HWCTX_DISCONNECT)
-		aie4_msg_destroy_context(ndev, priv->hw_ctx_id);
+	if (flags != AIE4_HWCTX_DISCONNECT && priv->hw_ctx_id != CTX_INVALID_ID) {
+		ret = aie4_msg_destroy_context(ndev, priv->hw_ctx_id, graceful);
+		if (ret > 0)
+			restore_id = ret;
+		else if (graceful && !ret)
```
wendyliang25, line 348: if graceful failed, on the firmware side, will it change to a non-graceful destroy automatically without driver taking any action or it will not destroy hwctx? will resume work if graceful destroy failed?
just want to confirm, as the  `aie4_hwctx_destroy()` consider it has completed regardless of the error, it just print warning when it fails.

donwalkarsoham: Hi Wendy,
Per the FW team this means the context was destroyed but its state couldn't be preserved, so no restore id comes back and there's nothing for the driver to re-issue — resume just creates the context fresh and works normally.
The parked jobs are to be aborted though. They can't be replayed against a new context, so resume fails them back to userspace before recreating it. Hence, will be adding this functionality in the latest revision.
Also, NO_RESTORE is a new message status code firmware added in protocol v6.6. I am adding it in the latest revision.
On your last point, you're right that destroy continues and only warns — that's deliberate for NO_RESTORE and a zero restore_id but we will be aborting jobs during resume.

drivers/accel/amdxdna/aie.c
```diff
@@ -68,14 +69,30 @@ int aie_send_mgmt_msg_wait(struct aie_device *aie, struct xdna_mailbox_msg *msg)
 		aie_destroy_chann(aie, &aie->mgmt_chann);
 
 	if (!ret && *hdl->status) {
-		XDNA_ERR(xdna, "command opcode 0x%x failed, status 0x%x",
-			 msg->opcode, *hdl->data);
+		if (!quiet)
```
wendyliang25: why is this required in this patch? will graceful shutdown introduce a period causing expected mgmt msg failure?

donwalkarsoham: Hi Wendy,
There's no period where mgmt messages fail. DESTROY_HW_CONTEXT sent with the graceful flag can return NO_RESTORE (0x4), meaning the destroy happened but not gracefully, so there's no state to restore. Per FW that usually means the context was already in a bad or TDR state on the way down.

This patch treats that as a normal outcome, but aie_send_mgmt_msg_wait() maps any non-zero status to -EINVAL and logs it at error level, so a plain suspend would print command opcode 0x30004 failed, status 0x4 when nothing actually failed. It's the first caller that expects a non-zero status — until now every non-zero status really was a failure, so the blanket error log was correct. The suppression is narrow: it's guarded by graceful, so a non-graceful destroy still reports 0x4 as an error, and aie4_msg_destroy_context() still warns on a genuine failure, now with the status included.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -322,8 +413,34 @@ void aie4_hwctx_destroy(struct amdxdna_hwctx *hwctx, enum aie4_hwctx_flags flags
 	if (priv->kernel_submit && has_reset)
 		wake_up_all(&priv->job_list_wq);
 
-	if (flags != AIE4_HWCTX_DISCONNECT)
-		aie4_msg_destroy_context(ndev, priv->hw_ctx_id);
+	if (flags != AIE4_HWCTX_DISCONNECT && priv->hw_ctx_id != CTX_INVALID_ID) {
+		ret = aie4_msg_destroy_context(ndev, priv->hw_ctx_id, graceful);
+		if (ret > 0)
+			restore_id = ret;
+		else if (graceful && (!ret || ret == -ENODATA))
```
wendyliang25: Hi @donwalkarsoham, does it mean when ret == -ENODATA, the aie4_msg_destroy_context() can return 0 when graceful so that here we just need to check if (graceful && !ret) ?

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -1485,6 +1598,13 @@ int aie4_hwctx_resume_all(struct amdxdna_dev_hdl *ndev)
 	amdxdna_for_each_client(xdna, client) {
 		idx = srcu_read_lock(&client->hwctx_srcu);
 		amdxdna_for_each_hwctx(client, hwctx_id, hwctx) {
+			/*
+			 * Abort the parked jobs before the create, which clears
+			 * the reset state the drain relies on.
+			 */
+			if (hwctx->priv->restore_state == AIE4_CTX_RESTORE_LOST)
+				aie4_hwctx_abort_jobs(hwctx);
```
wendyliang25: when AIE4_CTX_RESTORE_LOST, you don't know  when CERT stops running as FW doesn't guarantee CERT stops when it responded. and thus, you don't do set `reset` in hwctx destroy, instead you do it here, hoping at this point, CERT already stops, so that driver sets the hsa read_idx assuming it will not conflict with CERT. is this the reason you do it here?

donwalkarsoham: Hi Wendy, as David had suggested, the reason is the suspend path rather than CERT timing. aie4_hwctx_abort_jobs() waits for the worker to drain in aie4_hwctx_wait_for_running(), so calling it from aie4_hwctx_destroy() would stall the runtime suspend on the worker, and it would discard a live application's queued commands just because the device went idle. Resume is also the first point where the outcome is final, since a restore id can still be dropped there if the restore pool did not survive (aie4_clear_ctx_restore_ids()).

Your point about CERT is a good one though. job_abort() forces read_index forward, which must not race CERT, and doing it at resume after the power transition does make that safe. But if a destroy response does not guarantee CERT has stopped, the same concern applies to the existing AIE4_HWCTX_ERROR path, which drains right after the destroy reply with no power transition (aie4_pci.c:1547 and aie4_error.c:97). That is pre-existing and unchanged by this patch

drivers/accel/amdxdna/aie4_ctx.c
donwalkarsoham: Hi Wendy, yes. NO_RESTORE means the context was destroyed with nothing preserved, which is the same thing as a restore id of 0, so returning 0 is the right encoding and keeps a negative return meaning "the context was not destroyed". The -ENODATA was only there to tell NO_RESTORE apart from a SUCCESS with restore id 0, but the FW team has confirmed those are the same case (current FW always reports NO_RESTORE), and the caller treats them alike anyway. Will change aie4_msg_destroy_context() to return 0 so the check here becomes if (graceful && !ret). Thanks for catching it.

donwalkarsoham: Add the change in the latest version and updated the documentation, as per latest FW response.

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -273,14 +309,69 @@ static bool aie4_hwctx_has_reset(struct amdxdna_hwctx *hwctx)
 	return READ_ONCE(hwctx->priv->has_reset);
 }
 
+/*
```
xdavidz, line 308: This comment is off the trail for graceful destroy, you are describing what the existing destroy is doing and with graceful destroy.

Please re-consider adding redundant comment for destroy_context.
And only add specific comment for graceful destroy.

you should either add it above destroy_context or destroy_hwctx, that should cover the following cases:
1) a happy path that nothing goes wrong for a graceful destroy, aka. a valid restore id is returned and recovered successfully later.
2) a NO_RESTORE(0x4) is returned to indicate that destroy done but restore id is 0. How your code handle this when resume.
   2.1) when resume, those destoryed context should be recreated?
   2.2) what if recreation failed, what should we do?
3) any other Non-zero return code from destroy_context, how do you handle it during suspend and resume
   3.1) what we should do during suspend?
   3.2) what we should do during resume?

Basicall, I discussed this with Wendy, I think you can utilized existing TDR flow, unless you proved some behaviors are totally different.

xdavidz (changes requested):

drivers/accel/amdxdna/aie.c
```diff
@@ -52,7 +52,12 @@ void aie_destroy_chann(struct aie_device *aie, struct mailbox_channel **chann)
 	*chann = NULL;
 }
 
-int aie_send_mgmt_msg_wait(struct aie_device *aie, struct xdna_mailbox_msg *msg)
+/*
+ * @quiet_status: a reply status the caller handles itself. It still fails the
+ * call, it is just not logged here as a firmware error.
+ */
+int aie_send_mgmt_msg_wait_quiet(struct aie_device *aie, struct xdna_mailbox_msg *msg,
+				 u32 quiet_status)
```
wendyliang25: looks like when graceful shutdown fails, the implementation will abort all pending jobs, just wonder why we need to introduce a new function to suppress the error message when the `ret` returns fail?

donwalkarsoham: Hi Wendy.
As NO_RESTORE isn't a firmware error as the ctx is destroyed and its resources released, only the state isn't preserved — so the generic "command opcode 0x30004 failed" reads as if the destroy itself failed, and the specific warnings that follow say it better. But yes we do fail the call and abort the jobs, so an error log isn't wrong, and it isn't worth changing a helper shared by every mgmt message for one caller. Dropped the aie.c/aie.h change in the latest revision; the error now prints as before.

drivers/accel/amdxdna/aie4_ctx.c
donwalkarsoham: Done

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -273,14 +304,52 @@ static bool aie4_hwctx_has_reset(struct amdxdna_hwctx *hwctx)
...
+ * Tear down the firmware side of @hwctx and drop the driver state that goes
```
xdavidz: This smells too much AI words. A suggested version, please check the spell by AI for us.

xdavidz:
```
/*
 * The driver suspend requires to gracefully destroy in-flight hwctx and the
 * subsequent resume can resume those hwctx properly.
 *
 * The graceful flag is designed to ask the firmware to clean up in-flight
 * hwctx graceful, thus it handles the following scenarios:
 *
 *No# | firmware reply            | restore_id     | suspend    | resume
 * --------------------------|---------|------------|-------------------------
 * 1. |SUCCESS, restore_id != 0  | valid id       | keep id    | create ctx by id, resume jobs
 * 2. |NO_RESTORE (status 0x4)   | 0 invalid      | abort jobs | create new ctx without id
 * 3. |ERROR (status 0x1)        | -EINVAL | -          | create fresh, new jobs
 * 4. |no reply (timeout)        | -ETIME  | -          | create fresh, new jobs
 *
[DZ] I think you cannot stop suspend during ERROR or timeout, suspend finished with errors
     later during resume, you have to create fresh ctx, then this is the new jobs per ctx?

 * Firmware only adds a restore id once preemption has completed, so it answers
 * either SUCCESS with a non-zero id or NO_RESTORE.  A non-zero id is dropped
 * anyway if the restore pool did not survive the power transition, which moves
 * that ctx to the "create fresh" row with its jobs intact.
 *
[DZ] The above line is confusing, it indicates that we've got valid restore id, but if we
     use this restore id to create ctx during resume it failed. What do we do here?
     Seems your answer is that we re-create the ctx without the restore id.
     Do we need to try destroy the ctx before re-create the ctx?

 * Every row recreates the ctx on resume; only the id and the parked jobs
 * differ. 
[DZ] this English is hard to understand. "id and jobs differ" than what? 

 * NO_RESTORE means firmware destroyed the ctx without preserving it,
 * so its parked jobs can no longer be re-driven and the suspend aborts them
 * before moving on - the resume then builds an empty ctx.
[DZ] this make sense and match the case No.2.

  If any create
 * fails, aie4_hwctx_resume_all() gives up on the resume and its caller falls
 * into the existing recovery:
[DZ] again this English is unclear, do you mean you will fail the resume?

 * aie4_hwctx_suspend_all(clean_jobs), which
 * destroys every ctx as AIE4_HWCTX_ERROR and drains its jobs, then stops the
 * hardware.
[DZ] does this match the TDR handling today that if the resume_all failed, driver
   will clean_jobs and destroy any existing hwctx and mark the resume failed?
 *
 * ERROR means firmware did not act on the request at all (a stale ctx id, or
 * under SR-IOV a VF reaching for another VF's ctx), so that ctx can outlive
 * the call.  The suspend does not treat it specially, as the power transition
 * drops firmware state regardless.
[DZ] remove those unclear description.

[DZ] more importantly you need to keep the table above and explain the following situations:
1) For No.1, during resume, create ctx by restore id failed, what you will do?
2) For No.2, during resume, create new ctx failed, what you will do?
3) For No.3, during resume, should you destroy the ctx then re-create them? If any steps failed, what will you do?
4) For No.4, during resume, should you destroy or just re-create? If any steps failed, what will you do?

 */
```

xdavidz (changes requested):

donwalkarsoham:
Retest this please

xdavidz (dismissed):
The comments look good to me, thanks for handling this.
Please double check with @wendyliang25 if this handling look ok by her from TDR point of view.

drivers/accel/amdxdna/aie4_ctx.c
donwalkarsoham: Added suggested changes.

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -396,6 +398,108 @@ static int aie4_fw_load(struct amdxdna_dev_hdl *ndev)
...
+	struct amdxdna_hwctx *hwctx;
+	unsigned long hwctx_id;
+	int idx;
+
+	drm_WARN_ON(&xdna->ddev, !mutex_is_locked(&xdna->dev_lock));
+
+	amdxdna_for_each_client(xdna, client) {
+		idx = srcu_read_lock(&client->hwctx_srcu);
+		amdxdna_for_each_hwctx(client, hwctx_id, hwctx)
+			hwctx->priv->restore_id = 0;
```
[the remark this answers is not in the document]
donwalkarsoham: Firmware did reach a safe point here — it returned a nonzero restore id. Only the driver's copy of the pool was lost, so the queue is parked at a clean boundary and the packets are re-drivable.

The abort is keyed on firmware never reaching a safe point (AIE4_CTX_RESTORE_LOST, set only on NO_RESTORE), not on state being lost. Rows 3 and 4 re-drive too, which is the pre-graceful behaviour this patch keeps.

Aborting here would also fire on a non-error: aie4_save_ctx_restore_pool() returns -ENODATA when firmware's pool is simply empty. And no duplication — read_index only advances on completion, so the replay starts at the first unfinished packet.

drivers/accel/amdxdna/amdxdna_ctx.c
```diff
@@ -362,14 +362,27 @@ int amdxdna_drm_destroy_hwctx_ioctl(struct drm_device *dev, void *data, struct d
...
+	 * data, and there is no message to release it on its own. Resume before
+	 * the erase below, while the ctx is still visible to the resume walk, so
+	 * it is recreated and the destroy reaches firmware. Best effort only: a
+	 * ctx must always be destroyable, so a failed resume still destroys it
+	 * and leaves the pool entry behind until the next suspend or reset.
+	 */
+	pm_ret = amdxdna_pm_resume_get(xdna);
+	if (pm_ret)
+		XDNA_WARN(xdna, "Resume failed (%d), ctx %d restore data not reclaimed",
+			  pm_ret, args->handle);
```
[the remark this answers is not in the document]
donwalkarsoham: Fixed — amdxdna_drm_close() now takes the same best-effort resume reference, skipped when the client holds no ctx.

One correction: a stale entry can't be loaded later. A restore id only reaches firmware via aie4_hwctx_create(), and a new hwctx has a kzalloc'd priv, so its id is 0. This was a firmware pool leak, not a bad restore.

test/shim_test/io_test.cpp
```diff
@@ -807,6 +858,129 @@ TEST_io_suspend_resume(device::id_type id, std::shared_ptr<device>& sdev, arg_ty
...
 
+/*
+ * Suspend destroys every live context gracefully and saves the firmware context
+ * restore pool; resume writes the pool back and re-creates each context from the
+ * restore id firmware handed out. Hold one context idle across two auto-suspend
+ * windows and submit through it after each one.
+ *
+ * A driver that destroys contexts non-gracefully passes this too, since it
+ * rebuilds them from scratch. Nothing exposes the restore id to user space, so
+ * this guards the restore path against regressing rather than proving it ran.
```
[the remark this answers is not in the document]
donwalkarsoham: No interface reports whether a ctx was rebuilt from a restore id, so a direct assertion isn't possible today.

drivers/accel/amdxdna/aie4_pci.c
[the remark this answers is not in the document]
donwalkarsoham: INVALID, no fix. 
The abort jobs is done when firmware sets only on NO_RESTORE, meaning it never reached a safe preemption point. A context only reaches aie4_clear_ctx_restore_ids() because it did reach one and returned a nonzero restore id — so aborting is not needed. Losing the pool afterwards is a driver-side accident; the queue is still parked at a clean boundary and replays fine.

donwalkarsoham:
> ### 🟡 Changes recommended
> Unresolved critical and moderate findings affect parked-job recovery, DMA cache coherency, and suspended cleanup.
> 
> _Get a fresh assessment by requesting another Copilot review._
> 
> Review details
> ### Suppressed comments (1)
> **drivers/accel/amdxdna/aie4_pci.c:472**
> 
> * After `GET_CTX_RESTORE_POOL` firmware writes this DMA buffer, but this path never invalidates the CPU cache before retaining it for the later SET. On the non-coherent `dma_alloc_noncoherent` path, the resume can therefore hand firmware stale pool bytes; the existing telemetry read path flushes after a device write for the same reason. Flush the buffer after a successful get.
> 
> ```
> 	ret = aie4_msg_get_ctx_restore_pool(ndev, to_dma_addr(buf_hdl, 0),
> 					    to_buf_size(buf_hdl));
> ```
> 
> * **Files reviewed:** 8/8 changed files
> * **Comments generated:** 3
> * **Review effort level:** Lite

 INVALID. drm_clflush_virt_range() at line 469 already writes back and invalidates, and the CPU never touches the buffer between the GET and the SET. Its own cited precedent flushes after only because the CPU is about to read.

wendyliang25 (approved):

drivers/accel/amdxdna/aie4_ctx.c
```diff
@@ -1271,6 +1379,38 @@ void aie4_hwctx_wait_for_running(struct amdxdna_hwctx *hwctx)
...
+ * puts the worker on the drain path; aie4_hwctx_create() clears it again.
+ * User-mode submission keeps no driver-side job list, so there is nothing to
+ * abort there.
+ */
+void aie4_hwctx_abort_jobs(struct amdxdna_hwctx *hwctx)
+{
+	struct amdxdna_hwctx_priv *priv = hwctx->priv;
+
+	if (!priv->kernel_submit)
+		return;
```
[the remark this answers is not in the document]
donwalkarsoham: Valid, fixed. abort_umq_cmds() now advances read_index to the published write_index before returning.

drivers/accel/amdxdna/aie4_pci.c
```diff
@@ -396,6 +398,108 @@ static int aie4_fw_load(struct amdxdna_dev_hdl *ndev)
...
+	if (!size)
+		return -ENODATA;
+
+	if (size > AIE4_MAX_CTX_RESTORE_POOL_SIZE) {
+		XDNA_ERR(xdna, "Ctx restore pool size 0x%x exceeds 0x%lx", size,
+			 AIE4_MAX_CTX_RESTORE_POOL_SIZE);
+		return -EFBIG;
+	}
+
+	buf_hdl = amdxdna_alloc_msg_buff(xdna, size);
```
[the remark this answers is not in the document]
donwalkarsoham: This isn't a bug introduced here — aie.c isn't part of this patch. Can be handled in a separate PR.
Explanation - 
amdxdna_alloc_msg_buff() sets the direction for every caller, and firmware already reads two existing buffers through it: the coredump buffer list and the tile-memory write buffer. The restore pool is just a third case.
Also, "outside the IOVA path" is the default path, not an edge case — force_iova is off by default, and restore works there today.
So, the fix belongs in the allocator, where it covers all three.

xdavidz (approved):

wendyliang25 (approved):

drivers/accel/amdxdna/aie4_ctx.c
donwalkarsoham: Was added.

PR #1654, opened 2026-09-01
accel/amdxdna/ve2: fail submit with -ENOTRECOVERABLE after firmware MISC

rajkumar-xilinx:
Firmware misc_status is level-triggered, so a poisoned hwctx must not look like a malformed command. Latch the first MISC, reject later submits, and have the shim tell userspace to destroy/recreate the context.


rajkumar-xilinx:
Verification is in progress

chvamshi-xilinx (approved):

PR #1668, opened 2026-09-02
test/shim_test: Add FLR Classic/PF/VF shim test and disable telemetry…

hlaccabu:
… tests for VF

### DIFF
--- a/test/shim_test/shim_test.cpp
+++ b/test/shim_test/shim_test.cpp
@@ -107,6 +107,9 @@ void TEST_certlog_num_ucs_overflow(device::id_type, std::shared_ptr<device>&, ar
 void TEST_certlog_invalid_uc_index(device::id_type, std::shared_ptr<device>&, arg_type&);
 void TEST_certlog_payload_overflow(device::id_type, std::shared_ptr<device>&, arg_type&);
 void TEST_df_bw(device::id_type, std::shared_ptr<device>&, arg_type&);
+void TEST_aie4_flr(device::id_type, std::shared_ptr<device>&, arg_type&);
+void TEST_aie4_pf_flr(device::id_type, std::shared_ptr<device>&,
+                      std::shared_ptr<device>&);
 
 inline void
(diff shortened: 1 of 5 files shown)


test/shim_test/shim_test.cpp
```diff
@@ -136,6 +139,8 @@ enum hw_type {
   npu1,     // AIE2, device_id 0x1502
   npu4,     // AIE2, device_id 0x17f0
   npu3,     // AIE4, device_id 0x17f1 / 0x1b0a (canonical)
+  npu3vf,   // AIE4 VF, device_id 0x17f3 / 0x1b0c (SRIOV only)
+  npu3classic, // AIE4 Classic, device_id 0x17f1 / 0x1b0a (classic only)
```
maxzhen: The design here is to make sure there is no overlap across all hw types. If you need to introduce new types, please review existing ones and if you see overlapping, you can break existing one into multiple types as needed. Per our offline discussion, you can remove "npu3classic" completely and just use "npu3". If you can, please add a comment before enum hw_type to emphasize the no-overlapping rule.

test/shim_test/shim_test.cpp
```diff
@@ -208,6 +219,7 @@ static const hw_filter_entry hw_filter_table[] = {
   { npu1,    dev_filter_is_npu1   },
   { npu4,    dev_filter_is_npu4   },
   { npu3,    dev_filter_is_npu3   },
+  { npu3vf,  dev_filter_is_npu3_vf},
```
maxzhen: To be consistent: dev_filter_is_npu3_vf -> dev_filter_is_npu3vf

test/shim_test/shim_test.cpp
maxzhen: Please add a comment to indicate that any device can only belong to **one** hw type.

maxzhen (changes requested):

maxzhen (approved):

PR #1670, opened 2026-09-02
Clean up context after failed `set_dpm_level`

andrej:
xrs_allocate_resource() calls actions->load() to create the hardware context. That asks the firmware for a context and registers an interrupt named in the firmware reply. It then calls `set_dpm_level`.

When `set_dpm_level` fails -- which is something we have observed happening in our MLIR-AIE CI -- the error path never calls actions->unload().

This breaks every later attempted context allocation that would use the same interrupt index:

```
  [drm] *ERROR* Set hclock to 1600 failed, ret -110
  genirq: Flags mismatch irq 95. 00200000 (xdna_mailbox) vs. 00200000 (xdna_mailbox)
  xdna_mailbox.95: Failed to request irq 95 ret -16
```

### DIFF
--- a/drivers/accel/amdxdna/aie2_message.c
+++ b/drivers/accel/amdxdna/aie2_message.c
@@ -385,6 +385,7 @@ int aie2_create_context(struct amdxdna_dev_hdl *ndev, struct amdxdna_hwctx *hwct
 
 free_channel:
 	xdna_mailbox_free_channel(hwctx->priv->mbox_chann);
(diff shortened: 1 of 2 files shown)


maxzhen (approved):
The fix makes sense, but do we know why set_dpm_level has failed on your system? Any msg/hint? cc'ing @NishadSaraf

maxzhen:
ok to test

PR #1689, opened 2026-09-11
AIE4 auto coredump support

hlaccabu:


test/shim_test/io_test.cpp
```diff
@@ -1480,9 +1480,11 @@ TEST_io_coredump(device::id_type id, std::shared_ptr<device>& sdev, arg_type& ar
...
 
   auto dev = sdev.get();
-  // Use a dedicated timing-out ELF so the AIE MEM/REG read-write tests can keep
-  // using the original (completing) aie_debug ELF.
-  static const char* tag = "aie_debug_coredump";
+  auto pci_id = aie4_binary_device_id(device_query<query::pcie_device>(dev));
+  const bool is_aie4 = (pci_id == npu3_device_id || pci_id == npu3a_device_id);
+  // npu4 uses a dedicated timing-out aie_debug xclbin so the MEM/REG tests can
+  // keep the completing ELF. npu3 has no equivalent xclbin; reuse bad_timeout.
+  const char* tag = is_aie4 ? "bad_timeout" : "aie_debug_coredump";
```
maxzhen, line 1487: can we use existing dev filter functions?

drivers/accel/amdxdna/amdxdna_ctx.h
```diff
@@ -199,7 +199,10 @@ struct amdxdna_hwctx {
 	atomic64_t			job_free_cnt ____cacheline_aligned_in_smp;
 	wait_queue_head_t		job_free_wq;
 
-	/* Saved core dump, captured automatically on timeout if device auto_coredump is set. */
+	/*
+	 * Saved core dump if auto_coredump is set: AIE2 on timeout, AIE4 on firmware
+	 * context-error recovery.
+	 */
```
maxzhen, line 205: On both aie2 and aie4, auto-coredump will be saved on timeout only. For other errors, the coredump automatically obtained will not be useful. And we should not do it.

maxzhen (changes requested):

test/shim_test/io_test.cpp
hlaccabu: Not in io_test.cpp, we don't include shim_test.cpp in io_test.cpp currently.

drivers/accel/amdxdna/amdxdna_coredump.c
```diff
@@ -236,13 +236,37 @@ static bool amdxdna_saved_coredump_filter(struct amdxdna_hwctx *hwctx, void *arg
...
+			amdxdna_hwctx_walk(tmp_client, NULL, amdxdna_saved_coredump_filter,
+					   amdxdna_free_saved_coredump_cb);
+		}
+	}
+
+	return 0;
+}
+
 int
 amdxdna_set_auto_coredump_mode(struct amdxdna_client *client,
```
maxzhen, line 265: you created a new api, but who is calling amdxdna_set_auto_coredump_mode now?

maxzhen (changes requested):

drivers/accel/amdxdna/amdxdna_ctx.h
hlaccabu: Right, on AIE4 that timeout is implemented as firmware ctx-error recovery (AIE4 has recovery that the code names TDR but it does not have AIE2’s driver hang detector), and the dump has to be taken before we destroy the context. I can reword the comment to that so it doesn’t read as “dump all errors.” if you prefer.

test/shim_test/io_test.cpp
maxzhen: it's time to fix it. it is better to have one implementation for device filtering throughout shim test.

drivers/accel/amdxdna/amdxdna_ctx.h
maxzhen: timeout is timeout no matter how it's handled. please revert the change here.

drivers/accel/amdxdna/amdxdna_coredump.c
hlaccabu: AIE2 still goes through amdxdna_set_auto_coredump_mode, I created the _state() helper for AIE4 to avoid double copy_from_user. Now it copies once, programs hws debug mode, then applies that same state.state. aie4_set_force_preempt_state() does something similar to avoid double copy.

test/shim_test/io_test.cpp
hlaccabu: Got it, I will open a second PR to implement this shim test change and update this PR once that is merged.

drivers/accel/amdxdna/amdxdna_coredump.c
maxzhen: This is not good. Two common APIs doing almost the same thing. We can have one API which calls platform specific callback. Please introduce such a callback for enabling debug mode in firmware and aie2 does not need to provide it.

test/shim_test/io_test.cpp
maxzhen: sure. let's have a dev filter .h and include it whenever it's needed.

drivers/accel/amdxdna/amdxdna_coredump.c
hlaccabu: Done, but I still had amdxdna_set_auto_coredump_state live in amdxdna_coredump.c. Would you rather it be renamed aie4_* and moved to an aie4 specific file?

drivers/accel/amdxdna/amdxdna_coredump.h
```diff
@@ -30,7 +30,8 @@ char *amdxdna_get_hwctx_coredump(struct aie_device *aie,
 				 struct amdxdna_hwctx *hwctx);
 int amdxdna_get_auto_coredump_mode(struct amdxdna_client *client,
 				   struct amdxdna_drm_get_info *args);
-int amdxdna_set_auto_coredump_mode(struct amdxdna_client *client,
+int amdxdna_set_auto_coredump_mode(struct aie_device *aie,
+				   struct amdxdna_client *client,
```
maxzhen, line 34: you can't find the corresponding struct aie_device from client?

maxzhen (dismissed):

drivers/accel/amdxdna/amdxdna_coredump.h
hlaccabu: We can find xdna device, but to get to aie device we'd need xdna->dev_handle->aie chain, and dev_handle would be undefined in amdxdna_* common files because those are device specific. We would have to include aie2_pci.h or aie4.h inside amdxdna_coredump.c and I don't think we want to do that.

maxzhen: ok, makes sense. leave the cleanup to me, then.

maxzhen: It's done by https://github.com/amd/xdna-driver/pull/1692

NishadSaraf (dismissed):

NishadSaraf:
LGTM. Please rebase to latest main.

PR #1690, opened 2026-09-11
test/shim_test: relocate hw_type and dev_filter entries to new dev_fi…

hlaccabu:
…lter.h

### DIFF
--- a/test/shim_test/df_bw_test.cpp
+++ b/test/shim_test/df_bw_test.cpp
@@ -3,6 +3,7 @@
 
 #include "io.h"
 #include "dev_info.h"
+#include "dev_filter.h"
 #include "exec_buf.h"
 
 #include "core/common/device.h"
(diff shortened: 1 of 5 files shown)


test/shim_test/df_bw_test.cpp
```diff
@@ -156,14 +157,14 @@ void
 TEST_df_bw(device::id_type id, std::shared_ptr<device>& sdev, arg_type& arg)
 {
   auto dev = sdev.get();
-
-  const auto& info = get_binary_info(dev, df_bw_tag, nullptr);
+  const bool is_aie4 = dev_filter_is_aie4(id, dev);
+  const flow_type flow = is_aie4 ? FULL_ELF : PARTIAL_ELF;
```
maxzhen: why making this change? what is not good with previous code? i thought that flow matters, not dev type, no?

PR #1704, opened 2026-09-17
accel/amdxdna: implement the shmem+IPI mailbox transport for the aie4 platform

bisingha-xilinx:
The aie4 platform skeleton landed the platform mailbox as stubs returning -EOPNOTSUPP, which fails during probe since `aie4_query_fw()` sends its first command through them. Implement the transport so the driver talks to the RPU FW
over SHMEM + IPI on the VEK385 (npu12).

- Add the shmem ring ABI: ring layout and SPSC helpers shared with the RPU FW.
- Implement the mgmt rings, send, RX drain, doorbell and channel teardown.
- Report doorbell errors, which previously looked like a successful submit.

Three things are load-bearing: message id 0 is reserved for firmware-initiated messages and routed to the async callback; the RX drain is shared with `drain_channel()` because firmware completes some commands without raising the IPI; and the RX callback must ack on its channel or the zynqmp-ipi notification stays masked after the first interrupt.


drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -98,16 +172,195 @@ static int plat_map_region(struct amdxdna_dev *xdna, struct device *dev,
...
+	mb->rx_ring = mgmt + half + sizeof(struct shmem_ring_hdr);
+
+	/*
+	 * Ring data area is the half minus the header.  Round down to the
+	 * largest power-of-2 so the mask has all lower bits set.
+	 */
+	slots = rounddown_pow_of_two(half - sizeof(struct shmem_ring_hdr));
+
+	mb->tx_hdr->head = 0;
+	mb->tx_hdr->tail = 0;
```
wendyliang25, line 248: Hi @bisingha-xilinx, shall we get the tx head/tail from the ring itself, the same for doorbell tx ring? e.g. if we re-probe the driver?

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -98,16 +172,195 @@ static int plat_map_region(struct amdxdna_dev *xdna, struct device *dev,
...
+	mb->tx_hdr->ring_mask = slots - 1;
+	mb->tx_hdr->rsvd = 0;
+
+	mb->rx_hdr->head = 0;
+	mb->rx_hdr->tail = 0;
+	mb->rx_hdr->ring_mask = slots - 1;
```
wendyliang25, line 254: should the rx ring head/tail got from the ring itself instead of hardcode it to 0?

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
bisingha-xilinx: Hi @wendyliang25, thank you for reviewing.
Current rpu fw expects starting from 0 on rmmod/insmod. getting the head/tail form the ring may inherit a dead session's values which could let to unexpected behaviour. 

Please let me know your thoughts on it.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
bisingha-xilinx: Hi @wendyliang25, thank you.
I think it's the same ask as above. Please check the first response. Please let me know if I need to update now.
Thanks

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -211,45 +502,225 @@ struct mailbox_channel *xdna_mailbox_alloc_channel(struct mailbox *mb)
...
 
+/*
+ * Consume responses the RPU has already produced.  Firmware completes some
+ * commands (SUSPEND in particular) by writing the response without raising the
+ * completion IPI, so the sender calls this before declaring a command lost.
+ */
 void xdna_mailbox_drain_channel(struct mailbox_channel *mailbox_chann)
 {
-	/* TODO: consume mgmt responses already produced into the RX ring. */
+	plat_mailbox_drain(mailbox_chann->mb);
```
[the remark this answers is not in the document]
bisingha-xilinx: Fixed it with latest push. xdna_mailbox_drain_channel() returns early on NULL channel.
Marking this conversation resolved.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -211,45 +502,225 @@ struct mailbox_channel *xdna_mailbox_alloc_channel(struct mailbox *mb)
...
+	unsigned long flags;
+	u32 id;
+	int ret;
+
+	/*
+	 * The payload has to fit the ring and also the 11-bit body-size field
+	 * the firmware parses out of sz_ver.
+	 */
+	if (msg->send_size > mb->mgmt_ring_mask + 1 - sizeof(hdr) ||
+	    msg->send_size > SHMEM_MSG_BODY_SZ) {
```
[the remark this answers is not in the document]
bisingha-xilinx: This ix fixed on latest push. xdna_mailbox_send_msg() now rejects a payload that is not a multiple of 4 before producing.
Marking this conversation as resolved.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -211,45 +502,225 @@ struct mailbox_channel *xdna_mailbox_alloc_channel(struct mailbox *mb)
...
+	spin_unlock_irqrestore(&mb->msg_id_lock, flags);
+
+	if (ret) {
+		kfree(ifm);
+		return ret;
+	}
+
+	hdr.total_size = sizeof(hdr) + msg->send_size;
+	hdr.sz_ver = FIELD_PREP(SHMEM_MSG_BODY_SZ, msg->send_size) |
+		     FIELD_PREP(SHMEM_MSG_PROTO_VER, SHMEM_PROTOCOL_VER);
```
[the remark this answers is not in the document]
bisingha-xilinx: The two transports pci and plat both are mutually exclusive. In this plat transport total size is header+payload and both ends of this ring are written to that contract. and in amdxdna_shmem.h this is documented.

Marking this conversation as resolved.

drivers/accel/amdxdna/amdxdna_shmem.h
```diff
@@ -0,0 +1,262 @@
...
+		return -ENOSPC;
+
+	off = head & ring_mask;
+
+	if (off + total > size) {
+		gap = size - off;
+		*(u32 *)(ring_base + off) = SHMEM_TOMBSTONE;
+		head += gap;
+
+		if (head - tail + total > size)
```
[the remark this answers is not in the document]
bisingha-xilinx: the consumer, never reads the gap, because the head is not published on the -ENOSPC path. the sentinel value is written at the current unpublished head offset and the fw only ever reads below the published head. so, the sentinel sits in which fw can't reach.
I'll reword the comment in shmem_mgmt_produce.
Im marking this conversation as resolved as it is false alarm.

drivers/accel/amdxdna/amdxdna_shmem.h
```diff
@@ -0,0 +1,280 @@
...
+ */
+static inline int shmem_mgmt_produce(struct shmem_ring_hdr *hdr,
+				     void *ring_base, u64 ring_mask,
+				     const struct shmem_msg_hdr *msg_hdr,
+				     const void *payload,
+				     size_t payload_size)
+{
+	u64 size = ring_mask + 1;
+	u64 head = hdr->head;
+	u64 tail = hdr->tail;
```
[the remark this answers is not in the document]
bisingha-xilinx: I'll update in fresh push.
For now, marking this convo as resolved.

drivers/accel/amdxdna/amdxdna_shmem.h
```diff
@@ -0,0 +1,280 @@
...
+ *
+ * Returns 0 on success, -ENOSPC if ring is full.
+ */
+static inline int shmem_db_produce(struct shmem_db_ring *ring,
+				   u64 ring_mask, u32 hw_ctx_id,
+				   u64 *head_cached)
+{
+	u64 size = ring_mask + 1;
+	u64 head = *head_cached;
+	u64 tail = ring->tail;
```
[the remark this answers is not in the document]
bisingha-xilinx: I'll update in fresh push. I've the local changes ready
For now, marking this convo as resolved.

drivers/accel/amdxdna/amdxdna_shmem.h
```diff
@@ -0,0 +1,302 @@
...
+				     void *ring_base, u64 ring_mask,
+				     const struct shmem_msg_hdr *msg_hdr,
+				     const void *payload,
+				     size_t payload_size)
+{
+	u64 size = ring_mask + 1;
+	u64 head = hdr->head;
+	u64 tail = hdr->tail;
+	u32 total = sizeof(*msg_hdr) + payload_size;
+	u64 off, gap;
```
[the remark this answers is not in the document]
bisingha-xilinx: It is already fixed. closing this convo.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -211,45 +625,284 @@ struct mailbox_channel *xdna_mailbox_alloc_channel(struct mailbox *mb)
...
+ * into aie4 state that may be going away.
+ *
+ * The _irq accessors are required, not stylistic: the fan-out takes this same
+ * lock from hardirq context, so acquiring it here with interrupts enabled would
+ * let the IPI deadlock against us on the same CPU.
+ */
+int amdxdna_mailbox_plat_register_notify(struct mailbox *mb, u32 msix_idx,
+					 struct cert_comp *comp)
+{
+	return xa_err(xa_store_irq(&mb->notify_xa, msix_idx, comp, GFP_KERNEL));
```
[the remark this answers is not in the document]
bisingha-xilinx: xa_store_irq() with GFP_KERNEL is the documented pattern rather than a bug. closing this review comment.

drivers/accel/amdxdna/aie4_plat.c
```diff
@@ -48,22 +49,36 @@ int aie4_doorbell_setup(struct amdxdna_hwctx *hwctx,
...
 {
 	struct amdxdna_dev_hdl *ndev = hwctx->client->xdna->dev_handle;
+	int ret;
 
-	amdxdna_mailbox_plat_ring_doorbell(ndev->mbox, hwctx->priv->hw_ctx_id);
+	ret = amdxdna_mailbox_plat_ring_doorbell(ndev->mbox,
+						 hwctx->priv->hw_ctx_id);
+	if (ret)
+		XDNA_ERR(ndev->aie.xdna, "Ring doorbell for hw_ctx %d failed, ret %d",
+			 hwctx->priv->hw_ctx_id, ret);
```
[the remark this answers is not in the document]
bisingha-xilinx: Keeping this out of scope on this PR to avoid changing common fiiles. The failure is now logged. closing this conversation.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
wendyliang25: Hi @bisingha-xilinx, I have created ticket about how client knows RPU FW is up. for now, we assume that when client probes, RPU FW is already up. in this case, we set our index from the value's in the ring, client just don't assume it's init value, for rx, we just drain/drops the messages in the ring if there is any. Client should not assume the ring always starts from 0. if we do this, we cannot reprobe the driver.

bisingha-xilinx: Hi @wendyliang25 thank you for reviewing and your inputs. 
Sure, I'll update it in my next push.

Thanks.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
bisingha-xilinx: I've got the response in another conversation which is the same ask. So, Im closing this conversation and keeping the other open for now.

Thanks

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -211,45 +625,284 @@ struct mailbox_channel *xdna_mailbox_alloc_channel(struct mailbox *mb)
...
+	}
+
+unlock:
+	spin_unlock_irqrestore(&mb->tx_lock, flags);
+
+	if (ret) {
+		/*
+		 * Drop the inflight record either way: the caller gets the
+		 * error and its &xdna_notify goes out of scope, so no response
+		 * may be routed to it.
```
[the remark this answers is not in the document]
bisingha-xilinx: The head cant be rewound: shmem_mgmt_produce() publishes it with a dma_wmb() before the IPI, and the rpu fw reads the TX ring directly and completes some commands without waiting for the IPI, so by the time plat_ipi_kick() fails the message may already be consumed - and un-publishing it would corrupt the live ring. 

Not a valid ask. Closing this conversation.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -152,6 +499,40 @@ static void plat_mailbox_release(void *data)
 	if (!IS_ERR_OR_NULL(mb->tx_chan))
 		mbox_free_channel(mb->tx_chan);
 	cancel_work_sync(&mb->rx_work);
```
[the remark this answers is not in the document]
bisingha-xilinx: mbox_free_channel() does not free the mbox_chan - it clears chan->cl and returns the channel to the controller - so there is no freed object for a late callback to touch. What was wrong is that rx_stopped was only set by stop_channel(), so a probe-failure unwind reached plat_mailbox_release() with the flag still clear. Release now sets WRITE_ONCE(mb->rx_stopped, true) at the top, gating the callback on every teardown path rather than only the orderly one.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -98,16 +231,331 @@ static int plat_map_region(struct amdxdna_dev *xdna, struct device *dev,
...
+	int ret;
+
+	/*
+	 * The RX channel outlives the mgmt channel: it is released with the
+	 * platform device, while stop_channel() runs before the caller destroys
+	 * the cert-completion state and the work buffer.  Once stopped, nothing
+	 * here may touch that state, so only the ack below still runs.
+	 */
+	if (!READ_ONCE(mb->rx_stopped)) {
+		plat_mailbox_cert_notify(mb);
```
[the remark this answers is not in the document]
bisingha-xilinx: The post-stop_channel() window is already covered: a late rx_work finds mgmt_chann == NULL, cleared under rx_lock, so it returns without consuming; and cert completions are removed from notify_xa under the same lock the fan-out holds, before any cert_comp is freed.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -98,16 +231,331 @@ static int plat_map_region(struct amdxdna_dev *xdna, struct device *dev,
...
+
+	/*
+	 * Past the sentinel the peer is live, so its indices are the protocol's
+	 * truth and their values are adopted as they stand.  Second-guessing
+	 * how far apart they are would be host-side scaffolding around a
+	 * firmware defect: we own both ends of this protocol, so a
+	 * desynchronised ring is fixed on the RPU side.
+	 */
+	mb->tx_head_cached = mb->tx_hdr->head;
+	mb->db_head_cached = mb->db_ring->head;
```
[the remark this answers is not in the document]
bisingha-xilinx: Yes, it's a valid concern since now head/tails are taken from the rings itself instead of zeroing. 
MSG IDs are now unique by construction: derived from the TX ring position the message occupies, (u32)((tx_head_cached >> 2) + 1). head is adopted at probe and only advances, so every id a previous session used lies below it and no two messages share one. The only thing that rewinds head is firmware re-init, which zeroes head and tail together — so whenever the id space restarts, the ring is already empty. The shift is lossless because head stays 4-aligned, and the +1 keeps 0 reserved for RPU-initiated messages.
Already fixed in latest push

drivers/accel/amdxdna/amdxdna_shmem.h
```diff
@@ -0,0 +1,356 @@
...
+		return -EPROTO;
+	}
+
+	memcpy(msg_hdr, ring_base + off, sizeof(*msg_hdr));
+
+	if (msg_hdr->total_size < sizeof(*msg_hdr) ||
+	    !IS_ALIGNED(msg_hdr->total_size, sizeof(u32)) ||
+	    msg_hdr->total_size > head - tail ||
+	    msg_hdr->total_size > avail)
+		return -EPROTO;
```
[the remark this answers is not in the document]
bisingha-xilinx: This is already fixed. The code placement got changed.

Closing the review comment.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -98,16 +234,331 @@ static int plat_map_region(struct amdxdna_dev *xdna, struct device *dev,
...
+ * plain memory rather than through the IO accessors (see the rationale there),
+ * hence the __force casts here.
+ */
+static int plat_rings_init(struct mailbox *mb)
+{
+	void *mgmt = (void __force *)mb->mgmt_base;
+	resource_size_t half = mb->mgmt_size / 2;
+	u64 rx_head, tx_tail, db_tail, slots;
+	int ret;
+
```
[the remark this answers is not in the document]
bisingha-xilinx: This is already fixed. both the channel bases are strictly aligned now.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -98,16 +234,331 @@ static int plat_map_region(struct amdxdna_dev *xdna, struct device *dev,
...
+	/* Split mgmt region: first half is TX, second half is RX */
+	mb->tx_hdr = mgmt;
+	mb->tx_ring = mgmt + sizeof(struct shmem_ring_hdr);
+
+	mb->rx_hdr = mgmt + half;
+	mb->rx_ring = mgmt + half + sizeof(struct shmem_ring_hdr);
```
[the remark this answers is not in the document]
bisingha-xilinx: This is already fixed. both the channel bases are strictly aligned now.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -98,16 +234,331 @@ static int plat_map_region(struct amdxdna_dev *xdna, struct device *dev,
...
+static void plat_mailbox_cert_notify(struct mailbox *mb)
+{
+	struct cert_comp *cert_comp;
+	unsigned long flags;
+	unsigned long idx;
+
+	xa_lock_irqsave(&mb->notify_xa, flags);
+	xa_for_each(&mb->notify_xa, idx, cert_comp)
+		wake_up_all(&cert_comp->waitq);
+	xa_unlock_irqrestore(&mb->notify_xa, flags);
```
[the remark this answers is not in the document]
bisingha-xilinx: This is also not correct. The lock is the lifetime mechanism, which the suggestion's own "with a safe lifetime mechanism" qualifier concedes. Waking after unlocking reopens the exact use-after-free the lock closes.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -98,16 +236,340 @@ static int plat_map_region(struct amdxdna_dev *xdna, struct device *dev,
...
+	slots = (mb->doorbell_size - offsetof(struct shmem_db_ring, data)) /
+		sizeof(u32);
+	slots = rounddown_pow_of_two(slots);
+	mb->db_ring_mask = slots - 1;
+
+	mb->db_ring->ring_mask = mb->db_ring_mask;
+
+	ret = plat_wait_fw_alive(mb);
+	if (ret)
+		return ret;
```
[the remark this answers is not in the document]
bisingha-xilinx: Fixes are in local changes. I'll force push. 
Resolving it.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -211,45 +789,313 @@ struct mailbox_channel *xdna_mailbox_alloc_channel(struct mailbox *mb)
...
+ */
 int xdna_mailbox_start_channel(struct mailbox_channel *mb_chann,
 			       const struct xdna_mailbox_chann_res *x2i,
 			       const struct xdna_mailbox_chann_res *i2x,
 			       u32 xdna_mailbox_intr_reg, int mb_irq)
 {
-	/* TODO: mgmt shmem ring bring-up (no MSI-X regs on the platform). */
+	mb_chann->mb->rx_stopped = false;
+	mb_chann->mb->bad_state = false;
+	mb_chann->mb->mgmt_chann = mb_chann;
```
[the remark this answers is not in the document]
bisingha-xilinx: I've the fixes in local changes. I'll force push in a moment. 
Resolving this conversation.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -98,16 +236,340 @@ static int plat_map_region(struct amdxdna_dev *xdna, struct device *dev,
...
+		 * Id 0 marks an RPU-initiated message (async error/event
+		 * note) rather than a response to one of our commands, so it is
+		 * routed to the async sink instead of the inflight table.
+		 */
+		if (!msg_hdr.id) {
+			if (chann->async_cb)
+				chann->async_cb(chann->async_handle,
+						msg_hdr.opcode,
+						(void __iomem __force *)buf,
+						payload_size);
```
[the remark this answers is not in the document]
bisingha-xilinx: Documented at this call site now -- the `notify_cb` call below already carried
the rationale and this one did not. No misdecode or fault. Skipping this review comment.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
bisingha-xilinx: Hi @wendyliang25 Thank you for reviewing.
I've updated accordingly. And rpu fw also writes a sentinel in rings reserved area when it up to advertise its ready. now the drive also checks that sentinel while probing. Please review again.

Thank you.
Closing this conversation

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -211,45 +798,317 @@ struct mailbox_channel *xdna_mailbox_alloc_channel(struct mailbox *mb)
...
+		 * IPI-failure path the message is already published and is
+		 * deliberately left in the ring: head cannot be rewound,
+		 * because the RPU reads it directly and completes some
+		 * commands without waiting for the IPI, so it may have
+		 * consumed the message already.  A failed IPI means the
+		 * doorbell itself is broken, and one leaked ring slot is the
+		 * lesser loss.
+		 */
+		if (xa_erase(&mb->msg_xa, id))
+			kfree(ifm);
```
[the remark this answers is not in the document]
bisingha-xilinx: Fixed locally. I'll force push in a moment.
Closing this conversation.
Thanks

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -98,16 +236,349 @@ static int plat_map_region(struct amdxdna_dev *xdna, struct device *dev,
...
+		/*
+		 * Only drain when the RPU has actually queued a response.
+		 * Pure doorbell-completion IPIs leave the mgmt ring empty, so
+		 * scheduling rx_work for them just adds a needless wakeup on
+		 * the submit fast path.  SPSC-safe: the RPU publishes head
+		 * with a write barrier before raising the IPI, and tail is
+		 * host-owned so the cached copy is authoritative.
+		 */
+		if (READ_ONCE(mb->rx_hdr->head) != READ_ONCE(mb->rx_tail_cached))
+			schedule_work(&mb->rx_work);
```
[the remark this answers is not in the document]
bisingha-xilinx: It's not true. The fw already has the barrier in its produce side before trggering the ipi to the Linux. The flow on the other side writes/updates the head of rx_header -> goes thru the barrier -> then only kicks the IPI so the head of the rx_header is already updated a visible by the time this driver checks and schedules work in the rx_callback.

Im closing this review comment since it's not a valid concern.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -98,16 +236,349 @@ static int plat_map_region(struct amdxdna_dev *xdna, struct device *dev,
...
+ * call site so it binds to the observation itself.
+ */
+static int plat_wait_fw_alive(struct mailbox *mb)
+{
+	unsigned long timeout = jiffies + msecs_to_jiffies(SHMEM_FW_ALIVE_TIMEOUT_MS);
+
+	for (;;) {
+		/* READ_ONCE: the peer writes this field while we spin on it. */
+		if ((u32)READ_ONCE(mb->rx_hdr->rsvd) == SHMEM_FW_ALIVE_MAGIC) {
+			dma_rmb();
```
wendyliang25, line 262: why do we need dma_rmb() here?

wendyliang25: just wonder if rpu fw can kick IPI after it set the value. and thus if this value is not detected from client side, client can wait for interrupt ? but for now, it is ok.

drivers/accel/amdxdna/amdxdna_plat_ring.h
```diff
@@ -0,0 +1,392 @@
+/* SPDX-License-Identifier: GPL-2.0 */
```
wendyliang25, line 1: can we rename this file? as shmem can mean shmem between processes in Linux, maybe amdxdna_plat_ipc.h?

bisingha-xilinx: Hi @wendyliang25 thanks for reviewing. Yes, thats correct. Do you think amdxdna_plat_ring.h aligns more than amdxdna_plat_ipc.h?
Thank you.

chvamshi-xilinx (dismissed):

drivers/accel/amdxdna/amdxdna_plat_ring.h
wendyliang25: `amdxdna_plat_ring.h` is fine for me? any suggestion from copilot?

drivers/accel/amdxdna/amdxdna_shmem.h
```diff
@@ -0,0 +1,392 @@
...
+	u8  _pad1[56];			/* offset 72: pad to 128-byte total    */
+} __aligned(64);
+
+static_assert(offsetof(struct shmem_ring_hdr, tail) == 64);
+static_assert(sizeof(struct shmem_ring_hdr) == 128);
+
+/*
+ * Liveness sentinel the RPU writes into rsvd of the ring it produces into
+ * -- the host's RX ring.  It is set once the firmware has initialised its
+ * shmem clients and re-asserted on every IPI, so reading it is how the host
```
wendyliang25: replace `shmem clients` with npu clients. 
instead of mentioning RPU, we can say fw this is to line up with PCI bus implementation

drivers/accel/amdxdna/amdxdna_plat_ring.h
bisingha-xilinx: Hi Wendy, yes, it also suggested to have amdxdna_plat_ring.h since amdxdna_plat_ipc.h will also be kind of confusing just like amdxdna_shmem.h

Thank you.

drivers/accel/amdxdna/amdxdna_shmem.h
bisingha-xilinx: Hi @wendyliang25 
Sure. 
Thank you.

drivers/accel/amdxdna/amdxdna_shmem.h
```diff
@@ -0,0 +1,392 @@
...
+	u64 head;			/* offset  0: producer index           */
+	u64 ring_mask;			/* offset  8: (ring_data_size - 1)     */
+	u64 rsvd;			/* offset 16: reserved / FW alive magic */
+	u8  _pad0[40];			/* offset 24: pad to cache line boundary */
+	/* --- 64-byte cache line boundary --- */
+	u64 tail;			/* offset 64: consumer index           */
+	u8  _pad1[56];			/* offset 72: pad to 128-byte total    */
+} __aligned(64);
+
+static_assert(offsetof(struct shmem_ring_hdr, tail) == 64);
```
wendyliang25: please also rename the `shmem_` in this file

drivers/accel/amdxdna/amdxdna_plat_ring.h
bisingha-xilinx: Hi @wendyliang25 
I've renamed the file to amdxdna_plat_ring.h along with structs/functions/struct members. 
Thank you

drivers/accel/amdxdna/amdxdna_shmem.h
bisingha-xilinx: Yes, I've updated them while renaming the file to amdxdna_plat_ring.h.
Thank you.

drivers/accel/amdxdna/amdxdna_plat_ring.h
```diff
@@ -0,0 +1,398 @@
...
+ *
+ * Takes the cached head so this agrees with plat_ring_db_produce(): a WC
+ * read-back of the host-owned index can lag and report space the producer does
+ * not see.
+ */
+static inline bool plat_ring_db_has_space(struct plat_ring_db *ring,
+					  u64 ring_mask, u64 head)
+{
+	return (head - ring->tail) <= ring_mask;
+}
```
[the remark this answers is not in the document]
bisingha-xilinx: It's not a valid concern. This function is used by the wait_event_timeout() which only reads the tail and does arithmetic operations and returns. It doesnt access anything tail guards, so there is no second access to order it against and barrier to pair with either.
Other four places mentioned already have their barriers correctly.

drivers/accel/amdxdna/amdxdna_plat_ring.h
bisingha-xilinx: Hi 
Closing this conversation.
Thank you.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
bisingha-xilinx: Hi @wendyliang25 thank you for reviwing.
Assuming it's not a concern atleast for now. So, closing this conversation.
Thank you.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -152,21 +642,72 @@ static void plat_mailbox_release(void *data)
...
 		mbox_free_channel(mb->tx_chan);
 	cancel_work_sync(&mb->rx_work);
+
+	/* Anything still inflight can no longer be answered. */
+	xa_for_each(&mb->msg_xa, idx, ifm) {
+		xa_erase(&mb->msg_xa, idx);
+		kfree(ifm);
+	}
+	xa_destroy(&mb->msg_xa);
+	xa_destroy(&mb->notify_xa);
```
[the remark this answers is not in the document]
bisingha-xilinx: It's not a valid concern. the plat_mailbox_cert_notify() holds the same nodify_xa lock across the whole work which is the same lock xa_destroy() also holds. So, these two are mutually exclusive. 
Closing this conversation.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -152,21 +642,72 @@ static void plat_mailbox_release(void *data)
...
+	 * EL3, and the two are not ordered against each other by the SMC.
+	 * Nothing on the way there supplies it either: the mailbox framework
+	 * takes the channel lock, whose acquire says nothing about stores
+	 * already issued, and zynqmp-ipi goes straight to the SMC with no
+	 * barrier of its own.  This is the same dma_wmb() a writel() doorbell
+	 * would carry implicitly.
+	 */
+	dma_wmb();
+
+	spin_lock_irqsave(&mb->ipi_lock, flags);
```
wendyliang25: just wonder is mbox framework requires a lock here? we just kick IPI with data field "NULL" in this case.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -177,6 +718,14 @@ struct mailbox *xdnam_mailbox_create(struct drm_device *ddev,
 	mb->xdna = xdna;
 	mb->pdev = pdev;
 	INIT_WORK(&mb->rx_work, plat_mailbox_rx_work);
+	xa_init(&mb->msg_xa);
+	/* Walked from the IPI callback, so the xarray lock must be irq-safe. */
+	xa_init_flags(&mb->notify_xa, XA_FLAGS_LOCK_IRQ);
+	spin_lock_init(&mb->tx_lock);
+	spin_lock_init(&mb->db_lock);
+	spin_lock_init(&mb->ipi_lock);
```
wendyliang25: do we really need ipi_lock?

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -187,6 +736,48 @@ struct mailbox *xdnam_mailbox_create(struct drm_device *ddev,
...
+	 * Each region needs room for a whole record past its ring header, not
+	 * merely one byte: a region that clears the header by less than that
+	 * rounds down to a ring too small to ever carry one, which surfaces as
+	 * every send failing -ENOSPC instead of as the bad device tree it is.
+	 * For the mgmt halves the record is a message header; undersizing them
+	 * also underflows the capacity check in xdna_mailbox_send_msg(), which
+	 * subtracts that header from the ring size in unsigned arithmetic.
+	 */
+	if (mb->mgmt_size / 2 < sizeof(struct plat_ring_hdr) + sizeof(struct plat_ring_msg_hdr) ||
+	    mb->doorbell_size < offsetof(struct plat_ring_db, data) + sizeof(u32)) {
```
wendyliang25: why size compared with offset?

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -187,6 +736,48 @@ struct mailbox *xdnam_mailbox_create(struct drm_device *ddev,
...
+	}
+
+	/*
+	 * The ring headers are __aligned(64) so that head and tail sit on
+	 * separate cache lines and each index is a single-copy-atomic u64
+	 * against the firmware.  Both properties come from where the region
+	 * lands, which is device-tree data: the mgmt RX header sits at half the
+	 * region, so an odd size would misalign it and silently cost both.
+	 */
+	static_assert(__alignof__(struct plat_ring_db) == __alignof__(struct plat_ring_hdr));
```
wendyliang25: why do need this this assert? it is about the structure itself?

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -187,6 +736,48 @@ struct mailbox *xdnam_mailbox_create(struct drm_device *ddev,
...
+	if (!IS_ALIGNED((unsigned long)(void __force *)mb->mgmt_base, align) ||
+	    !IS_ALIGNED((unsigned long)(void __force *)mb->doorbell_base, align) ||
+	    !IS_ALIGNED(mb->mgmt_size / 2, align)) {
```
wendyliang25: to check alignment, is it better not to "/2" what  happen if mgmt_size is odd ?

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -211,45 +802,331 @@ struct mailbox_channel *xdna_mailbox_alloc_channel(struct mailbox *mb)
...
 			       const struct xdna_mailbox_chann_res *x2i,
 			       const struct xdna_mailbox_chann_res *i2x,
 			       u32 xdna_mailbox_intr_reg, int mb_irq)
 {
-	/* TODO: mgmt shmem ring bring-up (no MSI-X regs on the platform). */
+	struct mailbox *mb = mb_chann->mb;
+
+	mutex_lock(&mb->rx_lock);
+	WRITE_ONCE(mb->rx_stopped, false);
+	WRITE_ONCE(mb->bad_state, false);
```
wendyliang25: we can merge rx_stopped and bad_state into single field, if it needs multi state, can make it u32 instead of bool

wendyliang25: why ring is in bad_state? is ring just has full non-empty and empty? why it has state?, i think the ring and the IPI is stateless, only RPU FW has state.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -9,50 +9,188 @@
...
+	 */
+	struct xarray		notify_xa;
+
+	/*
+	 * Set before stop_channel() drains rx_work.  The RX channel stays live
+	 * until the platform device is released, so without this a late IPI
+	 * would queue the work again right after cancel_work_sync() drained it,
+	 * or walk cert-completion state the caller has already destroyed.
+	 */
+	bool			rx_stopped;
```
wendyliang25: why do we have rx_stop? it doesn't look like we have anything to stop.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -9,50 +9,188 @@
...
+
+	/*
+	 * Set when the RX ring yields something that is not a valid message.
+	 * An unexpected receive error means the link can no longer be trusted,
+	 * so the drain stops where it is and further sends are refused until
+	 * the channel is restarted.
+	 */
+	bool			bad_state;
```
wendyliang25: i think ring doesn't have bad state, it only has full, empty, and non-empty, and ipi kick in our case, is just kick, no bad state either.

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -98,16 +236,351 @@ static int plat_map_region(struct amdxdna_dev *xdna, struct device *dev,
...
+	 * helpers in amdxdna_plat_ring.h preserve 4-alignment without
+	 * establishing it, so this is the one place it can be.  A misaligned
+	 * index cannot be repaired without moving it out from under a live peer,
+	 * so refuse.
+	 *
+	 * The doorbell indices are exempt: they count slots, not bytes, and
+	 * head & ring_mask indexes data[] in bounds whatever its alignment.
+	 */
+	if (!IS_ALIGNED(mb->tx_head_cached, sizeof(u32)) ||
+	    !IS_ALIGNED(rx_head, sizeof(u32))) {
```
wendyliang25: why do we need to do this check?

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -211,45 +802,331 @@ struct mailbox_channel *xdna_mailbox_alloc_channel(struct mailbox *mb)
...
 }
 
 void xdna_mailbox_set_async_cb(struct mailbox_channel *mailbox_chann,
 			       void *async_handle, xdna_mailbox_async_cb_t async_cb)
 {
-	/* TODO: route firmware-initiated (id 0) messages from the mgmt RX ring. */
+	struct mailbox *mb = mailbox_chann->mb;
+
+	/* Pair with the RX consumer, which reads these under rx_lock. */
+	mutex_lock(&mb->rx_lock);
```
wendyliang25: async_handle is only required to set in device probe, why need to do rx_lock? what's the  rx_lock mutex tries to protect?

drivers/accel/amdxdna/amdxdna_mailbox_plat.c
```diff
@@ -211,45 +802,331 @@ struct mailbox_channel *xdna_mailbox_alloc_channel(struct mailbox *mb)
...
+ */
 void xdna_mailbox_stop_channel(struct mailbox_channel *mailbox_chann)
 {
-	/* TODO: quiesce the mgmt shmem ring. */
+	struct mailbox *mb = mailbox_chann->mb;
+	struct plat_inflight_msg *ifm;
+	unsigned long idx;
+
+	mutex_lock(&mb->rx_lock);
+	mb->mgmt_chann = NULL;
```
wendyliang25: this should be done when device is unprobe, is there any mailbox unregister?

PR #1711, opened 2026-09-18
accel/tools: split GENMASK_U64 feature test from BIT_U64

bkeil:
Some kernels, like 6.12.107+deb13-amd64, have received a backport of GENMASK_U64 but not BIT_U64.

When the current configure_kernel.sh defines both, the build fails with -Werror on the redefinition in linux/bits.h


maxzhen (approved):

maxzhen:
ok to test

